Issue #64, opened 2016-11-29
Merge computeTemp for rad and non-rad


maxpkatz:
Radiation has its own logic for computeTemp, this should be merged with Castro::computeTemp. Also, we should clarify if we want the radiation-specific options to live on; if they're not actively used, maybe just consolidate into the existing Castro code.

zingale:
once difference here is that the hydro uses a `growntilebox()` whereas radiation uses a `tilebox`

zingale:
also, the radiation version of `computeTemp` takes 2 arguments, while the hydro version only takes 1.

zingale:
cleanest fix is to just make radiation use the pure hydro methodology

PR #69, opened 2016-12-05
merge computeTemp from radiation to hydro

zingale:
This PR removes the radiation-specific version of the `computeTemp` function and instead has everything go through the hydro version.  There are some minor differences in how the internal energy is handled now for radiation for resets (e.g., it now does the same thing pure hydro does, which is a good thing).

Also, there is no test coverage of the `do_real_eos=0` option for radiation, but I _believe_ that code is accurately changed over.

Wait until the test suite finishes before we merge this (assuming that the differences are reasonable)


zingale:
Here are the test suite results:

http://bender.astro.sunysb.edu/Castro/test-suite/test-suite-gfortran/2016-12-05-002/index.html

errors look reasonable to me.

zingale:
Okay, the tests of the `do_real_eos = 0` don't show any changes.  The other tests show changes ~ 1.e-5 level.

okay to merge this?

maxpkatz:
LGTM

zingale:
closes #64

PR #71, opened 2016-12-06
Runparams

zingale:
this makes the radiation transport parameterization runtime parameters scalars instead of arrays of length 1.


maxpkatz:
Fine with me, though the other solution is to make them vectors of variable length.

zingale:
there's no need for them to be vectors -- that is a remnant of a multifluid radiation solver, where each fluid could have different properties.

Issue #81, opened 2016-12-14
move radiation lambda into qaux


zingale:
We can simplify a lot of interfaces if we move the radiation lambda into qaux

zingale:
it looks like the place to do this is in `Castro_hydro.cpp` in the call to `ctoprim`

zingale:
also, I believe it is the case that lamda is never updated in the hydro -- it is computed in the radiation code, so putting it into `qaux` should be fine

zingale:
@WeiqunZhang @maxpkatz any objections to this clean-up?

The idea would be to have `ctoprim` add lambda to `q_aux`, then then we never need to explicitly pass `lam` through the various interfaces, since we will have `q_aux`

maxpkatz:
No objection. Weiqun agreed too.

PR #87, opened 2016-12-31
this changeset puts the radiation flux limiter, lambda, inside qaux

zingale:
This is a cleanup -- it simplifies a lot of interfaces.  There should be no changes introduced.  The test suite confirms this:

http://bender.astro.sunysb.edu/Castro/test-suite/test-suite-gfortran/2016-12-30-001/index.html

Upon merge, we will squish these into a single commit


zingale:
this closes #81

Issue #34, opened 2016-09-12
PRECISION=SINGLE will not work


zingale:
Formally, boxlib supports double and single precision, and can be compiled to support either via the `PRECISION` variable in the makefiles.

However, all of the Fortran routines that we write assume that we pass data through as `double *`, so only double precision will work.

WeiqunZhang:
This is actually something I am working on now.  I am adding a fortran module to C_BaseLib.  It includes a c_real type which is C++ BoxLib's Real.  On fortran side, we can then use real(c_real) x.

WeiqunZhang:
The bl_fort_module is in BoxLib now.  Note that it also contains an integer parameter, bl_spacedim.

```
module bl_fort_module

  use iso_c_binding, only : c_float, c_double

  implicit none

  integer, parameter :: bl_spacedim = BL_SPACEDIM

#ifdef BL_USE_FLOAT
  integer, parameter :: c_real = c_float
#else
  integer, parameter :: c_real = c_double
#endif

end module bl_fort_module
```

zingale:
should we change the Castro Fortran code to use `real(c_real) :: x` now, or should we not worry about this.  It is unlikely that we will ever get the whole code, e.g., Microphysics, working in single precision.  Also unclear if this is desired.  

An alternate _fix_ is to just abort in the make if someone tries to compile in SINGLE precision.

maxpkatz:
We should change the Fortrn code to use real(c_real) independent of this change. But that would indirectly solve much of this problem.

zingale:
so does this mean that any constants in the code should then be like

```
1.0_c_real
```

zingale:
started a script in `Castro/util/scripts/dpconvert.py` that will make this change.  It does the declarations at the moment, but still needs to address constants and adding the use statement in the appropriate place.

zingale:
this is mostly addressed by PR #88

PR #88, opened 2017-01-04
switch all double precision declarations and constants to c_real

zingale:
use the BoxLib-compatible c_real type so we support single
and double precision options.  This was updated using
dpconvert.py.  This addresses issue #34

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,12 @@
 # changes since the last release:
 
+  -- to support both single and double precision, all of the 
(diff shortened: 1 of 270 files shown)


zingale:
this is being tested right now

zingale:
all tests passed:

http://bender.astro.sunysb.edu/Castro/test-suite/test-suite-gfortran/2017-01-04-001/index.html

maxpkatz:
Good by me to merge.

Issue #35, opened 2016-09-12
StateData Reactions_Type shouldn't need rho edot and edot


zingale:
We store both rho edot and edot in the `Reactions_Type` state data.  We should not need both

zingale:
note: the tricky bit here is that we need to ensure backwards compatibility with plotfiles

zingale:
this is not an urgent issue

PR #89, opened 2017-01-06
remove enuc from Reactions_Type

zingale:
this addresses #35 -- we had both rho_enuc and enuc in the Reactions_Type statedata -- seems redundant, so we remove enuc.


zingale:
this was finally addressed in #927

Issue #58, opened 2016-11-16
Make the mass/momentum/energy reporting more uniform


maxpkatz:
Some source terms report mass/energy/momentum added to the domain (hydro, gravity, etc.) but not every source term does and the reporting mechanism/formatting is not guaranteed to be consistent. We should have a single routine that takes an update MultiFab (which always has NUM_STATE components) and computes how much of these quantities that update adds and reports it to the stdout in a uniform manner.

PR #94, opened 2017-01-14
Diagnostics updates

maxpkatz:
This set of changes significantly simplifies how we print diagnostic information regarding updates to the state. Every source term -- including the user-defined external source terms, the built-in physics like gravity, the hydrodynamics, and density and energy resets -- will now print its updates in a uniform fashion if print_update_diagnostics == 1 (note that this parameter has been renamed from print_energy_diagnostics).

This allowed us to remove parameters like E_added from interfaces to Fortran kernels, which simplifies them a bit.

Also, to slightly simplify conditional logic, ext_src was placed first in the list of source terms to evaluate. This may result in small changes for some problems.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,13 @@
 # changes since the last release:
 
+  -- diagnostic information about how source terms update the
+     state has been overhauled and made uniform. All source terms,
+     including hydro and resets, print their changes to the state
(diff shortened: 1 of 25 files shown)


maxpkatz:
This addresses issue #58.

zingale:
all tests passed

PR #97, opened 2017-01-23
Writing particle information to the plotfiles using the standard BoxLib format

atmyers:
This will allow the particles to be read in by visualization / analysis codes that support the BoxLib format, such as yt.

### DIFF
--- a/Source/Castro.H
+++ b/Source/Castro.H
@@ -160,6 +160,10 @@ public:
     //
     void ParticleCheckPoint (const std::string& dir);
     //
+    // Write particles in plotfile directories
+    //
+    void ParticlePlotFile (const std::string& dir);
+    //
     // How to initialize at restart
     //
     void ParticlePostRestart (const std::string& restart_file);
--- a/Source/CastroParticles.cpp
+++ b/Source/CastroParticles.cpp
@@ -104,6 +104,18 @@ Castro::ParticleCheckPoint(const std::string& dir)
 }
 
 void
+Castro::ParticlePlotFile(const std::string& dir)
+{
+    if (level == 0)
+    {
+      //  We call TracerPC->Checkpoint instead of TracerPC->WritePlotFile
+      //  so that the particle ids also get written out.
+        if (TracerPC)
+            TracerPC->Checkpoint(dir, chk_tracer_particle_file);
+    }
+}
+
+void
 Castro::ParticlePostRestart (const std::string& restart_file)
 {
     if (level == 0)
--- a/Source/Castro_io.cpp
+++ b/Source/Castro_io.cpp
@@ -807,6 +807,11 @@ Castro::writePlotFile (const std::string& dir,
                        ostream&       os,
                        VisMF::How     how)
 {
+
+#ifdef PARTICLES
+  ParticleCheckPoint(dir);
+#endif
+
     int i, n;
     //
     // The list of indices of State to write to plotfile.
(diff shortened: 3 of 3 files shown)


maxpkatz:
Comments:

1) It looks like we already do the particle checkpoint as part of the Castro checkpoint routine. This change is  to add the same call to the writePlotFile routine. Are we sure that we don't want the corresponding writePlotFile routine from the TracerParticles class? What is the difference between a plotfile and a checkpoint for particles?

2) I know this is not strictly relevant to the PR, but it would be very nice if we had a test that used particles in Exec/.

3) I'd prefer if the PR gets assigned to development branch, not master.

WeiqunZhang:
Exec/hydro_tests/RT_particles is a test using tracer particles.

maxpkatz:
Thanks, I forgot about that one. I tested it and it looks like it makes no difference, the checkpoint and WritePlotFile routines do the same thing. I also looked at the code in C_ParticleLib/Particles.H and confirmed that WritePlotFile is just a wrapper to checkpoint. I'd still prefer if we called that wrapper for consistency's sake, but other than that (and doing it on development branch) I'm good with this.

atmyers:
The difference between a checkpoint and a plotfile for particles is the amount of data that gets written out. Checkpoints write out additional integer attributes for every particle, such as the particle's unique id and the MPI process it's assigned to, that are used to restart the simulation. Other than that, the format is exactly the same. 

As for which we want to use, it's a tradeoff between ease of debugging versus size of the plt files. It can be useful to access the extra information, but it makes the plot files bigger by 20 bytes per particle. Nyx, it looks like, doesn't write out the extra information. I'll defer to your judgement for which one you want.

maxpkatz:
Is the extra information useful for analysis purposes like visualization or post-processing? Normally I'd say that it makes sense for the MPI process to be checkpoint-specific data that you wouldn't need to know for post-processing, but you might want to know the particle's ID?

maxpkatz:
There's also a comment in Particles.H added by Ann a couple years back that says

"// For yt we need exactly the chk particle format so would need to set is_checkpoint = true"

Can you confirm whether or not this is still true?

atmyers:
The extra information is the id, the cpu, and three integers that store which cell the particle is in. Of those, I'd say the id is quite useful in the context of analysis / viz, especially for tracers. The others, not so much. 

I recently updated the yt reader, so it's no longer true that it *has* to be the checkpoint format.

maxpkatz:
Ok. I am not concerned about having a little extra data in the plotfile. So your solution seems fine to me. The only thing I'd request is for you to write a wrapper around this in CastroParticles.cpp called Castro::ParticlePlotFile (say) that calls the Checkpoint routine like you're already doing, and clearly comments why you're calling a checkpoint routine in a plotfile (so that ignorant people like me aren't confused in the future).

atmyers:
OK - will do.

zingale:
We definitely need the particle id -- that's the whole purpose of particles in castro, for Lagrangian tracers.  If that's in the plotfile, then I'll merge.

WeiqunZhang:
If particle us saved, the process id should also be saved.  Id itself is not unique. Id plus process id is. Note both are set when a particle us created and they do not change.

atmyers:
Both the particle ids and the process ids are currently being written to the plotfiles in this PR.

zingale:
great!

PR #98, opened 2017-01-26
Allow sponge to be active in an interior region

jaharris87:


jaharris87:
This only works with the radius controls (sponge_lower_radius and sponge_upper_radius).

Analogous controls for density might not be so straight-forward due to possible nonmonotonicities.

PR #110, opened 2017-02-24
start of the post_simulation() hook

zingale:
this is a WIP

is this interface okay?

### DIFF
--- a/Source/main.cpp
+++ b/Source/main.cpp
@@ -30,6 +30,7 @@
 #include <XGraph1d.H>
 #endif
 
+#include "Castro.H"
 #include "Castro_io.H"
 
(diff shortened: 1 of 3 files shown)


Source/main.cpp
```diff
@@ -161,6 +162,9 @@ main (int   argc,
 
     }
 
+    Castro* castro = dynamic_cast<Castro*>(&amrptr->getAmrLevels()[0]);
```
WeiqunZhang: Because post_simulation is a static function, it can be called directly,

Castro::post_simulation(amrptr->getAmrLevels());

zingale:
will relaunch from a different branch

Issue #108, opened 2017-02-23
add problem_post_simulation() hook


zingale:
We should add a hook to a problem-specific routine at the end of the simulaiton, `problem_post_simuation()` in `main.cpp` before the final plotfile (line 162).  This could be used, for example, to report error with respect to an analytic solution.

This could also help with removing some of the hardcoded problem stuff in the radiation code (issue #36)

zingale:
this is done now in #111

Issue #107, opened 2017-02-23
diffusion test problem should compare to analytic solution


zingale:
We should have the diffusion test problem compare to the analytic solution at the end of the simulation, as we do in Maestro's version.  This can report the L2 norm of the error to check for convergence.

zingale:
alternately, we can write a standalone script to do the error, based on `fgaussianpulse.f90` in the `Postprocessing/` stuff

zingale:
@maxpkatz suggests we create a derived plot variable to compute the analytic solution and the error from it.  Then the test suite can do the norm for us (or we can output that ourselves).  Also, we can have a post simulation hook to output the norm.

Also, we can access the derive function anywhere as:
```
MultiFab mf* = derive("name", time, 0);
```

zingale:
this is done now in #111

PR #111, opened 2017-02-25
implement a post_simulation hook

zingale:
this is the start of a generic `post_simulation()` function called at the end of the simulation.


zingale:
This now addresses #108 and #107

zingale:
Last style bit:

I put `post_simulation()` in `Prob.cpp`, but some problem setups already have their own `Prob.cpp`.  Therefore, either I need to:

* ifdef this functionality out (don't really like that)
* put it in a different file (like `Prob_post.cpp`)
* update all of the setups that have their own `Prob.cpp` to have the needed stub

maxpkatz:
Regarding the style question: until now, newly added problem hooks of this sort have been ifdef'd out with an additional preprocessor macro associated with each one. I am not necessarily particularly proud of this convention but it is the one we've got at present. I would be OK with putting it in a separate file. I am not OK with breaking compatibility by putting it on Prob.cpp without a preprocessor macro wrapped around it.

maxpkatz:
Note though that if you take the route of putting it in a different file, it has to be unique file just for that function. If we started adding things to do it in the future we would then be breaking compatibility again.  That's why there's at least some argument for the preprocessor approach.

Issue #114, opened 2017-03-14
general MOL integrator


zingale:
Implement a general method of lines integrator to compare against the CTU integrator.  This is conceptually simpler, but will have a more restrictive timestep than CTU.

Basic plan (work in progress on the `mol` branch)

* split the current `ppm()` routine into a pair: `ppm_reconstruct()` which just figures out the parabola and `ppm_int_profile()` which will integrate under the parabola for each characteristic.  The latter is not needed for MOL.
* create a `single_stage()` routine that mirrors the current `umdrv()` for CTU.  It will do the following:
  * take a current primitive variable state q
  * check the courno 
  * compute the flattening parameter
  * do ppm reconstruction
  * call the Riemann solver
  * compute the artifical viscosity
  * make the update (k in the usually RK notation) for this stage
* from C++ manage the stages of the integration:
  * construct the current U from any earlier stages (as needed) -- store this in state data as S_new
  * fill ghost cells on U
  * call `ctoprim`
  * call `single_stage()`

zingale:
Still need to sort out source terms, but there is no need for a primitive version of the source terms, so we would just need to construct the sources corresponding to the current U before calling the single_stage

Also, RK integration is probably just temporary until we put SDC integration in

zingale:
flux correction may have to be weighted by the integrator weights

zingale:
we will want `construct_old_gravity` to work from `Sborder` for the MOL stuff, not the old data (for the density).  It is not clear if this is the case.

zingale:
we might be able to get away with resetting S_old at the start of each stage's integration

this will make the gravity work without change, and still allows us to use the ghost cells filling of the state data to initialize the later stages.

This may mess up some syncs? not sure actually.

zingale:
remaining todo:
  * 1-d and 3-d PPM reconstruction needs to be split
  * 1-d and 3-d need the mol routine written
  * fluxes needed to be added with proper weighting in the registers

zingale:
we need to update `compute_cfl` to use the MOL constraint.  `estdt` was already updated

zingale:
we need to decouple the sdc iteration, the mol iteration and the retry iteration

zingale:
merged

PR #115, opened 2017-03-15
method of lines integrator

zingale:
This is the start of a method of lines integrator, following on the ideas from issue #114 

This is work in progress


zingale:
The idea for implementing this is to loop over `do_advance()`.

To support this, we will need to create several `k` MultiFabs that will hold the update for each of the stages for the Runge-Kutta 

We currently use `S_old` and `S_new` as MultiFab references to the `State_Type` `StateData`.  These have no ghost cells.  We use a separate MultiFab, `Sborder` with ghost cells to hold the initial state after reactions that we build the hydro update from.

For the MOL implementation, we'll need a new MultiFab (with ghost cells?) to cache this, `Sburn`.

Basic flow (using the sequence in the Docs flowchart)

1a. initialization: for stage 1, initialize `Sborder` from `S_old`
1b. initialization for stage > 1: initialize `S_new` with `Sburn` + dt * k# (as needed for RK state)
      FillPatch from `S_new` to `Sborder`.  Do `clean_state`

2. stage 1 only:
   * burn: for stage 1, react and store the update in `Sborder` -- this has ghost cells.
   * copy `Sborder` to `Sburn` for later updates

3. construct source terms from `Sborder`.  For MOL, skip the apply to `S_new`

4. do the hydro as usual to get the hydro source

we can skip the radial data update (move it earlier?) and the `clean_state` (steps 5 and 6)

we can skip the source correction (step 7)

for stage < last, exit -- the update is now the `k` needed for RK

8. reaction -- before we react, we need to construct `S_new` from `Sburn` and the appropriate weighting of `k`'s, and then for the final react.

PR #118, opened 2017-04-10
Exec/hydro_tests/KH/*.p5: fix problem number and filename reference

ickc:
The filename and the contents are inconsistent. Filenames are referring to problem 5, but the contents are referring to problem 4.

I tried to trace the history but it's difficult to see the history because of the reorganization in commit 9359f31. So here I made a guess that problem 5 is correct and 4 is wrong.

In case my guess is wrong and it should actually be problem 4, choose #119 instead.

### DIFF
--- a/Exec/hydro_tests/KH/inputs.2d.p5
+++ b/Exec/hydro_tests/KH/inputs.2d.p5
@@ -62,4 +62,4 @@ amr.plot_vars = ALL
 amr.derive_plot_vars = NONE
 
 # PROBIN FILENAME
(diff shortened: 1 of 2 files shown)


maxpkatz:
This is the correct one. Please point the PR to the development branch though, not master.

ickc:
Sorry, I forgot to merge to devleopment (I read that in the README...). Fixed now.

maxpkatz:
Thanks. Congrats on your first contribution to Castro!

PR #119, opened 2017-04-10
mv Exec/hydro_tests/KH/*.p5 to *.p4

ickc:
Also see #118. Choose from either #119 or #118.


maxpkatz:
Rejected in favor of #118.

PR #139, opened 2017-07-06
Smooth initial temp profile

KISSFORMISS:
And there are some parameter changes in input-det-x


zingale:
what are the units for `c`?  I don't understand what that parameter does.  Also, these should be runtime parameters that we can set via `probin`

PR #142, opened 2017-07-08
Fix the width and central position of temp transition definition

KISSFORMISS:


maxpkatz:
Can you please make the same change in Prob_2d.f90?

KISSFORMISS:
Hi, Max, it is done.

PR #148, opened 2017-08-03
unify Riemann solvers across dimensions

zingale:
This change has all the code use a single `nd` version of the Riemann solvers.  There were slight differences between the implementations in 1-, 2-, and 3-d, so some changes are expected.  I used the 3-d versions as the new implementation, so 3-d tests should be unchanged.


zingale:
note that all 3-d tests pass with this change

zingale:
Here's the full test suite run with these changes:

http://bender.astro.sunysb.edu/Castro/test-suite/test-suite-amrex-gfortran/2017-08-04/index.html

everything looks reasonable, except the retry test no longer works

PR #150, opened 2017-08-04
make the slope / ppm calculations dimensionally independent

zingale:


zingale:
still need to test this... once current test suite runs finish

zingale:
this is the test suite run for this branch

http://bender.astro.sunysb.edu/Castro/test-suite/test-suite-amrex-gfortran/2017-08-04-001/index.html

Issue #158, opened 2017-08-14
new directory structure


zingale:
It's time to further separate code into directories by topic.  This will make it easier to merge the Nyx MHD into Castro.

Some thoughts:

  * we should no longer have separate `Src_1d`, `Src_2d`, `Src_3d`, and `Src_nd` directories.  Since we are removing much of the dimensional dependence from the code, this sort of distinction will soon not be necessary

Organization proposal:

  * `Castro/Source/`
    * `diffusion/`
    * `driver/`
    * `gravity/`
    * `hydro/`
    * `mhd/`
    * `radiation/`
    * `reactions/`
    * `source_terms/`

zingale:
@WeiqunZhang @asalmgren please share any thoughts you might have on this.

asalmgren:
I agree with splitting out the files into directories.  I think it's helpful in Nyx although there's still too much in the Src directory itself.

zingale:
this has been merged

PR #159, opened 2017-08-16
reorganize the directory structures

zingale:
This groups source files by physics.  Implements #158


zingale:
might be easier to review this by just checking out the `reorg` branch and poking around there.

PR #171, opened 2017-09-08
Set a floor for timestep limiting due to changes in X

maxpkatz:
The current implementation has the problem that it can drag the timestep
down to very small values due to changes in trace isotopes. Usually for
burning simulations we do not care about changes to abundances that are
very small in absolute terms. This limiter (which defaults to 1.e-3) is
the abundance below which the timestep limiter is not applied.


zingale:
looks good to me

PR #180, opened 2017-09-15
Use a MultiFab to store the mass fluxes during a timestep

### LINKED ISSUE #178: Timestep retry scheme is not conservative
I observe a wdmerger collision case where the retry scheme is significantly non-conservative (reverses the sign of the energy on the grid in one step). I have a reproducer for this case and when I only allow the retry scheme to use a single subcycle (effectively disabling it, but still using the machinery), the scheme is still non-conservative, so there must be a fundamental issue.

maxpkatz:
The fluxes MultiFab holds this data normally when no subcycles occur.
But when subcycles occur, the fluxes MultiFab is intended to hold the
sum of the fluxes over all of the subcycles -- this is important for
getting refluxing right with AMR. But this introduces a problem because
the gravity source term references the fluxes MultiFab. A new MultiFab,
mass_fluxes, has been introduced and it stores the density hydro flux only
for a single timestep or subcycle.

Fixes #178

### DIFF
--- a/Source/driver/Castro.H
+++ b/Source/driver/Castro.H
@@ -619,6 +619,17 @@ protected:
     amrex::Array<std::unique_ptr<amrex::MultiFab> > rad_fluxes;
 #endif
 
+    //
+    // Temporary fluxes that only live for the length of the timestep.
+    //
+    amrex::Array<std::unique_ptr<amrex::MultiFab> > current_fluxes;
+#if (BL_SPACEDIM <= 2)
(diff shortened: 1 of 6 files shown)


maxpkatz:
@zingale please think about whether this needs to be integrated into MOL at this point.

maxpkatz:
I recommend running the test suite on this branch before merging.

zingale:
tests seem good

PR #181, opened 2017-09-15
Minor parameter change

adam-m-jcbs:
I had CFL issues with `convective_flame` that went away when I turned on the CFL retry parameter.  I added this to the problem setup `inputs.2d.test`.


zingale:
this may be triggered by the recent bug I introduced.  Let me fix that and then we should retry.

PR #185, opened 2017-09-17
Apply a small tolerance to the density floor when flux limiting

maxpkatz:
If the density floor used in the hydro flux limiter is the true small_dens, then this can create situations where the density actually goes below small_dens due to small numerical differences. The way to fix this is to use a density floor that is slightly above the true density floor in the flux limiting. This PR increases small_dens by 10% and uses that as the effective floor for the hydro flux limiter. This should cause many fewer density resets.


maxpkatz:
I'll look into that, but it's orthogonal to this issue.

Issue #9, opened 2016-08-14
source term tracing is only momentum terms


zingale:
we do parabolic reconstruction on the sources for all primitive state variables but only implement the source tracing for the momentum (velocity) sources

zingale:
note that `trace_nd.F90` considers the sources for all primitive variables, not just velocity (but it doesn't trace)

zingale:
when we fix this, we should eliminate `ppm_trace_sources` as an option and just leave it on as default -- it is the right way to go

zingale:
note that we will need to wrap the source term parts in the trans routines with `ppm_type`

zingale:
need to figure out how to deal with the tau source

zingale:
this has been merged

Issue #116, opened 2017-03-16
source terms for hydro prediction are not properly ghost cell filled


zingale:
in `construct_hydro_source`, there are a number of places where we do a `FillBoundary()`, for sources that will be used, for example, in tracing during the reconstruction:

```
    sources_for_hydro.FillBoundary(geom.periodicity())
    dSdt_new.FillBoundary(geom.periodicity());
```

But `FillBoundary` does not work for physical boundaries nor for coarse-fine interfaces.

We need to do a real fill here.

zingale:
I think we can use `FillDomainBoundary` here, and `FillPatchUtil`

zingale:
@WeiqunZhang is there a tutorial example that shows how we can properly fill the ghost cells for `sources_for_hydro` in `Castro::construct_hydro_source` in `Castro_hydro.cpp`?

This is not `StateData`.

WeiqunZhang:
Maybe a simple extrapolation is good enough.  To do something similar to fillpatch, we need to fill the physical boundaries.  So how do we do that?  We also need to interpolate from coarse in space and time.  To interpolate in time, we need the coarse level to have `sources_for_hydro` at the beginning and the end of the timestep.  So that means it is essentially a `StateData`.

zingale:
okay.  I think we should make it `StateData`.  Need to remind myself why we have dS/dt as `StateData` currently.

One thing we should do is have a separate `NSOURCES` variable, which is `NVAR - NSPEC`, and use that for the size, since we don't really have a use case for species sources in the Strang integrator.  We could make this an option.

zingale:
alternately... it might be possible to fill the ghost cells explicitly in the source construction by looping over them and initializing them from the `State_Type` MF

zingale:
#165 will help with making this into `StateData`

maxpkatz:
dS/dt is StateData for basically this same reason, that I didn't want to have to worry about explicitly doing ghost zone or domain boundary fills. However, it has the flaw from #154 where sometimes users override the normal fill. It was also done to preserve the predictor value on a restart. sources_for_hydro doesn't need to be saved after a timestep is done which is one reason why I did not make it StateData.

maxpkatz:
I should note though that the checkpoint issue is not a reason to avoid making it StateData. Any individual StateData can be easily excluded from being saved in the checkpoint.

maxpkatz:
[...]

PR #186, opened 2017-09-18
do source tracing on _all_ sources by default

zingale:
The changes the source tracing strategy to do source term tracing on all sources (not just momentum sources).  

~~It changes the name of the runtime parameter~~

Should probably wait until #116 is fixed


zingale:
this will fix #9

zingale:
need to fiure out a source for QTEMP for the source tracing with ppm_temp_fix = 3

zingale:
`convective_flame` crashes with this change -- need to figure out why

zingale:
I think that this is ready for review now...

PR #187, opened 2017-09-18
integrate ppm_temp_fix = 3 into the mainline

### LINKED ISSUE #166: ppm_temp_fix = 1 no longer present
In the merging of the `trace_ppm_?d.f90` routines into `trace_ppm_nd.F90`, we orphaned the custom `trace_ppm_2d.f90` that was in `Sod_stellar`.  This contains the eigensystem that works on T as a primitive variable (`ppm_temp_fix = 1`).  We should merge this back into the unified tracing routine.

zingale:
This integrates the temperature-based eigensystem into the main tracing.

Still needs testing, so don't merge.

Fixes #166


zingale:
this looks reasonable, but there could be some different approaches that could be explored before using it for production

PR #193, opened 2017-09-19
add discussion of flux limiter

### LINKED ISSUE #21: need to document the L-F flux approach for preventing density undershoots
The new scheme for protecting against density undershoots, and `castro.limit_hydro_fluxes_on_small_dens` needs to be documented.  Best to document all of the density reset stuff together

zingale:
this closes #21


maxpkatz:
The first sentence should read:

"Multi-dimensional hydrodynamic simulations often have numerical artifacts that result from sharp density gradients."

The sentence fragment reading:

"...setting the density component equal to some small, positive value..."

should be made more precise:

"...setting the density component equal to some a value just larger than the density floor, \runparam{castro.small\_dens}..."

zingale:
fixed

PR #194, opened 2017-09-20
[Development] Particle doc

taehoryu:
Notes

1. I erased some parts related to mass and cosmological variables, but I left all others assuming they were implemented in the code.
2. Please feel free to correct any part (some explanations may need to be clarified and corrected)
3. I tried to use the same font and format as other manual tex files. Please to feel free to alter/edit fonts and format in the file.
4. I added a few examples including figures, but they can be removed. 
  4-1. When I compile with the figures, it does not compile and it complains that the plot files are not found. I hope someone could help to fix this. But it compiles without \begin{figure}[h].....\end{figure}.
  4-2. Also, if the example with the figures is to be included in the manual, depending on the  configuration of the figures, the terms \textit{left}, \textit{right} and \textit{bottom} should be altered accordingly. 
5. I also provide a parameter table for the tracer particle scheme in a separate tex file (particle_parameters.txt). One needs to fill the default value for each variable. I have not tried to compile with that file, so it may cause some errors when this is included to compile.


Docs/Particles/Particles.tex
```diff
@@ -1,215 +1,149 @@
...
-\item Assign the mass of the particles onto the grid in the form of density, $\rho_{DM}$.  
-The mass of each particle is assumed to be uniformly distributed over a cube of side $\Delta x$, 
-centered at what we call the position of the particle.    We distribute the mass of each
-particle to the cells on the grid in proportion to the volume of the intersection of each cell
-with the particle's cube.   We then divide these cell values by $\Delta x^3$ so that the
-right hand side of the Poisson solve will be in units of density rather than mass.  
-Note that this is the {\it comoving} density.
+\section{Particle movement}
+%I did not remove this part just in case this is implemented in the code. But if not, please remove it.
+\noindent There are currently two different ways in which particles can be moved, random motion and motion by self-gravity.
```
zingale: It looks like these methods don't exist.  If you look in `CastroParticles.cpp` it looks like it just advects wiith the cell-centered velocity:
```
TracerPC->AdvectWithUcc(Ucc, level, dt); 
```

zingale:
Furthermore, it looks like `AMReX_TracerParticles.cpp` implements a midpoint method for the update.  Interestingly, there is also a routine that can update with the MAC velocities, which we might want to explore in the future (I _think_ that's what we do in MAESTRO)

Docs/Particles/Particles.tex
taehoryu: I got it. I was suspicious about this section, especially "motion by gravity", since I think particles are considered as massless particles in CASTRO. I will remove that part and add some more as I play with this. Then I will do the PR.

zingale:
is this redundant with the new PR?

zingale:
the new PR deprecates this one

PR #195, opened 2017-09-20
Update the sponge source term

### LINKED ISSUE #7: Sponge is not time-centered
The sponge is applied by subtracting dt \* (rho \* u)^(n+1) / sponge_timescale from the momentum. This is not time-centered; we should apply it in a predictor-corrector fashion like other source terms. As a consequence of this change, the source terms to the hydrodynamics would also "see" the sponge, making the hydro update more accurate.

### LINKED ISSUE #57: Sponge energy source term
The sponge computes the source term to the energy as

             src(UEDEN) = src(UEDEN) + (HALF * sum(snew(UMX:UMZ)**2) * rhoInv - ke_old) / dt

but this is slightly in tension with other physics where a more typical way to do this is

             src(UEDEN) = src(UEDEN) + dot_product(snew(UMX:UMZ) * rhoInv, src(UMX:UMZ))

that is, v dotted into the momentum source. I think we should consider syncing up the source term to do the latter.

maxpkatz:
- The sponge is now always time centered; the option to change this is removed.

- The energy source term is now correctly time centered.

- The energy source term is now calculated in the same way we do it for other sources.

Fixes #7 
Fixes #57


maxpkatz:
@zingale, please double check my logic for handling the old and new source terms in the same routine.

zingale:
don't we need a call to `update_sponge_params(&time);` for the corrector?

maxpkatz:
This has been corrected. The new strategy is to use a single source term function that is agnostic to whether the state is new or old, and pass in multiplicative factors that indicate whether we're doing the predictor (multiplicative factor 1.0) or the corrector (factors -0.5 and +0.5 for the old and new states, respectively). The update_sponge_params() call is now in the right place.

Incidentally, a source term formulation like this is how we need to write the sources to work for MOL. The only catch is gravity -- if we want to write the gravity source term in a time-agnostic fashion, we need to use not the hydro fluxes but rather an approximation to the fluxes in the form of edge-centered velocities.  This should give similar qualitative results to the fully conservative method but will probably be slightly less conservative in practice.

PR #199, opened 2017-09-25
merge the mol hydro routines into a dimensionless version

zingale:
This will eliminate the dimension-specific versions in favor of a dimension agnostic version.


zingale:
this is ready for review.  This is the test result for this PR:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2017-11-17-001/index.html

only 1 test fails -- the Sedov 2-d in axisymmetry, to roundoff.

PR #212, opened 2017-10-08
rename Conductivity_dir -> CONDUCTIVITY_DIR

zingale:


maxpkatz:
It almost certainly doesn't matter in this case, but can you add the backwards compatibility clause like we have for Network_dir and EOS_dir?

zingale:
done

PR #222, opened 2017-10-13
Use the correct (negative) scaling for the coarse pressure

maxpkatz:
This error was introduced sometime around 17.02 when the reflux was rewritten. It only matters for 2D and 1D.

Thanks to @WeiqunZhang for finding this.


zingale:
can you add a note to `CHANGES` referencing this PR?

PR #223, opened 2017-10-15
Move ctoprim to separate MFIter loop

maxpkatz:
The loop to do ctoprim is not slabbed like the PPM and Riemann solve are, so we get no real performance benefit from including it in the same MFIter loop. But we do take a performance hit because we are unnecessarily doing redundant work on ghost zones. This change removes the unnecessary ghost zone work, at the cost of allocating the whole array of primitive variables at one time.


zingale:
are you sure that we don't need the ghost cell loop for radiation?

maxpkatz:
What's different about radiation?

zingale:
does this mean at a C-F interface, we are now interpolating all of the quantities instead of making them thermodynamically consistent via the EOS?

maxpkatz:
It's not doing anything differently. It's just storing the work in a MultiFab rather than a series of temporary Fabs.

zingale:
do we think that this is ready to be merged?

maxpkatz:
I think it is correct but I am still uncertain whether we want to do this. So I'm going to leave it dormant for the moment.

maxpkatz:
I am now in favor of doing this. This is ready for review.

zingale:
this passed tests to roundoff

Issue #221, opened 2017-10-12
QCSML is no longer needed


zingale:
`QCSML` in the `qaux` array is not longer really needed.  We used to pre-compute the edge-centered cs, but now we do that in the Riemann solvers directly.

PR #227, opened 2017-10-16
Adressing issue #221, removed QCSML. Now, a grep doesn't show the var…

guadabsb15:
I think, no more QCSML

### DIFF
--- a/Source/hydro/riemann_nd.F90
+++ b/Source/hydro/riemann_nd.F90
@@ -288,7 +288,7 @@ contains
           clsql = gcl*pl*rl
           clsqr = gcr*pr*rr
 
-          csmall = max( max(small, small * qaux(i,j,k3d,QC)) , max(small, small * qaux(i-sx,j-sy,k3d-sz,QC)) ) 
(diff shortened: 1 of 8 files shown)


Source/hydro/riemann_nd.F90
```diff
@@ -288,7 +288,7 @@ subroutine riemanncg(ql, qr, qpd_lo, qpd_hi, &
           clsql = gcl*pl*rl
           clsqr = gcr*pr*rr
 
-          csmall = max(qaux(i,j,k3d,QCSML), qaux(i-sx,j-sy,k3d-sz,QCSML))
+          csmall = max( max(small, small * qaux(i,j,k3d,QC)) , max(small, small * qaux(i-sx,j-sy,k3d-sz,QC)) )
```
maxpkatz: This logic can be simplified, i.e.

csmall = max(small, max(small * qaux(i,j,k3d,QC), small * qaux(i-sx,j-sy,k3d-sz,QC))

Same for the two following changes.

guadabsb15: True! :) I will do that!

Issue #59, opened 2016-11-16
Pull out EOS call from react_state


maxpkatz:
Rather than calling the EOS manually in ca_react_state, we should use a call to computeTemp (or something like that) to update the temperature after a burn. This reduces code redundancy slightly but also this simplification of react_state will be helpful for some things I want to do in the future on GPUs.

maxpkatz:
This is fixed by commit 1c6b6c123e8420ae13d7a813be7bc22e9c848006

PR #230, opened 2017-10-16
use an EXTRA_THERMO flag to disable composition derivs

zingale:
this turns off the compostional derive.

Requires Microphysics https://github.com/starkiller-astro/Microphysics/pull/59 to be merged


maxpkatz:
Looks like you forgot to push the new eos_type.F90 :-)

Also can you add to CHANGES? It's remotely possible that someone was relying on these derivatives in their problem.

zingale:
sure...

PR #233, opened 2017-10-20
Particle doc (Oct19)

taehoryu:
1. The section "randomly-assigned initial positions" is removed.
2. The section "Particle movement" is removed.
3. The default values in the table in "particle_parameters" are updated.
4. Employing the cell-centered velocity and face-centered velocity to advance particles is mentioned.

### DIFF
--- a/Source/driver/_cpp_parameters
+++ b/Source/driver/_cpp_parameters
@@ -4,6 +4,7 @@
 # in the inputs file, b is the variable name in the C++ class.
 
 
+
 @namespace: castro Castro static
 
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


zingale:
did you generate the parameter table manually?  We automatically generate those for the other runtime parameters from `_cpp_parameters`.   But I think we need to move the particle parameters into this file, so we can pick the up.

Also, can you use the macros `\runparam{}` and others (see the definitions in `CastroUserGuide.tex`) to make the parameters in the prose so they are put into the index?

taehoryu:
Yes, I filled in the numbers manually. I did not know they could be automatically generated. Could you tell me some more about what should be done with this? Then I will try. 

Also, I will look at the `CastroUserGuide.tex` and try to use the macro.

zingale:
I think we should make the table generation a separate issue.  We need to move the ParmParse of the parameters out of the C++ code and into our scripts.

zingale:
I've created issue #235 to track the particle parameters

zingale:
now there is a conflict here.  I think that you need to pull from the `development` to get the updated `_cpp_parameters` file

zingale:
this branch was lost -- will reissue PR from revived branch

PR #237, opened 2017-10-20
Array to vector

WeiqunZhang:
amrex::Array is deprecated.


zingale:
I had already remove PArray from the docs, so resolved that conflict

PR #238, opened 2017-10-24
Fix a typo in the Microphysics repo urls

eftov:
Hi,

I saw a typo in the Getting Started chapter, so this fixes the Microphysics repo urls.


zingale:
thanks for catching that!

PR #241, opened 2017-11-02
Use AMReX's new MLMG linear solver

WeiqunZhang:
This pull-request is not expected to have any effects on results by default.  The new MLMG linear solver in AMReX is currently 3d only.  Thus it's only compiled in when it's 3d.  The default linear solver is still the Fortran solver in AMReX.

There are four new parameters.

- gravity.use_mlmg_solver = 0 :  This boolean flag determines if mlmg solver is used.
- gravity.mlmg_max_fmg_iter = 0 :  This integer parameter determines how many FMG cycles will be performed before switching to V-cycle.
- gravity.mlmg_agglomeration = 0 :  This boolean flag determines if AMR level 0 grids will be agglomerated as the grids are coarsen in the multi-grid hierarchy.
- gravity.mlmg_consolidation = 0 :  This boolean flag determines if grids on an AMR fine level that is in a single-level solve or the lowest AMR level of a multi-level composite solve will be consolidated as the grids are coarsen in the multi-grid hierarchy.   Here, consolidation means the grids will be moved to fewer MPI processes.

The verbosity of the solver is controlled by `gravity.v`.  We could introduce a separate runtime parameter if we want.  The solver itself is not affected by any `mg.*` parameters.


WeiqunZhang:
I turned off agglomeration and consolidation by default.   I ran some tests on Cori KNL.  It was not always a win to have them on.   They will need more works in AMReX.

PR #243, opened 2017-11-10
Take correct post-restart timestep if step was limited to hit plot interval

### LINKED ISSUE #242: Incorrect timestep limiting after a restart when using plot_per_is_exact
castro.plot_per_is_exact allows the timestep to be shortened so that a plot interval is hit exactly (052e3c5056e2526b657a2c2f28310ea62debbd49). A complication with this is that we use the last timestep to set a limiter on the current timestep: dt_new = change_max * dt_old. This issue was handled in the initial commit, by saving the last non-shortened timestep and using that in the new calculation. However, the code did not correctly handle restarts: if you restart just after a shortened timestep, the code will limit based on that shortened timestep.

This does not cause incorrect results: the code still gets to the target stop_time, it just takes longer to get there. However, it is arguably a bug since it violates our usual assumption that the result of a simulation should not depend on whether we checkpoint/restart.

maxpkatz:
If a timestep was limited to hit a plot interval, a restart would see this shortened dt when calculating the next dt, and would limit the timestep to change_max * dt. This situation was worked around during normal evolution because we do not want to have to catch back up to the longer timestep every time we stop to hit a plot interval. This commit also ensures this behavior if we're restarting after a shortened timestep.

A new file is added to the checkpoint directory that stores the previous, longer timestep in this situation. On a restart, we check if this file exists, and if so, read in that longer timestep. This fixes the problem.

Fixes #242

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,11 @@
 # changes since last release
 
+  -- As of 17.10, there is a new option castro.plot_per_is_exact. If this
+     is set to 1, timesteps will be shortened to exactly hit the time interval
+     specified by amr.plot_per. An issue with this was fixed where an incorrect
+     timestep would be taken after a restart if the previous step had been
+     shortened.
+
   -- We can now use the new multigrid solver from AMReX (implemented
      in C++) instead of the older Fortran solver.  This is enabled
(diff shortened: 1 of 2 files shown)


zingale:
can you add a note in `CHANGES`?

PR #247, opened 2017-11-13
add to _cpp_parameters

taehoryu:
add particle info to _cpp_parameters


maxpkatz:
This seems to delete some recently added parameters, not sure why.

PR #248, opened 2017-11-13
Development : update "_cpp_parameters" and "CastroParticles.cpp"

taehoryu:
update "_cpp_parameters" and "CastroParticles.cpp"

### DIFF
--- a/Source/driver/_cpp_parameters
+++ b/Source/driver/_cpp_parameters
@@ -534,7 +534,7 @@ reset_checkpoint_step        int           -1
 do_tracer_particles          int           0
 
 # whether the local densities at given positions of particles are stored in output files
-timestamp_density            int           0
+timestamp_density            int           1
 
 # whether the local temperatures at given positions of particles are stored in output files
 timestamp_temperature        int           0
@@ -548,7 +548,7 @@ particle_init_file           string        ""
 
 
 # the level of verbosity for the tracer particle (0 or 1)
-v           		       int           0
+(v, verbose)                int            0
 
 # the name of a file with new particles at restart 
 particle_restart_file        string        ""
--- a/Source/driver/_cpp_parameters
+++ b/Source/driver/_cpp_parameters
@@ -531,24 +531,16 @@ reset_checkpoint_step        int           -1
 #-----------------------------------------------------------------------------
 
 # permits tracer particle calculation to be turned on and off
-do_tracer_particles          int           0
-
-# whether the local densities at given positions of particles are stored in output files
-timestamp_density            int           1
-
-# whether the local temperatures at given positions of particles are stored in output files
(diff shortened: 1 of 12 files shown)


maxpkatz:
Can you please run Source/driver/mk_params.sh

Source/particles/CastroParticles.cpp
```diff
@@ -236,20 +200,17 @@ Castro::TimestampParticles (int ngrow)
     {
 	first = false;
 
-	ParmParse ppp("particles");
 
 	// have to do it here, not in read_particle_params, because Density, ..., are set after
 	// read_particle_params is called.
 
 	int timestamp_density = 1;
```
zingale: we also need to remove these default value settings -- these will now be done elsewhere in the code

Source/particles/CastroParticles.cpp
```diff
@@ -236,20 +200,17 @@ Castro::TimestampParticles (int ngrow)
...
-	ppp.query("timestamp_density", timestamp_density);
 	if (timestamp_density) {
 	    timestamp_indices.push_back(Density);
 	    std::cout << "Density = " << Density << std::endl;
 	}
 
 	int timestamp_temperature = 0;
```
zingale: need to remove this default

Source/driver/_cpp_parameters
```diff
@@ -525,6 +525,45 @@ reset_checkpoint_step        int           -1
...
+
+#-----------------------------------------------------------------------------
+# category: particles
+#-----------------------------------------------------------------------------
+
+# permits tracer particle calculation to be turned on and off
+do_tracer_particles          int           0
+
+# whether the local densities at given positions of particles are stored in output files
+timestamp_density            int           0
```
zingale: it looks like the old default value of this was `1` -- we should keep the defaults unchanged

zingale:
I added some comments in the source

you need to also remove the default value of `particle_verbose` in `CastroParticles.cpp`

For `v`, we should declare it such that the query uses `particle_verbose` as the cpp_name, like:
`(v, verbose)`

(search for `verbose` in the `_cpp_parameters` file to see an example

Source/particles/CastroParticles.cpp
```diff
@@ -11,7 +11,6 @@ using namespace amrex;
 
 AmrTracerParticleContainer* Castro::TracerPC =  0;
 int Castro::do_tracer_particles              =  0;
```
zingale: this should also be removed

zingale:
there's still some issues here.  I'm going to try to fix them on your fork via a PR that would then update this PR...

zingale:
okay, I confirmed that the parameters do what they are supposed to do with this change

PR #249, opened 2017-11-13
Unify sources into a single MultiFab

### LINKED ISSUE #165: consolidate the source MFs into a single one
We currently have separate MFs for each of the physics component's contributions to the source.  We should replace this with only a single MF for the source term for all physics.

Need to ensure if we can still do the Coriolis implicitly in this framework.

maxpkatz:
Source terms (gravity, rotation, etc.) have now been all coalesced into a single MultiFab. This reduces the memory footprint of the code. This negates the need for the code parameters update_state_between_sources and coalesce_update_diagnostics, so they have been removed. This will cause a change in results: the previous code behavior was to update the state with each successive source term as it was applied at the new time. Now every source term will be calculated using the same new-time state (the one coming out of the hydro) and the source terms are all applied to the state in one shot at the end.

Closes #165


CHANGES
```diff
@@ -1,5 +1,15 @@
...
 
+  -- Source terms (gravity, rotation, etc.) have now been all coalesced into
+     a single MultiFab. This reduces the memory footprint of the code. This
+     negates the need for the code parameters update_state_between_sources and
+     coalesce_update_diagnostics, so they have been removed. This will cause
+     a change in results: the previous code behavior was to update the state
+     with each successive source term as it was applied at the new time. Now
+     every source term will be calculated using the same new-time state (the one
+     coming out of the hydro) and the source terms are all applied to the state
+     in one shot at the end.
```
zingale: perhaps note that both methods are second-order accurate

Source/diffusion/Castro_diffusion.cpp
```diff
@@ -13,54 +13,57 @@ Castro::construct_old_diff_source(Real time, Real dt)
...
+    add_spec_diffusion_to_source(new_sources, SpecDiffTerm, time, 0, mult_factor);
+    add_viscous_term_to_source(new_sources, ViscousTermforMomentum, ViscousTermforEnergy, time, mult_factor);
 #endif
 
     // Time center the source term.
 
-    new_sources[diff_src]->mult(0.5);
+    mult_factor = -0.5;
+    Real old_time = time - dt;
+
```
zingale: how does this work -- we are no longer multiplying by +1/2 -- is that absorbed in the Fortran now?

Source/driver/Castro_advance.cpp
```diff
@@ -425,9 +425,7 @@ Castro::finalize_do_advance(Real time, Real dt, int amr_iteration, int amr_ncycl
 
 	dSdt_new.setVal(0.0, NUM_GROW);
 
-	for (int n = 0; n < num_src; ++n) {
-	    MultiFab::Add(dSdt_new, *new_sources[n], Xmom, Xmom, 3, 0);
-	}
+        MultiFab::Add(dSdt_new, new_sources, Xmom, Xmom, 3, 0);
```
zingale: why do we only need dSdt for the momenta? and not other quantities?

Source/hydro/Castro_hydro.cpp
```diff
@@ -23,8 +23,7 @@ Castro::construct_hydro_source(Real time, Real dt)
 
     sources_for_hydro.setVal(0.0);
 
-    for (int n = 0; n < num_src; ++n)
-	MultiFab::Add(sources_for_hydro, *old_sources[n], 0, 0, NUM_STATE, NUM_GROW);
+    MultiFab::Add(sources_for_hydro, old_sources, 0, 0, NUM_STATE, NUM_GROW);
```
zingale: since `old_sources` is already the sum, do we still need a separate `sources_for_hydro`?

Source/sources/sponge_nd.F90
```diff
@@ -37,8 +37,8 @@ subroutine ca_sponge(lo,hi,state,state_lo,state_hi,source,src_lo,src_hi, &
     real(rt)         :: state(state_lo(1):state_hi(1),state_lo(2):state_hi(2),state_lo(3):state_hi(3),NVAR)
     real(rt)         :: source(src_lo(1):src_hi(1),src_lo(2):src_hi(2),src_lo(3):src_hi(3),NVAR)
     real(rt)         :: vol(vol_lo(1):vol_hi(1),vol_lo(2):vol_hi(2),vol_lo(3):vol_hi(3))
-    real(rt)         :: dx(3), dt, time
-    real(rt), value  :: mult_factor
+    real(rt)         :: dx(3)
+    real(rt), value  :: dt, time, mult_factor
```
zingale: why are we making these `value`?  Shouldn't we be consistent with other routines?

CHANGES
maxpkatz: Done

Source/diffusion/Castro_diffusion.cpp
maxpkatz: The mult_factor is used in the Saxpy that is done at the end of add_*_diffusion_to_source; so the net result is that we get 0.5 * new_time - 0.5 * old_time, which is the same as we had before.

Source/driver/Castro_advance.cpp
maxpkatz: The comment above that line addresses this:

"Since we want (S^{n+1} - S^{n}) / dt, we only need to take twice the new-time source term, since in the predictor-corrector approach, the new-time source term is 1/2 * S^{n+1} - 1/2 * S^{n}. This is untrue in general for the non-momentum sources, but those don't appear in the hydro anyway, and for safety we'll only do this on the momentum terms."

I'm not sure what is meant by this not being the case for non-momentum sources, as that was quite a while ago.

Source/hydro/Castro_hydro.cpp
maxpkatz: Yes, because we need a separate container to hold this so that we can add dSdt. Note also that I plan to make the sources data StateData (and to have zero ghost cells), so we'll need sources_for_hydro around so we can FIllPatch into a non-zero number of ghost zones.

Source/sources/sponge_nd.F90
maxpkatz: This is part of a general code transformation I am trying to make that will aid us for GPUs: when we are passing scalar, read-only values to the GPU, it makes the code a lot simpler if we pass them by value rather than by reference (if we do the latter, then we have to create an explicit allocation on the GPU for them).

PR #253, opened 2017-11-16
Fill hydro source term ghost zones with valid data

### LINKED ISSUE #56: FillBoundary in get_react_source_prim doesn't address physical boundaries or C-F interfaces
For SDC, we compute the primitive source term that approximates a reaction source for the hydro prediction in `get_react_source_prim()`.

At the end of the construction, we do

```
react_src.FillBoundary(geom.periodicity());
```

but this doesn't address C-F interfaces or physical boundaries.  We may need to do a `FillPatch` here instead.

### LINKED ISSUE #116: source terms for hydro prediction are not properly ghost cell filled

maxpkatz:
This PR fixes #116 and fixes #56. The Source_Type StateData is co-opted to hold the actual old- and new-time sources (previously it held the source term predictor). This StateData is used to fill the sources_for_hydro MultiFab which provides the source terms used in the hydrodynamic update. Since it is StateData, this operation is done with a FillPatch. Consequently the sources_for_hydro data has meaningful data in both physical domain ghost zones and ghost zones at a coarse-fine interface (previously it only had meaningful data on fully interior ghost zones).

The source term predictor is still captured: at the beginning of a step, if we're using it, we compute the time derivative of the sources from the last step, and save dt/2 * dS/dt in sources_for_hydro; later, we'll add this timestep's old sources to it.


zingale:
please add a note to `CHANGES`

maxpkatz:
The code will run when restarting from a checkpoint made before this change, but as currently written it will not apply the source term predictor from that checkpoint (it will get zeroed out). This would only affect the first timestep after a restart from that checkpoint. Do we care enough about that to fix it?

zingale:
this also addresses #56

PR #260, opened 2017-11-23
[WIP] Rework the process for calculating the SDC non-reacted state

maxpkatz:
Instead of starting with the old state and then adding the old and new sources and the hydro source, start with the final state and subtract the contributions from reactions.

### DIFF
--- a/Source/sources/Castro_sources.cpp
+++ b/Source/sources/Castro_sources.cpp
@@ -414,19 +414,37 @@ Castro::get_react_source_prim(MultiFab& react_src, Real dt)
 
     int ng = 0;
 
-    // Carries the contribution of all non-reacting source terms.
-
-    MultiFab A(grids, dmap, NUM_STATE, ng);
-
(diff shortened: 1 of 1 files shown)


maxpkatz:
Marking this as [WIP] to start because the plotfile diff from this change is quite large so I suspect I did something wrong.

maxpkatz:
I believe the issue is that for SDC, the Reactions_Type data contains the full coupled integration including reactions and advection, whereas in non-SDC it only contains the reactions (which is what would have been necessary to make this work).

maxpkatz:
I suppose we could fix that by storing ydot_a in the Reactions_Type. It's a bit of a hack, but the Reactions_Type is kind of meaningless with SDC anyway.

maxpkatz:
Ignore that last comment, I think it doesn't make any sense.

maxpkatz:
Closing, as this did not end up being as easy as I had originally hoped.

PR #261, opened 2017-11-25
Skip the advance if we detect a CFL violation

maxpkatz:
If we've detected a CFL violation, there is no point in continuing to do the advance; we can stop immediately and return, and then abort or retry. This saves a little time when we are doing a retry, but more importantly prevents bad failure modes when we attempt an advance with CFL > 1, such as NaN's being generated in the state.

This also corrects a hole that has been present in the subcycling logic: if we detect a CFL violation while subcycling, we must abort. In the future this will be relaxed to allow for recursive subcycling.


maxpkatz:
Gated by #223

maxpkatz:
I never updated the comment, but since #223 was already merged in, this should now be good to go.

PR #262, opened 2017-11-25
Fix the source term predictor when doing a retry

maxpkatz:
This fixes an issue where the sources going into the hydro were not updated correctly during a retry. If we were using the source term predictor, we wouldn't update to account for the smaller timesteps (or the most recent data from the previous subcycled timestep).

### DIFF
--- a/Source/driver/Castro.H
+++ b/Source/driver/Castro.H
@@ -473,6 +473,8 @@ public:
 
     void computeTemp (amrex::MultiFab& State);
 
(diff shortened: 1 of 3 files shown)


maxpkatz:
This was actually a retry bug independent of the source term predictor, we were never zero'ing out the sources_for_hydro prior to each subcycled timestep, so they would get added together.

maxpkatz:
Note that this is a bug from #253, it was working correctly prior to then.

PR #265, opened 2017-11-26
Fix the calculation of the source term predictor

maxpkatz:
For the momentum sources, the calculation of dS/dt was incorrect as a result of #253, and was also incorrectly applied to all of the state components. This reverts the behavior to the previous method, where we only apply the source term predictor to the momentum and we account for the predictor-corrector nature of the source terms.


maxpkatz:
Incidentally, this issue is related to why it's non-trivial to do the source term predictor for all sources (for #186).

PR #271, opened 2017-12-03
Pass the state for the source terms as an argument to the source constructor

### LINKED ISSUE #168: old source terms should take argument of state to build from
For MOL and perhaps SDC, we will want the "old" source term function to take an argument that is the state to build from, instead of always assuming it is `S_old`

### LINKED ISSUE #268: External source term corrector incorrect when reactions enabled
Source/sources/Castro_external.cpp recalculates the old-time external source when doing the new-time corrector source. This change was made for #249; previously, we had stored the old-time ext_src data, so we used it directly. However, it uses the old-time state, S_old to do so. When using reactions, this is incorrect because the actual old-time source used Sborder, not S_old, to calculate the external source, and Sborder has had dt/2 of burning tacked on to S_old.

### LINKED ISSUE #269: Old-time diffusion source does not respect reactions Strang splitting
Most of our source terms use Sborder for constructing the old-time source, as this has had dt/2 of Strang-split burning. The old-time diffusion source does not respect this scheme, as it uses S_old for calculating the various diffusion terms.

maxpkatz:
This generalizes the algorithm so that one can choose which state MultiFab to send in for calculating the source terms. This is a step in the direction of enabling source term calculations at arbitrary times, which is useful for MOL, and also allows the old source terms to be either calculated with either S_old or Sborder (which can be useful if trying to reconstruct the source terms at a later time when Sborder is no longer available).

This changes results for diffusion if reactions are also enabled, since diffusion was previously using the unreacted S_old state, not Sborder.

Fixes #269 
Fixes #268 
Closes #168

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,8 @@
 # changes since last release
 
+  -- Minor inconsistencies in how the external and diffusion source terms
+     were constructed when simultaneously using reactions (#268, #269)
+     have been fixed (#271).
 
 # 17.12
 
--- a/Source/diffusion/Castro_diffusion.cpp
+++ b/Source/diffusion/Castro_diffusion.cpp
@@ -9,23 +9,23 @@ using std::string;
 using namespace amrex;
 
 void
-Castro::construct_old_diff_source(MultiFab& source, Real time, Real dt)
+Castro::construct_old_diff_source(MultiFab& source, MultiFab& state, Real time, Real dt)
 {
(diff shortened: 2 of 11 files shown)


zingale:
These changes produce roundoff level differences in the output:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2017-12-03-001/index.html

Issue #267, opened 2017-11-29
Simplify how we define QRHO, URHO, ...


zingale:
We have variables `QTHERM` and `NTHERM` that are used in defining `NVAR` and `QVAR`, but these do not really seem to be used or needed.  Keeping them around is confusing.

maxpkatz:
QTHERM and NTHERM are useful for bookkeeping purposes inside ca_set_method_params, but I agree they do not need to be kept as module variables outside of the lifetime of that routine.

zingale:
Currently we define `QRHO`, `QU`, ... and the conservative counterparts `URHO`, `UMX`, ... manually in `Castro_nd.F90` and set the total to `QVAR` and `NVAR`.  We also separately define `Density`, `Xmom`, ... in C++ in `Castro_setup.cpp` and rely on us doing these in the same order.  We should automate all of this.

E.g., define a text file `variables` of the form:

```
# C++ name     F90 name      C++ add-to          F90 add-to      preprocessor
Density             URHO            NUM_STATE       nvar
```

And a similar section for primitive variables, which I think are only needed in Fortran.

This file can be parsed by python to write include files for both C++ and Fortran.

zingale:
We actually do the C++ first and pass the indices into Fortran to set the Fortran versions (offsetting by 1, because Fortran is 1-based).

maxpkatz:
This is good by me. We actually want to do this for GPUs, because it helps a lot of NVAR, etc. are known at compile time.

zingale:
I wasn't originally thinking of doing this at compile time, but rather statically, but we could do it at compile time.  The issue then, I think, is that it will basically always rebuild when we type `make` since we'll force the headers to be regenerated to capture any changes in the preprocessor.  

We might be able to cache the old files so they are not overwritten unless things change, in which case `make` would continue to do nothing after a successful build.

maxpkatz:
I wasn't suggesting doing it at compile time versus statically. I was suggesting that if we do it statically, then NVAR will be known at compile time.

zingale:
okay.  Yes, I think that can be done.

zingale:
we can get rid of `ca_init_godunov_indices()` as part of this, since we will define everything together in `ca_set_method_params`

zingale:
to make everything known at compile time, we need to fix #274 first

PR #275, opened 2017-12-10
change how we set variables to index state arrays

zingale:
We now list variable indices in a plain text file `_variables`.  These are used as the index keys into state
arrays both in C++ (for conserved state) and Fortran (for all the states).  This is the first part of a clean-up of how we define variables.  This addresses part of #267.  At the moment, we run the script `set_variables.py` manually to write the Fortran routines that C++ include file.  We also do some checking to ensure that we are synced up between C++ and Fortran.

### DIFF
--- a/Source/driver/Castro.H
+++ b/Source/driver/Castro.H
@@ -566,7 +566,6 @@ public:
     static int       Shock;
 #endif
     static int       QVAR;
-    static int       QRADVAR;
     static int       NQAUX;
     static int       NQ;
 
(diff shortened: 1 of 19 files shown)


zingale:
This also eliminates `QRADVAR` -- we didn't need it, and simplifies the script that writes the method parameter stuff to group together parameters by if-def.

zingale:
all tests pass now.  This is ready for review

maxpkatz:
What is the difference between QVAR and NQ? It looks like NQ holds three extra variables, QPTOT, QREITOT, QRAD. What is special about them?

zingale:
At the moment, QVAR is used for sources, NQ is used for primitive state arrays.  Those radiation variables don't have source terms.

We could change this such that we always use QVAR for state arrays and define an NQSRC for the primitive source array.

Issue #279, opened 2018-01-14
split Riemann solvers between state and flux computation


zingale:
The approximate state Riemann solvers currently return the flux directly, and a subset of the state according to the NGDNV indices.  But for higher-order methods, we need the full interface state directly, so we should split the approximate state solvers to have the state computation as one routine and the flux computation separately.  We can then store a subset of the state if needed in the NGDNV slots, which are used in MFs later.

zingale:
This will be done on the `mol` branch.

zingale:
note: we will not change the HLL solvers -- it is not clear how to use them with high order methods.

zingale:
The easiest way to think about doing this is have `cmpflx` still return the fluxes and Godunov state, which will allow this to work for all Riemann solvers.  But have the `riemanncg` and `riemannus` solver return the interface state (indexed with `NQ`) and then a flux function accept this and return the flux and the `NGDNV` components that we normally want to store.

The one term keeping us from cleanly separating this is the flux limiter -- it is not part of the standard q state (since we don't do any reconstruction on it), so maybe we need to simply pass this one quantity out separately.

zingale:
we need to ensure that `compute_hybrid_fluxes` uses the correct indexing if we internally switch from `NGDNV` to `NQ`

PR #283, opened 2018-01-19
Split the approximate state RIemann solvers so we can access the full interface state

zingale:
This split gives us more flexibility, and is needed for 4th order methods.  This addresses #279.  Hydro is passing at the moment, but radiation and hybrid momentum still need work.


zingale:
this is ready for review.  All tests now pass with bitwise identical results.

Issue #287, opened 2018-02-02
move p div U term to sources for MOL integration


zingale:
For MOL integration, we can construct the p div U term using a cell-centered div{U} and include it in the sources MF that get passed to the single stage routine.  This will allow us to handle this term to fourth order once we do a cell-center to average transformation on the source MF.

Issue #286, opened 2018-02-02
MOL integration only needs 1 ghost cell (at most) for source terms


zingale:
There is no need for NUM_GROW ghost cells for MOL integration, since we don't do characteristic tracing.  The 4th order will need a single ghost cell (to allow for computation of the laplacian of the source), but the other MOL will not need any source ghost cells.

A potential exception is for certain boundary conditions, but I don't think we have a way of dealing with this right now anyway.

maxpkatz:
Could having NUM_GROW = 1 for MOL work?

zingale:
no, since we need NUM_GROW to be 4 for the reconstruction of the space.  But we can just but a check in there or a `do_ctu ? NUM_GROW : 1` :)

Issue #164, opened 2017-08-31
sources with MOL


zingale:
There appear to be some problems with source terms with the MOL driver.

1. It looks like we are only calling `construct_old_gravity` and `do_old_sources` for the 0th stage, so they will be inconsistent when applied to the later stages

~~2. If we fix 1, then we risk further inconsistency.  In particular, since we need to update the "old" source each stage, our predictor-corrector approach that subtracts off 1/2 S_old won't work, because we will have overwritten S_old by that of a later stage.~~

~~2a. actually, as Max reminded me yesterday, we reconstruct the old sources when we do the correction, so what we need to do is restore `Sborder` to be `Sburn`, i.e., the state that the original old sources were computed from.~~  I think that this is not the issue

zingale:
The source consolodation ( #165 ) will help with 2a above.

zingale:
it make may life easier to just write a completely separate driver for MOL along with separate routines for evaluating the sources.

zingale:
actually, we do need to change the MOL sources to be evaluated and applied as part of the RHS update if we want to implement a well-balanced scheme.

zingale:
#168 was merged, which will make this easier

zingale:
this is being addressed in the `mol` branch.  At the moment, pure hydro seems to work.  

sources don't work with `max_level` > 0.  This is likely due to the ghost cell filling of the sources -- we no longer fill the new-time-level Source_Type in StateData.  This is because we only need the "old" sources as input to the righthand side.  We should be able to fix this simply by storing the sources for the last stage in the new time slot (we need to make sure that the last stage's time is the new time).

We also should probably pass in the current stage time to `do_advance_mol()`, and this will be used for the ghost cell fills (however for RK integration, this isn't really correct, since we cannot use the stage estimates for ghost cell fills with subcycling.  SDC will get us around this).

We also don't expect to work with `castro.grav_source_type = 4` yet (but it should be straightforward to fix this).

zingale:
this seems to be fixed now for single level in `mol`

Issue #137, opened 2017-06-26
pradial weighting for MOL integration


zingale:
For the MOL solver, in `Castro_hydro.cpp`, we construct the radial pressure for flux correction as:
```
#if (BL_SPACEDIM <= 2)                                                                                                                                                         
        if (!Geometry::IsCartesian()) {                                                                                                                                        
          P_radial[mfi].plus(pradial,mfi.nodaltilebox(0),0,0,1);                                                                                                               
        }                                                                                                                                                                      
#endif
```
but this will be wrong for multiple stages.  This should have the stage weightings (`b_mol`), but also is that still physically relevant?

PR #288, opened 2018-02-04
various fixes to MOL

zingale:
This is a rewrite of the MOL driver.  In particular, it addresses:

* #287 (moving p divU to sources)
* #286 (number of ghost cells for MOL)
* #164 (treatment of sources with MOL)
* #291 (initialization as true averages)
* #137 (pradial weighting)

and prepares use for 4th order


zingale:
still needing to be done:

* converting the sources to cell-averages after filling
* converting the state to cell-averages after initialization
* fixing the time-centering of the sources -- in particular, after the update we need to make sure Source_old is at time-level n and Source_new is at time-level n+1.  We never actually fill the sources at n+1 in our update

zingale:
current status for 4th: 2-d Cartesian seems to work, but not cylindrical.

zingale:
we need to do a transverse Laplacian correction on the Godunov state used for the pressure gradient as well

zingale:
actually, the bigger issue is that the Laplacian and transverse Laplacian terms likely need to know about axisymmetry

zingale:
This is ready for review.

In particular, the treatment of sources is perhaps a little hacky.  Note that self-gravity is not expected to work yet, because we cannot at the moment compute phi for a given input state.

Source/driver/Castro.cpp
```diff
@@ -340,6 +343,14 @@ Castro::read_params ()
 	pp.add("ppm_trace_sources",ppm_trace_sources);
       }
 
+    // fourth order implies do_ctu=0
+    if (fourth_order == 1 && do_ctu == 1)
+      {
+	if (ParallelDescriptor::IOProcessor())
+	    std::cout << "WARNING: fourth_order requires do_ctu = 0.  Resetting do_ctu = 0" << std::endl;
```
maxpkatz: I would prefer this to throw an error rather than change the value of do_ctu.

zingale: fair enough... I did this mostly for when I was debugging

Source/driver/Castro.cpp
```diff
@@ -890,6 +901,32 @@ Castro::initData ()
...
+
+             const int idx = mfi.tileIndex();
+
+             ca_make_fourth_in_place(BL_TO_FORTRAN_BOX(box),
+                                     BL_TO_FORTRAN_FAB(Sborder[mfi]),
+                                     &idx);
+           }
+
+         // now copy back the averages
+         MultiFab::Copy(S_new, Sborder, 0, 0, NUM_STATE, 0);
```
maxpkatz: This should use S_new.nGrow() ghost zones.

maxpkatz: Although I guess strictly unnecessary due to the FillPatch that comes next.

maxpkatz: Maybe we can skip that next FillPatch on line 898 if we did this 4th order conversion already, since we effectively already did the FillPatch we need.

Source/driver/Castro_advance.cpp
```diff
@@ -112,7 +112,8 @@ Castro::advance (Real time,
     } else {
       for (int iter = 0; iter < MOL_STAGES; ++iter) {
 	mol_iteration = iter;
-	dt_new = do_advance(time, dt, amr_iteration, amr_ncycle);
+        // TODO: should be time or time + c_mol[iter]*dt?
```
maxpkatz: Is this question resolved?

zingale: yes.  I think it is correct now -- with the trick of making the old and new sources consistent.

Source/driver/Castro.cpp
zingale: the issue is that the `ca_make_fourth_in_place` only deals with the interior.  It cannot do all of the FAB, because ghost cells are needed for the Laplacian

Issue #296, opened 2018-02-20
Fortran interface adjustment for GPUs


maxpkatz:
The (lo, hi) that the Fortran function is operating over should always be the first two arguments to the function.

guadabsb15:
In order to make that change in the Derive_nd subroutines, there has to be some changes done in the AMReX side

maxpkatz:
Yeah... that's special. I'm handling the launching of that kernel in AMReX separately, and it's a little messy because any change to this could break other AMReX codes. Let's leave it alone for now.

PR #300, opened 2018-02-21
F interface lohi

guadabsb15:
Here is for the issue  #296


maxpkatz:
@zingale can you hit this with the test suite please

zingale:
several problems fail because we still need to change the problem_tagging_?d routines in problem setup directories

guadabsb15:
right! I totally forgot, I will modify those too

zingale:
I'll rerun the failed tests

zingale:
everything passes!

PR #307, opened 2018-03-13
MHD

guadabsb15:
So far it looks like the MHD solver works in the absence of B fields


Source/driver/_cpp_parameters
```diff
@@ -76,6 +76,10 @@ hybrid_hydro                 int           0                  y
 # 2: extrema-preserving ppm
 ppm_type                     int           1                  y
 
+#Lupe adding for MHD thing
+ppm_reference                int           1                  y             MHD
```
zingale: we'll eventually want to remove these -- they are old features

Source/mhd/Castro_advection_mhd_3d.F90
```diff
@@ -0,0 +1,905 @@
...
+              end if
+           end if
+
+           ! Define the magneto-accoustic speed from the EOS
+           ! Note: QREINT is currently just "e"
+
+
+
+           eos_state % rho = q(i,j,k,QRHO)
+           eos_state % e   = q(i,j,k, QREINT)
```
zingale: are we sure that q(i,j,k,QREINT) is not (rho e) here?

zingale: okay... it looks like it is just e (as shown below)

Source/mhd/ct_upwind.f90
```diff
@@ -0,0 +1,1312 @@
...
+			do i = lo(1), hi(1)
+		!Left State
+				!X-direction 
+				!-> Affected by Y flux
+				uL(i,j,k,UMAGX,1,1) = um(i,j,k,UMAGX,1) + dt/(3.d0*dx)*(Ez(i,j+1,k) - Ez(i,j,k))
+				uL(i,j,k,UMAGZ,1,1) = um(i,j,k,UMAGZ,1) + dt/(6.d0*dx)* &
+										((Ex(i,j+1,k+1) - Ex(i,j,k+1)) + &
+										 (Ex(i,j+1,k  ) - Ex(i,j,k)))
+				!-> Affected by Z flux
+				uL(i,j,k,UMAGX,1,2) = um(i,j,k,UMAGX,1) - dt/(3.d0*dx)*(Ey(i,j,k+1) - Ey(i,j,k))
```
zingale: shouldn't this be `uL(i,j,k,UMAGX,1,1)`, instead of `...1,2)`?

zingale: actually, I don't understand what that last index is for

zingale:
Note: our current test suite passes when run on this branch:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/

as well as the GPU test suite.

Source/driver/Derive_F.H
```diff
@@ -0,0 +1,85 @@
...
+#define _Derive_F_H_
+#include <AMReX_BLFort.H>
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#ifdef MHD
+  void ca_dermagcenx
```
maxpkatz, line 11: These will need to be converted to the new Derive interface.

Source/driver/meth_params.template
```diff
@@ -17,6 +17,10 @@ module meth_params_module
 
   ! number of ghost cells for the hyperbolic solver
   integer, parameter     :: NHYP    = 4
+#ifdef MHD
+  integer, parameter     :: MAXADV  = 5
```
zingale, line 21: I don't think `MAXADV` is used anywhere

Source/hydro/Castro_hydro.cpp
```diff
@@ -129,6 +129,12 @@ Castro::cons_to_prim_fourth(const Real time)
 
     MultiFab& S_new = get_new_data(State_Type);
 
+#ifdef MHD
+    MultiFab& Bx = get_new_data(Mag_Type_x);
+    MultiFab& By = get_new_data(Mag_Type_y);
+    MultiFab& Bz = get_new_data(Mag_Type_z);
+#endif
+
```
zingale, line 137: are these needed in Castro_hydro.cpp?

maxpkatz:
I'd prefer if we avoided checking in unclean notebooks into the main repo.

Source/hydro/Castro_hydro.cpp
guadabsb15: nope, doesn't seem like we need them there

Source/driver/meth_params.template
guadabsb15: agree, we don't need it. At least grep suggests that :)

guadabsb15:
> For the tests in the MHD directory, can you please remove the tests that are duplicates of existing problem setups like DustCollapse and wdmerger? I'm OK with you adding new inputs files to existing setups, and adding the initmag calls where necessary. That is preferable to maintaining two versions of each setup.

I think it is done

zingale (approved):

PR #310, opened 2018-03-15
SDC infrastructure

zingale:
This is the start of the SDC integration


zingale:
as of 8387a4e523f81b6007ffab402e79a81a474cc59f, the SDC (pure hydro) converges as 4th order

zingale:
This is ready for review.  I'll run the test suite on it.  This adds most of the infrastructure needed for the new SDC integration.

maxpkatz:
It is a little confusing to have both the castro.fourth_order and castro.sdc_order = 4 options. I suggest that we use sdc_order instead of fourth_order inside do_advance_sdc, and then replace fourth_order with an mol_order parameter, which will control the behavior inside do_advance_mol.

Source/driver/Castro_advance.cpp
```diff
@@ -565,6 +597,202 @@ Castro::do_advance_mol (Real time,
...
+  // we just did the update, so now recompute the "old" reactive
+  // source for the next SDC iteration.  We don't need to do this for
+  // m = 0, since that state never changes.
+
+  for (int m = 1; m < SDC_NODES; ++m) {
+    // use a temporary storage
+    // TODO: do we need a clean state here?
+    MultiFab::Copy(S_new, *(k_new[m]), 0, 0, S_new.nComp(), 0);
+    expand_state(Sborder, cur_time, -1, Sborder.nGrow());
+    construct_old_react_source(Sborder, *(R_old[m]));
```
maxpkatz: Since we already did this for the `sdc_iteration == 0` case in the main loop above, don't we only need to apply this step for `sdc_iteration > 0`?

Source/driver/Castro_advance.cpp
```diff
@@ -565,6 +597,202 @@ Castro::do_advance_mol (Real time,
...
+  }
+#endif
+
+  // store the new solution
+  MultiFab::Copy(S_new, *(k_new[SDC_NODES-1]), 0, 0, S_new.nComp(), 0);
+
+
+  // I think this bit only needs to be done for the last iteration...
```
maxpkatz: I agree that we don't care about the source terms being at the correct time level within the SDC iterations, it only matters that they're consistent at the end.

Source/driver/Castro_sdc.cpp
```diff
@@ -0,0 +1,175 @@
...
+    // for 4th order reacting flow, we need to create the "source" C
+    // as averages and then convert it to cell centers.  The cell-center
+    // version needs to have 2 ghost cells
+    for (MFIter mfi(*k_new[0]); mfi.isValid(); ++mfi) {
+
+      const Box& bx = mfi.tilebox();
+
+      ca_sdc_compute_C4(BL_TO_FORTRAN_BOX(bx),
+                        BL_TO_FORTRAN_3D((*A_new[m_start])[mfi]),
+                        BL_TO_FORTRAN_3D((*A_old[0])[mfi]),
```
maxpkatz: One thing to consider here is that if we want to consider extending this to arbitrary order in the future, this method doesn't extend well. We could consider defining a single MultiFab that has SDC_NODES components. The cost is that we spend a little more memory since we can't alias the `m == 0` components.

zingale: I have no plans to go beyond 4th order.  I don't even what to think about what the spatial stencils look like

Source/driver/Castro_advance.cpp
zingale: yes, I think you are right

zingale: no... wait.  This is different.  Up above, we constructed the initial reactive source term that will be used for the integration over the temporal nodes during the update from SDC iteration 0 -> 1.  Now we are getting ready for the next SDC iteration, and we have new states on the temporal nodes m > 0, so we need to store the burn term for this state, which will be the initial burn for the next iteration.

Now... it may be that we don't need to do this extra step for the very _last_ SDC iteration.

zingale:
> It is a little confusing to have both the castro.fourth_order and castro.sdc_order = 4 options. I suggest that we use sdc_order instead of fourth_order inside do_advance_sdc, and then replace fourth_order with an mol_order parameter, which will control the behavior inside do_advance_mol.

yes, I agree.  I will think about how to best clean this up.

maxpkatz:
To prevent Castro_advance.cpp from getting too long, I recommend that do_advance_sdc go into a new file, Castro_advance_sdc.cpp. I will make the same change on development, moving advance_mol to a new file, and also splitting out the CTU advance into a new file, and renaming it to do_advance_ctu.

zingale:
> To prevent Castro_advance.cpp from getting too long, I recommend that do_advance_sdc go into a new file, Castro_advance_sdc.cpp. I will make the same change on development, moving advance_mol to a new file, and also splitting out the CTU advance into a new file, and renaming it to do_advance_ctu.

yes, I agree.

Source/driver/Castro_sdc.cpp
zingale: going to defer this for later

PR #314, opened 2018-03-30
revise how runtime parameters are output to job_info

### LINKED ISSUE #313: have job_info indicate which parameters are changed from default
We do this in Maestro, and should implement it in Castro for all the parameters in `_cpp_parameters`

zingale:
This now outputs the runtime parameters like Maestro, showing all of the known parameters
and indicating with an `*` those that are changed from their defaults.

This closes #313


zingale:
turns out I cannot access the runtime parameters in the Gravity or Diffusion class when outputting the job_info stuff.  I can either make a separate job_info output function in each class, or I can make these variables not protected.

Suggestions?

maxpkatz:
If they're not protected in Castro, I guess they don't need to be protected in Gravity or Diffusion.

In both cases it would probably be better if we had a method automatically generated for every parameter.

maxpkatz:
No, I mean for every parameter, a method that obtains the value of that parameter.

zingale:
this is ready for review.  I went with a single function in each namespace that does the outputting.

zingale:
this passes all tests.  Okay to merge?

Source/driver/Castro_io.cpp
```diff
@@ -878,9 +878,21 @@ Castro::writeJobInfo (const std::string& dir)
   jobInfoFile << " Inputs File Parameters\n";
   jobInfoFile << PrettyLine;
 
-  ParmParse::dumpTable(jobInfoFile, true);
+  //ParmParse::dumpTable(jobInfoFile, true);
```
maxpkatz: What's going on here?

zingale: oh, I can remove that.  That was the old way of outputting the params to job_info, but it only worked for those that were set in inputs.  This works now for everything that was defined in _cpp_parameters

PR #315, opened 2018-04-03
Mlmg diffusion

WeiqunZhang:
This pull request adds the option to use MLMG as diffusion operator.  Two new runtime parameters are added. 

- diffusion.use_mlmg_solver = 0
  This is a boolean flag.
- diffusion.mlmg_maxorder = 4
  The other valid value is 3.  This parameter controls the order of polynomial interpolation at boundaries.


zingale:
This doesn't compile for 1-d, e.g., in `Exec/science/flame`

WeiqunZhang:
should compile now

PR #318, opened 2018-04-13
Ecp/two moment

kenshen04:
Edits in two_moment_nd to use InitThornado_Patch arrays instead of temporary arrays


zingale:
you should use `bl_allocate` and the memory pool instead of the Fortran `allocate`

kenshen04:
Done (I think...)

Issue #320, opened 2018-04-19
make mlmg the default


zingale:
Can we switch the multigrid to mlmg for both gravity and diffusion by default?

Is there any reason not to?

Related, can we rip out the old stuff?

maxpkatz:
Do we have a recent test suite run demonstrating the diffs?

maxpkatz:
@WeiqunZhang are all geometries in 2D and 3D supported now by MLMG?

zingale:
let me see if I can generate one

I did it for the diffusion, and the differences were minimal.

WeiqunZhang:
Yes, all geometries supported by AMReX (i.e., Cartesian, 1d spherical and 2d cylindrical) are supported by MLMG.

maxpkatz:
I have no objection to switching to MLMG and ripping out the old stuff if the diffs are negligible.

zingale:
diffusion_test now does axisymmetric, and the new solver seems to agree with the old one (and the analytic solution)

So still not sure what's up with the convective_flame.  My new guess is that perhaps they bring things to the edges differently, and it is the `diffuse_cutoff_density` that is treated different.

zingale:
Here's the difference in the density state using the default `inputs.2d.testsuite`

![diffs_density](https://user-images.githubusercontent.com/7817509/39337704-5daf619a-498c-11e8-9dfa-d413b2d462cb.png)

zingale:
those lines are on grid boundaries.

zingale:
and here it is if I don't use the diffusion cutoff, `castro.diffuse_cutoff_density`.

![diffs_density_nocutoff](https://user-images.githubusercontent.com/7817509/39337770-c63d46aa-498c-11e8-82de-227bad14f253.png)

zingale:
this is a log scale, so clearly the agreement is much better without the diffusion cutoff.  Does the new method bring the conductivity to the interfaces differently than the only scheme?  In particular, what happens if the conductivity is zero?

@WeiqunZhang ^

WeiqunZhang:
It shouldn't matter if the conductivity is zero.  I suspect it has something to do with metric terms.  Could you try to do `cc(i,j) = cc(i,j) * (1.d0/r)` in ca_unweight_cc ?

zingale:
instead of 
```
cc(i,j) = cc(i,j) / r
```
?

WeiqunZhang:
Yes

zingale:
[...]

PR #326, opened 2018-05-01
Mlmg defaults

zingale:
This uses the MLMG solver by default for both diffusion and gravity.  Testing showed roundoff-level differences for gravity, but larger differences with diffusion and axisymmetric.  Tests with the `diffusion_test` problem showed that the old diffusion solver did not work correctly with multilevel and axisymmetry.  See #320 for some discussion.


maxpkatz:
Can you please update the CHANGES in development?

PR #335, opened 2018-05-31
Fix logic for dealing with CFL violations during retries

### LINKED ISSUE #334: CFL failure issues during retries
In order to use retries, you have to set castro.hard_cfl_limit = 0. This allows you to not abort when a CFL violation is detected before the hydro; the hydro step is skipped, and we retry with smaller timesteps.

If a CFL violation still occurs during one of the retries, this same strategy causes problems, because we haven't yet implemented recursive subcycling (#179). It will basically lead to nonsense evolution. So we can either (a) crash at this point or (b) allow the step to be finished, potentially with unstable results.

maxpkatz:
castro.hard_cfl_limit now has the behavior that it is ignored during the initial advance if you're using a retry, and is enforced during the retries. Thus the default behavior is now that when a CFL violation is detected and retries are enabled, a retry will be taken, and if a subcycle in that retry violates CFL stability, the run aborts.

Fixes #334

### DIFF
--- a/Source/driver/Castro_advance.cpp
+++ b/Source/driver/Castro_advance.cpp
@@ -118,7 +118,7 @@ Castro::advance (Real time,
 
     // Optionally kill the job at this point, if we've detected a violation.
 
-    if (cfl_violation && hard_cfl_limit)
+    if (cfl_violation && hard_cfl_limit && !use_retry)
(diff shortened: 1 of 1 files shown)


maxpkatz:
This also fixes incorrect evolution that would occur during retries if you do set castro.hard_cfl_limit = 0. Now the evolution will be consistent.

PR #336, opened 2018-06-01
Clean state

zingale:
This allows for expand_state to work on state data directly and has the effect that we do a clean_state
before the FillPatch


zingale:
There are some differences here, since we swap the order of the `FillPatch` and the `clean_state`, but I think those are to be expected.

zingale:
Here's the tests for this branch

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2018-06-01-002/index.html

this is ready for review

maxpkatz:
I am fine with this change in general. Still need to go through in more detail and make sure I agree with all the diffs.

zingale:
that's fine, let's wait until after the proposal

zingale:
I'd like to bump this up.  This is a blocker on the SDC stuff now, so I'd like to get this merged.  I can rerun the test suite on this if desired.

Source/driver/Castro.cpp
```diff
@@ -3551,40 +3573,59 @@ Castro::clean_state(MultiFab& state) {
...
+
+  // this is the "preferred" clean_state interface -- it will work
+  // directly on the StateData.  is_new=0 means the old data is used,
+  // is_new=1 means the new data is used.
 
-    // Enforce a minimum density.
 
 #ifndef AMREX_USE_CUDA
+    MultiFab& state = is_new == 1 ? get_new_data(State_Type) : get_old_data(State_Type);
     Real frac_change = enforce_min_density(state_old, state);
```
maxpkatz: Is this call supposed to be in here? I would expect it to be in the actual clean_state interface below.

Source/sources/Castro_sources.cpp
```diff
@@ -14,8 +14,6 @@ Castro::apply_source_to_state(MultiFab& state, MultiFab& source, Real dt, int ng
     AMREX_ASSERT(state.nGrow() >= ng);
 
     MultiFab::Saxpy(state, dt, source, 0, 0, NUM_STATE, ng);
-
-    clean_state(state);
```
maxpkatz: Since it looks like every call to apply_source_to_state still gets an explicit clean_state call after it, can we put the clean_state call back here and simplify the code a little?

Source/driver/Castro.cpp
zingale: I think I put this here only for the frac_change bits.  I need to know which state I am enforcing the minimum density on.

Source/sources/Castro_sources.cpp
zingale: the reason I took it out is because eventually I want to get rid of the `clean_state()` that takes a MF in favor of the version that works directly on state date.  I got rid of all but one of those, but hope to remove the last in the future.

Source/driver/Castro.cpp
zingale: oh... I think this is a merge issue.  Let me fix

zingale: right, the call is not supposed to be there.  Nor do I need the `AMREX_USE_CUDA` logic there, it belongs in the actual `clean_state`.  Fixed now.

zingale:
Latest test run is here:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2018-06-26-004/index.html

I think I'm okay with this level of diffs

convective_flame seems to be the largest, but the main state variables are still okay there.

zingale:
okay to merge?

Source/sources/Castro_sources.cpp
maxpkatz: But can we put it back in, just using the new interface?

zingale: there was a reason why I didn't think so, but let me look again...

zingale: oh... so I'd need to change the interface of the source stuff to pass in an `is_new` like variable, I guess that's doable.

zingale: okay... I remember now... I need to know which bin it is in the StateData for the 4th order stuff, hence the change of the interface.  But if I pass that info into apply_source_to_state, we are fine.  So I'll make that change

PR #338, opened 2018-06-03
Remove dependence on F_BaseLib

maxpkatz:


maxpkatz:
Closed because I did this wrong.

PR #344, opened 2018-06-12
Implement dynamic retry capability

### LINKED ISSUE #179: Timestep retry scheme should be fully recursive
The timestep retry scheme detects if a timestep constraint was violated, and then retries with a fixed number of cycles that satisfy that constraint. For the most part, this is OK -- the underlying equations are well enough behaved that shrinking the timestep should cause them to satisfy the criterion in the way one would expect. However, sometimes something absurd happens like the velocity increasing by 10 orders of magnitude, and taking 10 billion subcycled timesteps is not the right answer -- perhaps the timestep needs to be a little smaller but not that much smaller to be well behaved. Also, there are cases where there is just no clear guidance on how to generate a safer timestep -- for example, if a negative density is generated in a step.

A solution to this is a recursive timestep scheme, where we shrink the timestep by some factor, take an advance, and then check the stability criterion again. We continue in this way until we've found the timestep that satisfies all of the constraints, and then subcycle from there until we are done.

maxpkatz:
Retries can now be done in a dynamic fashion. An advance is itself a subcycled advance always, and we keep subcycling until we are done with the step. By default we still do only a single timestep, but this helps in cases where we have to reject the timestep for (say) a CFL violation, and then during the retry the CFL criterion is violated again. In the past, we would simply have to abort the run if this happened. Now we can cut the timestep again and keep going.

Closes #179


maxpkatz:
Also new in this commit: if a burn fails, we will reject and retry the timestep.

zingale:
can you add a note to `CHANGES`?

maxpkatz:
I believe this is ready for review; I have been testing it for a few weeks on the Detonation problem which hits lots of retries, and it seems to be fairly robust.

zingale:
okay.  I'll rerun the suite on it and then review

maxpkatz:
Test suite results: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2018-07-02-001/index.html

These are all roundoff and caused by explicitly resetting the StateData time levels. I see no reason to fix this, so I'm content with this diff.

PR #345, opened 2018-06-12
Convert cm to m for Thornado

kenshen04:


zingale:
it looks like you are assuming that xmin = 0 in creating these coordinates -- is that right?  You might want to offset by prob_lo

PR #346, opened 2018-06-19
make the state vector sizes known at compile time

### LINKED ISSUE #274: make ngroups known at compile time
For GPUs, it would help a lot if we know `QVAR`, `NQ`, ... at compile time.  We need to restructure the code to make ngroups known at compile time.

zingale:
This is needed for GPUing.  Addresses #267 

Still need to fix #274 before the merge

At the moment, everything passes the regression test suite except radiation.


zingale:
This now fixes #274

zingale:
this is ready for review

Issue #350, opened 2018-06-28
remove source terms for species


zingale:
In general, we don't need to have the ability to have source terms for the species.  We can make it a compile-time option, off by default, and this will save a lot of memory, and work (tracing).  This can be part of #256

maxpkatz:
In theory we could do the same thing with other source terms (density, temperature).

zingale:
true, but I would like to keep the source array contiguous.  If I move "shock" to be before the species stuff, then I can just lop them off and save a lot of space.

Then I can define a `NSRC` and `NQSRC` to dimension these

zingale:
although, I suppose this is what indirect indexing is for...

zingale:
actually, maybe the thing to do is to indicate in `_variables` if it should be in the source array

zingale:
for the old SDC method, we need to continue to have sources for the species

zingale:
the current version of this is on `species_sources`

zingale:
it might be easier to just start with the primitive variable sources, and not worry about the conservative state sources.

For the primitive sources, we need to

- [ ] define a compile-time parameter PRIM_SPECIES_HAVE_SOURCES that is used to determine whether the primitive variable state species have sources.  We will default it to FALSE until SDC is set to TRUE.
- [ ] make NQSRC a new parameter that is determined at compile time.  This will be used to dimension the primitive source arrays
- [ ] change srctoprim to skip the species sources if we are not using them
- [ ] change the dimensions of Ip_src and Im_src and have the bounds of the loop that fills them be over NQSRC
- [ ] change the trace routines to only add the species sources if they exist

zingale:
doing this on `no_species_sources`

zingale:
With #519, this is essentially implemented.

PR #352, opened 2018-06-30
[WIP] Species sources

zingale:
This by default turns off hydro sources for species -- we don't need them, with the exception of the old SDC.


zingale:
still need to define `NSRC` and `NQSRC` in C++

zingale:
this is part of #350

zingale:
this has large errors with some tests

the issue is that we cannot simply turn off sources for passively advected quantities, since those are more than just the sources.  So we need to be careful with this logic.

zingale:
this is too much to do in one PR.  There are a lot of unexplained failures (some seemly related to just changing the order of the conserved variables in the state vector).

zingale:
it might be easier to just start with the primitive variable sources, and not worry about the conservative state sources.

For the primitive sources, we need to

- [ ] define a compile-time parameter PRIM_SPECIES_HAVE_SOURCES that is used to determine whether the primitive variable state species have sources.  We will default it to FALSE until SDC is set to TRUE.
- [ ] make NQSRC a new parameter that is determined at compile time.  This will be used to dimension the primitive source arrays
- [ ] change srctoprim to skip the species sources if we are not using them
- [ ] change the dimensions of Ip_src and Im_src and have the bounds of the loop that fills them be over NQSRC
- [ ] change the trace routines to only add the species sources if they exist

PR #361, opened 2018-07-06
Fix incorrect flux register logic when using retries

### LINKED ISSUE #357: Reflux is incorrect when using retries
The reflux scheme is incorrect when using retries, because it does not clear the contribution to the flux register when a timestep is rejected and retried.

maxpkatz:
The fluxes were not properly zeroed out when doing a retry. This led to incorrect refluxes.

Fixes #357


maxpkatz:
Gated by #360

Issue #211, opened 2017-10-05
reassess parameter defaults


zingale:
The default behavior of Castro should be to use the values of the runtime parameters that give the best accuracy or performance.  Some default values have not been updated in a long time, despite code improvements.  We should propose parameter changes in this issue.

maxpkatz:
**Parameter changes I think are unambiguously good:**

- castro.use_retry = 1

Retries are a great way to avoid CFL violations that result in the simulation dying. Most newer users do not even realize this is a possible failure mode until it hits them deep into a simulation and it causes significant disruption.

- castro.retry_tolerance = 0.02

A small tolerance should be applied so that we do not retry for very small violations of the stability criterion.

- castro.hard_cfl_limit = 0

This should be done in conjunction with the retry default, or else we'll never even get to the retry because we crash immediately upon detecting a CFL violation.

- castro.source_term_predictor = 1

Doing a lagged prediction of the source terms in the hydro should always be more accurate.

- castro.use_custom_knapsack_weights = 1

This is the sort of performance option that the user should never have objection to -- in general, we should do performance improvements for the user unless we think there is any reason they would not want this.


**Changes I think are very good but have at least some reason for caution:**

- castro.limit_fluxes_on_small_dens = 1

It's always nice to avoid density resets; density resets are quite bad and unpredictable. The reason for caution here is that the flux limiter can also (slightly) modify the flow in smooth regions that don't quite get to a reset.



Changes I think are good but are somewhat a matter of personal choice:

- castro.state_interp_order = 0

We have seen that the linear state interpolation can have problems keeping sum(X) = 1 during a fillpatch. It is not good for users to crash for a reason like this during an AMR run.

- castro.update_state_between_sources = 0

The default now is that after every source term is calculated at the new time, we update the state immediately before calculating the next source term. This results in an inconsistency where different source terms see a different new-time state, and also involves an EOS call before every extra source term.

zingale:
Isn't the update_state_between_sources going to be obsolete by #165?

maxpkatz:
Yes, that's true.

maxpkatz:
For castro.source_term_predictor we should probably wait until we fix #154

zingale:
we should change the default of 

```dual_energy_update_E_from_e``` to 0

maxpkatz:
I'd prefer just to remove it entirely. I don't think it's a good option.

zingale:
dual_energy_update_E_from_e has been removed

[...]

PR #368, opened 2018-07-13
remove dual_energy_update_E_from_e -- it is never a good idea

zingale:
this completely removes dual_energy_update_E_from_e.  We have other dual energy mechanisms to
handle this, and doing it always is not a good idea.

This also override eos_input_is_constant from StarKiller to be true by default

Note: some answers will change, so I will run the test suite.


zingale:
This connects to #211

zingale:
test results are here:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2018-07-13-001/index.html

PR #376, opened 2018-08-06
Check whether there are any zones to burn before entering react_state()

maxpkatz:
If the user has set react_{rho,T}_{min,max} then we can save some time before the burning has started in the simulation by checking to see if any of the zones meet the minimum or maximum criteria for burning. If none do, we can skip the call to react_state(). This ends up saving some cycles: for wdmerger, I measure that the cost of doing this check is very small compared to the time spent in react_state() even in the case where none of the zones end up being OK to burn, so it is worth always doing if the limiters are on. To get a sense of how much this helps, I ran a 3-level wdmerger collision job with 128 processors, and a coarse timestep is about 20% faster at the beginning when no burning is occurring.

### DIFF
--- a/Source/reactions/Castro_react.cpp
+++ b/Source/reactions/Castro_react.cpp
@@ -497,13 +497,10 @@ Castro::valid_zones_to_burn(MultiFab& State)
     // Finally check on whether min <= rho <= max
     // and min <= T <= max. The defaults are small
     // and large respectively, so if the limiters
-    // are not on,  these checks will not be triggered.
+    // are not on, these checks will not be triggered.
 
-    if (large_dens >= react_rho_min && small_dens <= react_rho_max) {
-        return true;
-    }
-
-    if (large_T >= react_T_min && small_T <= react_T_max) {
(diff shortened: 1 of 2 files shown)


maxpkatz:
Please run the test suite on this.

zingale:
this passes all tests

Source/reactions/Castro_react.cpp
```diff
@@ -387,3 +395,123 @@ Castro::react_state(Real time, Real dt)
...
+
+bool
+Castro::valid_zones_to_burn(MultiFab& State)
+{
+
+    // The default values of the limiters are 0 and 1.e200, respectively.
+
+    Real small = 1.e-10;
+    Real large = 1.e199;
+
```
zingale: this seems a little fragile

maxpkatz: Yeah, I can do better than this if we record what the default values in _cpp_parameters are.

maxpkatz:
Yes, this logic doesn't depend on whether we're tiling the reactions or not.

PR #377, opened 2018-08-06
Skip the initial EOS call prior to starting the burn

maxpkatz:
Since we always call clean_state after updating the state, T should already be consistent with e, so this EOS call is unnecessary. This requires the deletion of the dual_energy_eta3 parameter.


maxpkatz:
We should test this by looking at how much answers change between development and this branch on the test suite. The differences should decrease as a function of the Helmholtz EOS tolerance.

maxpkatz:
Note that #376 is still worth doing, even if this is accepted, as there's still some logistical work that we do to set up the burn that can be avoided. Measuring this on a small simulation on bender, only about half of the speedup from #376 comes from avoiding the EOS call.

zingale:
test output is here:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2018-08-06-001/index.html

PR #381, opened 2018-08-09
Source tracing

### LINKED ISSUE #256: don't trace under sources that are zero
when we fix #9, we will treat all sources consistently in hydro.  But since we have a source for every state variable, there will be a lot of needless work for equations that have no sources (e.g. density, usually, and species).  So we should check if the norm0 of the source is 0, and if so, skip the ppm reconstruction and tracing of the sources.  This will save on flops.

It is better to do it this way that to just assume certain sources are zero, since the user can have time-dependent external sources.  So we need to check each step.

zingale:
only trace under sources if they are non-zero.  Otherwise, we are just doing unnecessary work.
closes #256

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,8 @@
 # changes since last release
 
+  -- we now only trace under sources that are non-zero, to save
+     computational expense.  #381
+
   -- The parameter dual_energy_update_E_from_e has been removed,
      and the default behavior is now that the total energy (E)
      will not be changed when we reset the internal energy (e).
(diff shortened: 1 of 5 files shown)


maxpkatz:
This PR failed the flame_wave regression test. I am not sure why yet, but I do not that the diff is zero if diffusion is turned off.

maxpkatz:
It is due to ghost zones. The PPM reconstruction uses information from outside [lo, hi] but we're only testing on the source being zero inside that region. If the source is non-zero in the ghost zones, then this will get it wrong.

zingale:
For the record, here's the reason this arises with `flame_wave`.  I use a diffusion cutoff, that below some density, zeros out the conductivity to turn off diffusion (this is needed because of the timestep limit in explicit diffusion).  But this makes the source term discontinuous.  I _think_ we'll still have an issue even if I check in one ghost cell, but that's okay.

PR #386, opened 2018-08-15
use the names of the state component indices when setting up the BCs

### LINKED ISSUE #379: State_Type order fixed in Castro_setup.cpp
Even through we have the new `_variables` system for defining variables in Castro (and making it all known at runtime), we still are defining the names of the variables manually in `Castro_setup.cpp`, e.g., as:

```
#ifdef SHOCK_VAR
  cnt++; set_scalar_bc(bc,phys_bc); bcs[cnt] = bc; name[cnt] = "Shock";
#endif
```
We should already know the total count (`cnt`) at this point, and we can define the name and BC via the `Shock` index directly.

This would allow us to shuffle the order of the indices.

zingale:
this makes sure we are not dependent on the ordering.
Closes #379

### DIFF
--- a/Source/driver/Castro_setup.cpp
+++ b/Source/driver/Castro_setup.cpp
@@ -374,25 +374,55 @@ Castro::variableSetUp ()
   Vector<std::string> name(NUM_STATE);
 
   BCRec bc;
-  cnt = 0;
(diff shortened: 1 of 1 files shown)


zingale:
running this through the test suite now

zingale:
is there any reason _not_ to merge this?

PR #396, opened 2018-09-17
Sponge

harpolea:
Offloads the sponge routine to the gpu

### DIFF
--- a/Source/sources/sponge_nd.F90
+++ b/Source/sources/sponge_nd.F90
@@ -4,12 +4,12 @@ module sponge_module
 
   implicit none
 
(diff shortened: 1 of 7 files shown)


Source/sources/sponge_nd.F90
```diff
@@ -53,7 +62,9 @@ subroutine ca_sponge(lo,hi,state,state_lo,state_hi,source,src_lo,src_hi, &
     real(rt) :: src(NVAR)
     real(rt) :: local_state(NVAR)
 
-    src = ZERO
+    src(:) = 0.0d0
```
maxpkatz: Can you make this ZERO again please, not 0.0d0

PR #408, opened 2018-09-21
no slab version of CTU

zingale:
This is a rewrite of the CTU solver to not use slabbing to ease the offloading to GPUs.  Not ready for merging.


zingale:
this is ready for review.  Here's the test suite run:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2018-10-17-002/index.html

note: 3d pure hydro can be about 10% slower, but this will more than be made up on GPUs, which was the point of this reorganization.  I think I can get back about 1/2 of that performance drop on CPUs with some tweaking that I'll play with now.

Source/driver/check_for_nan.f90
```diff
@@ -0,0 +1,29 @@
+module nan_check
```
maxpkatz: This already exists in AMReX, do we need to duplicate it in Castro?

zingale: oh, guess not.  Where is it in AMReX?  I was probably going to move it out anyway one I am happy with the debugging, but it is invaluable for debugging.

maxpkatz: It's in the FArrayBox class, it's implemented in C++ (contains_nan).

zingale: right, but this is not a FAB at this point.  This is called from within the Fortran itself, when it is just an array.  That's where I needed it.

Source/hydro/ppm_nd.F90
```diff
@@ -145,39 +143,39 @@ subroutine ppm_type1(s, s_lo, s_hi, ncomp, n, &
...
 
-    if (s_lo(1) .gt. ilo1-3 .or. s_hi(1) .lt. ihi1+3) then 
+    if (s_lo(1) .gt. lo(1)-3 .or. s_hi(1) .lt. hi(1)+3) then
          call amrex_error("Need more ghost cells on array in ppm_type1")
     end if
 
 #if (AMREX_SPACEDIM >= 2)
-    if (s_lo(2) .gt. ilo2-3 .or. s_hi(2) .lt. ihi2+3) then
+    if (s_lo(2) .gt. lo(2)-3 .or. s_hi(2) .lt. hi(2)+3) then
          call amrex_error("Need more ghost cells on array in ppm_type1")
```
maxpkatz: I suppose we should now also add an error check for the z dimension

Source/hydro/ppm_nd.F90
```diff
@@ -615,108 +633,112 @@ subroutine ppm_type2(s, s_lo, s_hi, ncomp, n, &
...
-          !
-          if ((sedge(i,j)-s(i,j-1,k3d,n))*(s(i,j,k3d,n)-sedge(i,j)) .lt. ZERO) then
-             D2  = THREE*(s(i,j-1,k3d,n)-TWO*sedge(i,j)+s(i,j,k3d,n))
-             D2L = s(i,j-2,k3d,n)-TWO*s(i,j-1,k3d,n)+s(i,j,k3d,n)
-             D2R = s(i,j-1,k3d,n)-TWO*s(i,j,k3d,n)+s(i,j+1,k3d,n)
-             sgn = sign(ONE,D2)
-             D2LIM = sgn*max(min(C*sgn*D2L,C*sgn*D2R,sgn*D2),ZERO)
-             sedge(i,j) = HALF*(s(i,j-1,k3d,n)+s(i,j,k3d,n)) - SIXTH*D2LIM
-          end if
+    do k = lo(3)-1, hi(3)+1
```
maxpkatz: Should this block and the next one have dg(3) guards on the k loop?

Source/hydro/riemann_nd.F90
```diff
@@ -60,20 +60,13 @@ subroutine cmpflx(qm, qp, qpd_lo, qpd_hi, &
     integer, intent(in) :: s_lo(3), s_hi(3)
 
     integer, intent(in) :: idir
-    ! note: ilo, ihi, jlo, jhi are not necessarily the limits of the
-    ! valid (no ghost cells) domain, but could be hi+1 in some
-    ! dimensions.  We rely on the caller to specific the interfaces
-    ! over which to solve the Riemann problems
-    integer, intent(in) :: ilo, ihi, jlo, jhi, kc, kflux, k3d
+    ! note: lo, hi necessarily the limits of the valid (no ghost
```
maxpkatz: This comment may become confusing later because we're already on a trend towards it being a general rule that the (lo, hi) to loop over are not necessarily the valid domain.

Source/hydro/riemann_nd.F90
```diff
@@ -370,47 +369,55 @@ subroutine riemann_state(qm, qp, qpd_lo, qpd_hi, &
...
+             do i = lo(1), hi(1)
 
-             eos_state % rho = qp(i,j,kc,QRHO)
-             eos_state % p   = qp(i,j,kc,QPRES)
-             eos_state % e   = qp(i,j,kc,QREINT)/qp(i,j,kc,QRHO)
-             eos_state % xn  = qp(i,j,kc,QFS:QFS+nspec-1)
-             eos_state % aux = qp(i,j,kc,QFX:QFX+naux-1)
+                ! this is an initial guess for iterations, since we
+                ! can't be certain that temp is on interfaces
+                eos_state % T = 10000.0e0_rt
```
maxpkatz: It looks like this was missing before -- does this change cause tolerance-level diffs?

Source/hydro/ppm_nd.F90
zingale: yes.  Although this will be modified in a later PR anyway, since lo and hi will have a different meaning then.

Source/hydro/ppm_nd.F90
zingale: yes.  Good catch.

Source/hydro/riemann_nd.F90
zingale: I don't think we had any tests that exercised this, but yes, it would change things to EOS tolerance.

Source/hydro/Castro_ctu_1d.F90
```diff
@@ -354,7 +354,7 @@ subroutine umeth(q, q_lo, q_hi, &
 #endif
                 qaux, qa_lo, qa_hi, &
                 shk, shk_lo, shk_hi, &
-                1, lo(1), hi(1)+1, 0, 0, 0, 0, 0, &
+                1, lo, hi+1, &
```
maxpkatz: Needs to be hi+dg I think, not hi+1

zingale: yes :)

zingale: in my defense, I am working on getting rid of this function entirely in a different branch.

PR #414, opened 2018-09-27
remove allow_negative_energy

zingale:


Source/radiation/CastroRad_2d.f90
```diff
@@ -738,7 +737,7 @@ subroutine ca_estdt_rad(lo, hi, u,u_l1,u_l2,u_h1,u_h2, &
         eos_state % xn  = u(i,j,UFS:UFS+nspec-1) * rhoInv
         eos_state % aux = u(i,j,UFX:UFX+naux -1) * rhoInv
 
-        if (eos_state % e .gt. 0.e0_rt .or. allow_negative_energy.eq.1) then
+        if (eos_state % e .gt. 0.e0_rt) then
```
maxpkatz: I don't think this conditional should ever be possibly false now, so can we remove it?

zingale: works for me

PR #418, opened 2018-10-04
Gravity gpu

harpolea:
Offloads some of the gravity routines to the gpu. 

I've tested with all the gravity_tests problems, and get agreement with the development branch to a relative error of < 1e-14. 

Some routines involving arguments with `dataPtr`s have not been offloaded - while it was possible to offload those involving Vectors (using `CudaManagedAllocator`), I did not to offload those involving `dataPtr`s to `FArrayBox` objects as I was unable to work out how to pass these to the gpu.


zingale:
this fails to compile with gfortran:

```
     allocate(gravity_type)
             1
Error: Allocate-object at (1) with a deferred type parameter requires either a type-spec or SOURCE tag or a MOLD tag```

we should switch `gravity_type` from a string to an integer, and define `POISSON_GRAV=0`, `MONOPOLE_GRAV=1`, ...

**Comment by zingale** [comment]
is this ready for merging?  Should I rerun the suite with it?

**Comment by harpolea** [comment]
I believe it was running correctly when I stopped working on it a couple of months ago, but it probably needs some more checking to make sure it still works. The recent merges I’ve done I basically just patched it until it compiled again, so it could do with some more rigorous checking

**Comment by harpolea** [comment]
I’ll rerun some of the problems and compare against the dev branch solutions

**Comment by zingale** [comment]
For 2-d problems we get the error:
```
0::Assertion `dir >=0 && dir < AMREX_SPACEDIM' failed, file "/home/zingale/development/AMReX//Src/Amr/AMReX_AmrLevel.cpp", line 558 !!!
```
From the backtrace, this arises at:
```
    gravz.define(getEdgeBoxArray(2), dmap, 1, 0);                                                                                                                
```
we probably should not be doing that in 2-d.  I think we need some ```#if AMREX_SPACEDIM``` checks in there to mark off the higher dimensional stuff in 1- and 2-d

Issue #149, opened 2017-08-04
merge PLM and PPM tracing routines


zingale:
The basic structure of the characteristic tracing is the same for PLM and PPM, but we have separate tracing routines.  We should merge these together.

zingale:
also, should we trace all the source terms?  If you look in `trace_nd.F90`, it considers sources for all quantities, not just velocity (in 1-d)

zingale:
this is actually quite easy.  We can get rid of the PLM routines and instead define and I(nu) as

I(nu) = q_i + 1/2 (1 - dt/dx lambda(nu) ) dq_i

for the left state (and an analogous one for the right state), and then just use this in the PPM tracing routines instead of the integrals under the parabola.  

Then, we define the reference state as it is currently done with PLM, e.g., for the left state,

q_ref = q_i+ 1/2 (1 - dt/dx max(lambda(+),0)) dq_i

Otherwise, all the remaining operations will be the same as with the PPM, so this really simplifies the codes.

zingale:
These are different enough now in how we do some of the tracing that I am going to close this.

PR #427, opened 2018-10-20
ctu noslab opt

zingale:
This is blocked by #408 

This further optimizes and simplifies the noslab CTU by creating a single tracing routine regardless of dimension.  With this change, each of the tracing routines is specific to an eigensystem.  This largely eliminates the need to do #149, since we now want to keep these tracing routines as simple as possible.


zingale:
some errors are larger than expected (especially 2-d axisymmetric), so this needs to be looked at before merging.

zingale:
This seems to be fixed in c81e64e

zingale:
test results are here:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2018-10-24-001/index.html

PR #431, opened 2018-10-28
Put conductivity into the eos type

dwillcox:
Move conductivity into the eos type for interfacing with the conductivity routines.

Uses the same interface as for the conductivity-update branch in Microphysics.

Depends on https://github.com/starkiller-astro/Microphysics/pull/164


zingale:
this fails to compile:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2018-10-28-001/index.html

dwillcox:
That fixes the compile failures on my local branch.

PR #432, opened 2018-11-06
more CTU GPUing

zingale:
This set of changes merges the trans routines and moves the source terms in PLM into the trace routines instead of having them sprinkled into the trans routines.


zingale:
This also fixes a bug in the SDC stuff -- we were not previously taking the reactive source terms into account in the hydro for 1-d.

zingale:
Here are the tests:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2018-11-15-005/index.html

SDC change will be big due to bug fix.  2-d will change slightly (roundoff for pure hydro) due to the merge of the trans routines.

maxpkatz:
The convective_flame and evrard_collapse diffs are pretty big, any ideas what is going on there?

zingale:
My guess for evard_collapse is that it uses `ppm_type = 0`, and I changed how the source terms are dealt with there.  Now they are applied during tracing and already in the states when it comes to the transverse stuff.  It is second-order accurate either way, but this is cleaner now.

I'll look into convective_flame more closely.  But that problem is always sensitive.

maxpkatz:
For evrard_collapse: yes, that makes sense! Not sure why we are using PLM there but that's a question for another day.

For convective_flame: I'm OK to accept the "reaction problems are sensitive" explanation but I was curious why reacting_bubble which is also a 2D burning problem didn't show large diffs.

PR #442, opened 2018-11-16
more GPU changes

zingale:
This is more work on the GPU port for CTU.  Specificially, it:

* eliminates ppm_type = 2
* unifies the `ppm_reconstruct()` routines (the CUDA and original one)
* fixes a bug in how the CUDA MOL branch was assigning interface states (it mixed them up)


zingale:
all tests pass

Source/hydro/Castro_hydro.cpp
```diff
@@ -504,9 +504,10 @@ Castro::construct_mol_hydro_source(Real time, Real dt)
            BL_TO_FORTRAN_ANYD(flatn[mfi]));
 
       // Do PPM reconstruction to the zone edges.
+      int put_on_edges = 1;
 #pragma gpu
-      ca_ppm_reconstruct_cuda
-          (AMREX_INT_ANYD(obx.loVect()), AMREX_INT_ANYD(obx.hiVect()),
+      ca_ppm_reconstruct
```
maxpkatz: I think this call is now missing the NQ and n arguments after q[mfi]?

guadabsb15:
The following figure shows a comparison of density  for the Sod_stellar problem in 3D, using the inputs-test1-helm. Blue lines is output from code on gpu and red on cpu 

3d_more_recent

Source/hydro/Castro_mol_nd.F90
```diff
@@ -241,209 +219,101 @@ subroutine ca_mol_single_stage(lo, hi, time, &
...
-     endif
+     end do
+  end if
 
-     ! Compute F^x at kc (k3d)
-     call cmpflx(qxm, qxp, It_lo, It_hi, &
-                 flux1, flux1_lo, flux1_hi, &
-                 q1, flux1_lo, flux1_hi, &  ! temporary
+  ! Compute F^x at kc (k3d)
+  call cmpflx(qm(:,:,:,:,1), qp(:,:,:,:,1), It_lo, It_hi, &
```
maxpkatz: Is it possible to avoid this type of slicing? I recall that OpenACC tends not to like it, and we want to keep that route open.

Source/hydro/ppm_nd.F90
```diff
@@ -1267,19 +185,26 @@ subroutine ca_ppm_reconstruct_cuda(lo, hi, &
 
                 end if
 
-                qp(i  ,j,k,n,1) = sp
-                qm(i+1,j,k,n,1) = sm
+                if (put_on_edges == 1) then
```
maxpkatz: Can you explain, for my edification, what this option does?

Source/hydro/Castro_mol_nd.F90
zingale: I can either explicitly put a copy here, or I can have the cmpflx routine take the states of this form (with the interface direction as the last dimension) and then select the direction internally.  Preference?

Source/hydro/ppm_nd.F90
zingale: put_on_edges will take the left and right values of the parabola and use them to initialize the left and right interface states.  I'm open to a different name.

Source/hydro/Castro_mol_nd.F90
maxpkatz: I prefer the latter.

zingale: me too

zingale: this is fixed now.  `cmpflx` takes two additional arguments, the number of components in the state arrays and the which component we are using

zingale: tests still pass

Source/hydro/riemann_nd.F90
```diff
@@ -1822,32 +1818,32 @@ subroutine HLLC(ql, qr, qpd_lo, qpd_hi, &
 
              if (S_r <= ZERO) then
                 ! R region
-                call cons_state(qr(i,j,k,:), U_state)
+                call cons_state(qr(i,j,k,:,comp), U_state)
```
maxpkatz: Same comment about the slicing here -- since in general this can be non-contiguous, I think we're better off avoiding it.

zingale: oh you saw that ;)

maxpkatz: Since the slicing was already there, we could defer that change if you like.

PR #455, opened 2018-12-04
[WIP] still more GPU changes

zingale:
this eliminates all of the remaining bl_allocates downstream of the main hydro drivers.

PR #456, opened 2018-12-06
more more more GPU CTU changes

zingale:
This makes explicit the range of cells looped over in the transverse routines.
blocked by #455


zingale:
all tests passed.  This can merge after #455

zingale:
all tests pass
this is ready for review and merge

PR #461, opened 2018-12-14
Doxygen docstrings

harpolea:
This converts the existing docstrings into a format that can be interpreted by doxygen, allowing them to be included in the API generated by breathe/sphinx. 

I've also added a python script that is run after breathe has been executed which organizes the API's filelist into a more helpful structure.


zingale:
this seems to need python3, so for my machine (atleast), I needed to replace `python make_api.py` with `python3 make_api.py`

zingale:
I think using `python3` is compatible with PEP-394:
https://www.python.org/dev/peps/pep-0394/

harpolea:
Oh that will be the f strings - it will be probably be fine if I switch those to using `.format` instead

zingale:
I'd prefer it to require python3 anyway.  I just think we need to change the name in the makefile.

zingale:
This looks great to me.  Opens the door to lots of docstringing in the future.

Issue #465, opened 2018-12-19
put GPU hackathon notes into docs


zingale:
During the hackathon we made a list of best practices for porting solvers to the GPUs.  We should move these into the official User's Guide

PR #466, opened 2018-12-19
Gpu porting

harpolea:
These are a cleaned up version of the notes made during the hackathon on offloading things to GPU. I've done some formatting and added some examples to help illustrate things

This addresses #465


zingale:
fantastic

PR #468, opened 2018-12-20
Gravity gpu

harpolea:
This offloads 5 of the gravity routines to gpu. I've tested against all the `gravity_tests` and `rotating_torus` - it gives the same answers except for:
-  `evrard_collapse` (I cannot figure out what's wrong as the answers diverged when I ported `prescribe_phi` and `prescribe_grav`, two routines that essentially do nothing). 
- `DustCollapse` also  developed relative errors of <1e-14 when I offloaded `ca_gsrc` - my guess would be that is something to do with `sum` and `dot_product` having different implementations on the gpu?
- `rotating_torus` only agrees when `ca_gsrc` is on the gpu if hybrid momentum is switched off. It's not clear to me why this would be, as the only difference is it calling the routine `set_hybrid_momentum_source` which doesn't seem to do anything complicated

The gravity routines that have not been offloaded either act on regions outside of `lo:hi` or have `Vector.dataPtr()` objects passed to them (and therefore require dealing with `CudaManagedAllocator`'s to offload).


zingale:
should I test and review this?

maxpkatz:
Closing since this is now pretty far out of sync with the main code and much of it has been done elsewhere.

Issue #364, opened 2018-07-08
dxnuc tagging parameters should be in Fortran, not C++


maxpkatz:
These basically work the same way as the Fortran tagging parameters in the tagging namelist, so for uniformity, we should move them there.

I doubt anyone other than me is using these, so I'll probably just do it as a breaking change.

maxpkatz:
Note: these are currently named "dxnuc" and "dxnuc_max" since I hadn't anticipated implementing the latter when I originally created this. When we move to Fortran, we could consider renaming the former to "dxnuc_min" for symmetry.

PR #473, opened 2019-01-03
add enuc tagging

### LINKED ISSUE #437: add refinement on enuc
it would be nice to be able to refine on nuclear energy generation rate, enuc

zingale:
closes #437


zingale:
This also closes #364 
I moved the `dxnuc` stuff to Fortran in the tagging module.

Note: even though a bunch of inputs had `dxnuc` set, none of them had `max_dxnuc_lev` set, so it looks like it never took effect?

maxpkatz:
> Note: even though a bunch of inputs had dxnuc set, none of them had max_dxnuc_lev set, so it looks like it never took effect?

I have used this in some of my testing on ignition problems, but it has not yet made it into any of the production inputs files since I haven't yet found a use case for it.

PR #476, opened 2019-01-13
Added option for linear dropoff of diffusion term

KiranEiden:
Added the diffuse_cutoff_density_hi parameter. The diffusion term will
now drop off linearly between diffuse_cutoff_density_hi and
diffuse_cutoff_density when the former is set to be higher than the latter.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,16 +1,22 @@
 {
     "creators": [
         {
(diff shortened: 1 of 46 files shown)


zingale:
Note this does not modify the timestep constraint, but I think we are still safe

PR #480, opened 2019-01-17
CTU reorg

zingale:
This is the start of splitting the CTU functions into pieces that are grouped together by the same lo:hi, so we can offload to the GPU


Source/hydro/Castro_ctu_nd.F90
```diff
@@ -104,245 +92,95 @@ subroutine umeth(q, qd_lo, qd_hi, &
...
+    compute_shock = .true.
 #else
+    compute_shock = .false.
+#endif
+
     ! multidimensional shock detection -- this will be used to do the
     ! hybrid Riemann solver
-    if (hybrid_riemann == 1) then
-       call ca_shock(lo-dg, hi+dg, &
+    if (hybrid_riemann == 1 .or. compute_shock) then
```
maxpkatz: Previously, we would zero out the `shk` data if `SHOCK_VAR` is defined but `hybrid_riemann /= 1`. I think this changes the logic, because now `shk` will have non-zero data if `SHOCK_VAR` is defined regardless of the value of `hybrid_riemann`. Not sure if that has any impact down the line.

Source/hydro/Castro_ctu_nd.F90
```diff
@@ -356,11 +194,14 @@ subroutine umeth(q, qd_lo, qd_hi, &
...
-    ! preprocess the sources -- we don't want to trace under a source that is empty
+    ! preprocess the sources -- we don't want to trace under a source
+    ! that is empty.  Note, we need to do this check over the entire
+    ! grid, to be sure, e.g., use vlo:vhi.  We cannot rely on lo:hi,
+    ! since that may just be a single zone on the GPU.
     if (ppm_type > 0) then
        do n = 1, QVAR
-          if (minval(srcQ(lo(1)-2:hi(1)+2,lo(2)-2*dg(2):hi(2)+2*dg(2),lo(3)-2*dg(3):hi(3)+2*dg(3),n)) == ZERO .and. &
-               maxval(srcQ(lo(1)-2:hi(1)+2,lo(2)-2*dg(2):hi(2)+2*dg(2),lo(3)-2*dg(3):hi(3)+2*dg(3),n)) == ZERO) then
+          if (minval(srcQ(vlo(1)-2:vhi(1)+2,vlo(2)-2*dg(2):vhi(2)+2*dg(2),vlo(3)-2*dg(3):vhi(3)+2*dg(3),n)) == ZERO .and. &
```
maxpkatz: Occurs to me that we have to refactor this check to do it on the GPU.

maxpkatz:
Test suite run?

maxpkatz:
From Slack:
yeah, it passed
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2019-01-17-007/index.html
I fixed the oddeven stuff after that run

PR #483, opened 2019-01-19
more CTU GPU reorg

zingale:
This gets all of the allocates out of the transverse and flux subroutine and makes all of the interfaces specify lo and hi for all quantities.


zingale:
this has a bug in it... I reimplemented this PR in ctu-reorg-2-redo

PR #484, opened 2019-01-20
more CTU reorg

zingale:
This makes explicit the bounds of all of the arrays passed into the cmpflx and trans routines and also
moves all of the allocations into the main driver.  This prepares us for the GPU offload now.


zingale:
this seems to have slowed down quite a bit over `development`.  Need to understand this.

maxpkatz:
The slowdown was not real. Merging.

PR #487, opened 2019-01-22
create bindings for trans* and cmpflx_plus_godunov and reorganize calls in main driver

zingale:
This should be the last step before offloading these to GPUs

### DIFF
--- a/Source/driver/Castro_F.H
+++ b/Source/driver/Castro_F.H
@@ -632,6 +632,128 @@ extern "C"
      amrex::Real& zang_lost);
 
 
+  void transx(const int lo[], const int hi[],
+              const BL_FORT_FAB_ARG_3D(qym),
+              BL_FORT_FAB_ARG_3D(qymo),
+              const BL_FORT_FAB_ARG_3D(qyp),
+              BL_FORT_FAB_ARG_3D(qypo),
+#if AMREX_SPACEDIM == 3
+              const BL_FORT_FAB_ARG_3D(qzm),
+              BL_FORT_FAB_ARG_3D(qzmo),
+              const BL_FORT_FAB_ARG_3D(qzp),
+              BL_FORT_FAB_ARG_3D(qzpo),
+#endif
+              const BL_FORT_FAB_ARG_3D(qaux),
+              const BL_FORT_FAB_ARG_3D(fx),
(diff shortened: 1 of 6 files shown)


zingale:
this is not yet passing tests

zingale:
all tests now pass

Source/driver/Castro_F.H
```diff
@@ -631,6 +631,161 @@ extern "C"
      amrex::Real& yang_lost,
      amrex::Real& zang_lost);
 
+  void ctu_normal_states
+     (const int lo[], const int hi[],
```
maxpkatz: I believe we should stick to const int*, not const int [].

Source/driver/Castro_F.H
```diff
@@ -631,6 +631,161 @@ extern "C"
...
+              const BL_FORT_FAB_ARG_3D(fx),
+#ifdef RADIATION
+              const BL_FORT_FAB_ARG_3D(rfx),
+#endif
+              const BL_FORT_FAB_ARG_3D(qx),
+#if AMREX_SPACEDIM == 2
+              const BL_FORT_FAB_ARG_3D(area1),
+              const BL_FORT_FAB_ARG_3D(vol),
+#endif
+              const amrex::Real hdt, const amrex::Real cdtdx,
```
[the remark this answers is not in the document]
zingale: turns out it is used because of the area factors.  Some terms are grad instead of div, so we need to do things slightly differently.

Source/hydro/Castro_ctu_nd.F90
```diff
@@ -466,1340 +467,356 @@ subroutine ctu_normal_states(lo, hi, &
...
 #if AMREX_SPACEDIM == 3
-    integer, intent(in) :: rf3_lo(3), rf3_hi(3)
+    integer, intent(in) :: radflux3_lo(3), radflux3_hi(3)
 #endif
+    integer, intent(inout) :: nstep_fsp
 #endif
-    integer, intent(in) :: domlo(3), domhi(3)
 
-    real(rt), intent(in) :: dx(3), dt
+    integer, intent(in) :: verbose
```
maxpkatz: We will want this to be passed by value eventually

PR #491, opened 2019-01-23
clear some more bits out of consup

zingale:
This removes some more bits out of consup to allow for GPU offloading


zingale:
all tests pass

PR #492, opened 2019-01-25
Analytical solution vs Plotfile for Rad2TShock python script

ChrisDeGrendele:


zingale:
can you replace the 
```
from numpy import *
from pylab import *
```
with the more standard
```
import numpy as np
import matplotlib.pyplot as plt
```
and add the appropriate namespaces where needed?

(pylab is no longer recommended: https://matplotlib.org/faq/usage_faq.html#matplotlib-pyplot-and-pylab-how-are-they-related)

PR #498, opened 2019-02-09
C++ version of the CTU driver

zingale:
All the CTU kernels are now called from C++.

This needs an optimization pass


maxpkatz:
Approved for correctness. Tests are passing as well.

Next step is to decrease the memory footprint by moving allocations back inside MFIter loops as local Fabs.

PR #501, opened 2019-02-09
have cmpflx call riemann_state to reduce duplicate code

### LINKED ISSUE #479: merge riemann_state and cmpflx
It seems that `cmpflx` can simply call `riemann_state` for the first part of the Riemann solve, reducing duplicate code.

zingale:
closes #479

### DIFF
--- a/Source/hydro/riemann_nd.F90
+++ b/Source/hydro/riemann_nd.F90
@@ -185,95 +185,19 @@ contains
 
     !$gpu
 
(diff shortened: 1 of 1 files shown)


zingale:
note: I've confirmed that the CUDA path still compiles

zingale:
all tests passed: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2019-02-09-001/index.html

Issue #321, opened 2018-04-23
create T_guess parameter


zingale:
When inverting the EOS (e.g., calling with `eos_input_re`), we need an initial guess for T.  Often we just hardcode a value in the code, but we should have this be a runtime parameter, since different initial guesses make sense for different EOSes.

zingale:
this has been merged

PR #509, opened 2019-02-13
Added T_guess parameter

harpolea:
Addresses #321 

This adds `T_guess` as a runtime parameter so it can be specified at runtime rather than hardcoded (e.g. see changes to `riemann_nd.F90`).

I've tested it by running `reacting_bubble` and nothing seems broken.


zingale:
I think there is a another guess in `riemann_nd.F90`, around like 1551:
```
eos_state % T = 1.e4  ! a guess
```

harpolea:
^ fixed in 9b9b10c

PR #514, opened 2019-02-18
CTU plm ppm split

zingale:
This splits `ctu_normal_states` into separate routines for ppm and plm


zingale:
This also makes the arrays that hold the integrals under the parabolas hold only one dimension at a time.  This significantly cuts down on the memory usage.  The results all agree in the test suite and the CPU code runs faster now.

Issue #508, opened 2019-02-13
convert the fsedov and radiation routines in Diagnostics/ to C++


zingale:
F_Baselib is going away, so we want these in the C++ framework

PR #515, opened 2019-02-18
Cpp diagnostics

harpolea:
This addresses #508 

Converts the Fortran versions of the Diagnostics tests fsedov and radiation to C++


zingale:
so there is now a single routine replacing all of the previous ones?  So for a spherical problem we add `--sphr`? 

Can we drop a README.md in there?

harpolea:
Yes, because the files are now read in using AMReX_Dataservices.H, they require compiling with the correct `DIM` flag, so I combined all the Sedov routines into one file. A README is a good idea - I did a brief one for the Radiation routines as the compilation is a bit confusing (let me know if there's a way to do it that doesn't require editing two files), but shall make sure there are README's for the other two

zingale:
if I run on the 3-d sedov problem (with `inputs.3d.sph`), the code core dumps and the output is seems to have a lot of gaps.  I ran the tool as:
```
sedov_3d.ex --sphr -p sedov_3d_plt00119 -s sedov.out
```

zingale:
oh... I needed to set `--xctr`, `--yctr`, and `--zctr`.  Now the plot looks good.  But it still segfaults at the end -- that might be an AMReX issue?

harpolea:
I suspect that may indeed be an AMReX issue (possibly to do with the Dataservices class?). I'll try running it through gdb/valgrind and see if I can spot something

Diagnostics/Sedov/sedov_util.f90
```diff
@@ -0,0 +1,369 @@
...
+
+  integer, intent(in) :: lo(3), hi(3)
+  integer, intent(in) :: plo(3), phi(3), nc_p
+  integer, intent(in), value :: nbins
+  real(rt), intent(in) :: p(plo(1):phi(1),plo(2):phi(2),plo(3):phi(3),0:nc_p-1)
+  real(rt), intent(inout) :: dens_bin(0:nbins-1)
+  real(rt), intent(inout) :: vel_bin(0:nbins-1)
+  real(rt), intent(inout) :: pres_bin(0:nbins-1)
+  real(rt), intent(inout) :: e_bin(0:nbins-1)
+  integer, intent(inout) :: volcount(0:nbins-1)
```
zingale: this needs to be real, not integer, since the volume that you are adding to it, `vol` is real.

zingale:
there may be more instances of volcount needing to be real -- I didn't check them all.  But at the moment, we don't get the right plot for this test problem:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2019-03-12/Sedov-2d.html

also, the center is at the origin for this problem, so I added xctr to this routine

zingale:
also... it might be nice to grab the center from the plotfile's job_info directly.

Diagnostics/Sedov/sedov_util.f90
harpolea: fixed in 9a1b0b6

zingale:
okay, these work for the test suite sedov problems, run as:
```
./sedov_2d.ex -p Sedov-2d_plt00130 --sphr --xctr 0 --yctr 0 -s Sedov-2d_plt00130_new.txt
./sedov_2d.ex --xctr 0.5 --yctr 0.5 -p Sedov-2d-tau_plt00158 -s Sedov-2d-tau_plt00158_new.out
./sedov_3d.ex --sphr --xctr 0.5 --yctr 0.5 --zctr 0.5 -p Sedov-3d_plt00127 -s Sedov-3d_plt00127_new.txt
```

PR #517, opened 2019-02-19
rename QVAR -> NQSRC

zingale:
also fix a number of places where NQ should have been used instead of QVAR


zingale:
this is ready for review.  All tests are passing.

PR #519, opened 2019-02-20
get rid of species sources by default

### LINKED ISSUE #350: remove source terms for species

zingale:
This change eliminates species sources by requiring us to set the parameter `PRIM_SPECIES_HAVE_SOURCES`.  To support this the `set_variables.py` script was updated to now allow the "adds-to" field to have an optional if-def requirement.  

This change also fixes 2 bugs.  In `srctoprim`, we were converting the velocity sources twice.  And in the radiation tracing, we were not adding species sources.

This closes #350 

Note: this should be merged after #517


zingale:
This is the test suite output:  http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2019-02-19-005/index.html

Source/hydro/advection_util_nd.F90
```diff
@@ -692,6 +692,13 @@ subroutine ca_srctoprim(lo, hi, &
        n = upass_map(ipassive)
        iq = qpass_map(ipassive)
 
+       ! we already accounted for velocities above
+       if (n == QU .or. n == QV .or. n == QW) cycle
```
maxpkatz: I think you should be testing on iq here, not n?

zingale: yes.  Thankfully the velocity indices are the same.  Wonder if this is wrong elsewhere...

zingale: I checked the other places and it looked okay elsewhere.

Issue #403, opened 2018-09-19
create a model_parser_init


zingale:
some problems (like `flame_wave`) use the `model_parser_module` as a container to hold the data, but don't read in a model.  So we should have a way to allocate the module data outside of the reader (e.g., a `model_parser_init()`

zingale:
this has been implemented.

PR #520, opened 2019-02-21
model_parser_init function

harpolea:
This addresses #403, and creates a model_parser_init function which allocates and initializes the module information for problems that don't read this information in from a file.


zingale:
this looks good.  I'll switch flame_wave over to this once it is merged.

PR #523, opened 2019-02-24
Add rebalanced splitting option for CTU

maxpkatz:
This PR implements rebalanced splitting, as described in [Speth et al. (2013)](https://dspace.mit.edu/handle/1721.1/86001).

Balanced splitting is when we select a particular linear term to add to the reaction and advection operators that ensures that both see the same equilibrium state. Thus this method is better able to maintain equilibrium conditions when we are near them, and in an exact equilibrium state, balanced splitting exactly maintains the equilibrium (unlike normal Strang splitting). This approach can be lower error in general compared to normal Strang splitting.

Balanced splitting can be implemented using an instantaneous evaluation of the reaction and advection operators at the beginning of the timestep, but this is known to be unstable in general unless the timestep is strongly limited. My experiments with the Detonation problem bear this out. For example, if the instantaneous nuclear energy generation rate at the beginning of the timestep is much larger than the average over the full dt/2 burn, the balancing operation kicks the state way too far out of equilibrium and can lead to an instant ignition.

Rebalanced splitting modifies balanced splitting by using the integrated value of the reaction and advection operations from the last timestep to select the balancing state vector. This is not proven to be unconditionally stable in the paper, but in practice they find it to work for advection-reaction problems. My testing with the Detonation problem shows that it does not seem to cause stability problems there.


maxpkatz:
I believe this can also be applied to the non-reacting source terms. Right now we apply these source terms in a predictor-corrector fashion, but it is not clear that the added complexity of the predictor-corrector actually helps us compared to doing Strang splitting for these sources. Strang splitting would be a lot simpler to implement and wouldn't require us to explicitly deal with the source terms in the hydro. And then it would be much more straightforward to keep the source term approaches in sync between CTU and MOL/SDC. So, I plan to investigate in a separate PR whether it makes sense to switch to Strang splitting. If it does, then we could use rebalanced splitting on the sum of all the Strang-split sources. One benefit I see from this is that it would be a way for us to indirectly capture most of the benefit from methodologies like well-balanced gravity, without requiring the complexity of actually implementing those specific solutions.

Source/driver/Castro.H
```diff
@@ -41,9 +41,12 @@ enum StateType { State_Type = 0,
 #endif
 #ifdef SDC
 #ifdef REACTIONS
-		 SDC_React_Type
+		 SDC_React_Type,
 #endif
 #endif
+#ifdef REACTIONS
+                 Balance_Type,
```
zingale: it's okay to have an extra `,` at the end?

maxpkatz: It seems to compile with gcc. I haven't tried with other compilers. I am hoping that it is generally OK, so that we don't have to worry about what ends up being the last item depending on what gets ifdef'd out.

maxpkatz: Though actually I'm about to make it defined always, so it won't matter in this case.

zingale: okay, but if it is defined always, can we ensure that it doesn't allocate the memory unless needed?

maxpkatz: Indeed. We are only allocating the memory if the rebalanced_splitting option is enabled.

zingale:
> I believe this can also be applied to the non-reacting source terms. Right now we apply these source terms in a predictor-corrector fashion, but it is not clear that the added complexity of the predictor-corrector actually helps us compared to doing Strang splitting for these sources. Strang splitting would be a lot simpler to implement and wouldn't require us to explicitly deal with the source terms in the hydro. And then it would be much more straightforward to keep the source term approaches in sync between CTU and MOL/SDC. So, I plan to investigate in a separate PR whether it makes sense to switch to Strang splitting. If it does, then we could use rebalanced splitting on the sum of all the Strang-split sources. One benefit I see from this is that it would be a way for us to indirectly capture most of the benefit from methodologies like well-balanced gravity, without requiring the complexity of actually implementing those specific solutions.

I am not sure that's the same as well-balancing.  The goal of well-balancing is for the interface states to be predicted in a fashion that HSE is preserved.

maxpkatz:
Indeed it's not the same as well-balancing. However I think it still better preserves HSE. Consider a 1D spherical star simulation that is initialized exactly in HSE. In the first timestep with predictor-corrector we'll get an update in the zone that looks like:
```
(rho v)^{n+1} = (rho v)^n - dt / 2 * ((rho g)^n + (rho g)^{n+1})) + dt * dp/dx^{n+1/2}.
```
By construction we have that dp / dx^n = (rho g)^n, but the time-centered dp/dx estimate is not restricted in this way (without applying well-balancing), so in general we can violate HSE.

Now consider the case where we Strang split these operations instead. We will do
```
(rho v)^* = (rho v)^n - dt / 2 * ((rho g)^n)
(rho v)^** = (rho v)^* + dt * dp/dx^* = (rho v)^n - dt / 2 * (rho g)^n + dt * dp/dx^*
(rho v)^n+1 = (rho v)^** - dt / 2 * (rho g)^** = (rho v)^n - dt / 2 * ((rho g)^n + (rho g)^**) + dt * dp/dx^*
```
Again, this will not preserve HSE since the first Strang split gravity update knocks the state out of equilibrium and we aren't constructing the pressure update to exactly cancel it.

But with balanced splitting we evaluate at time-level n the balance vector,
```
c^n = (1/2) * ((dp/dx)^n + (rho g)^n)
```
So the Strang operations become
```
(rho v)^* = (rho v)^n - dt / 2 * (rho g)^n + dt / 2 * c^n = (rho v)^n
(rho v)^** = (rho v)^* + dt * (dp/dx)^* - dt * c^n = (rho v)^n + dt * (dp/dx)^n - dt * c^n = (rho v)^n
(rho v)^{n+1} = (rho v)^** - dt / 2 * (rho g)^** + dt / 2 * c^n = (rho v)^n - dt / 2 (rho g)^n + dt / 2 * c^n = (rho v)^n
```
That is, at each Strang step, subtracting off the balance vector makes the advective and gravity terms identically a no-op if we're already in HSE.

One subtlety worth keeping in mind, if we do this, is that we need to carefully initialize c^n in the first timestep. Right now I'm not doing this, so that the balancing does nothing in the first timestep, and then we fill in the balancing vector for the second timestep. This should be OK for reacting simulations since we don't usually start our simulations in a reaction equilibrium, but we need to revisit it if we're going to use it for HSE.

PR #530, opened 2019-03-02
[WIP] Add self-consistent field initialization

maxpkatz:


zingale:
@harpolea can I ask you to review this? since this is similar to some of the stuff you are doing for rotating stars

harpolea:
@mkatz is this ready to review now?

maxpkatz:
It is in good shape now. But it still has the limitation that you have to specify the equatorial and polar radius, in addition to the maximum density of the star. This means that the only way to find a good combination is through trial and error. As far as I know, no one has come up with a better way here. But I'm thinking about whether I can find one, that allows you to specify a more natural variable (such as rotation period).

harpolea:
You could try something like the method in [Jackson et al. 2005](http://dx.doi.org/10.1086/426587), where instead of the central density, equatorial and polar radii they specify the total mass, rotational parameters and chemical composition?

harpolea:
Do you have an inputs file for a rotating case? There's only a non-rotating star currently

maxpkatz:
I was looking at Jackson et al. recently, I'll think about whether it's feasible here.

I'm going to add a rotating case soon. I just have to do some trial and error to find one.

harpolea:
Ok cool. Everything else looks good to me.

maxpkatz:
Great. I think I'll want to merge it soon, and then try to do a better approach later. The main thing I have to fix (other than adding the rotating case) is making it work with AMR. Should be easy.

maxpkatz:
I've added a rotating test case. I've also added AMR-awareness to the algorithm. One challenge with AMR is that it seems to often result in cases where the low-density material oscillates indefinitely due to discretization error. (I can't tell if this is the result of AMR, or AMR is just coincidentally exposing a limitation of the method.) So I also added a check that prevents low-density material from stalling the convergence testing.

After looking at the Jackson et al. paper more closely, I don't want to attempt to directly reproduce their work. However, to the extent that their idea boils down to an SCF inner loop (to obtain a particular density distribution) and a Newton-Raphson outer loop (to converge on the desired mass and rotation rate), that is the same idea I had.

PR #532, opened 2019-03-03
remove the code to extrapolate diffusion terms to ghost cells

zingale:
we do proper ghost cell fills now, so this is not needed.


maxpkatz:
Approved. We could also resize the TempDiffTerm, SpecDiffTerm, etc. to have zero ghost zones, I think.

PR #535, opened 2019-03-03
Only apply sources on valid domain, then fill S_new ghost zones

maxpkatz:
The non-reacting source terms (including the hydro) are only calculated on the valid domain. The source terms are then FillPatched so that there is meaningful data in the ghost zones. So, it is fine to do apply_source_to_state and then do it on all of the State_Type ghost zones. However, the hydro_source has no ghost zones (and there is no meaning of a FillPatch, as this is a MultiFab which is locally defined on each level. So, we have to do a call to expand_state after we apply the hydro source. If this is going to be required anyway, then we should skip adding the (non-hydro) source terms to the State_Type ghost zones, and then defer all of the ghost-zone updates until after the corrector source terms have been applied.

Note that we still need to do the FillPatch in do_old_sources and do_new_sources, so that there's valid data in those ghost zones for interpolation on fine levels, and so that the FillPatchAdd to sources_for_hydro in CTU has data on the ghost zones.


maxpkatz:
This also fixes a bug in apply_source_to_state: we only need to call clean_state on the ghost zones that we actually did anything to.

This PR should change results, so let's run the test suite on it to understand what the size of the diffs looks like.

PR #537, opened 2019-03-03
Remove clean_state from apply_source_to_state

maxpkatz:
apply_source_to_state() is currently called in three places in CTU: after the old sources have been calculated, after the hydro source has been calculated, and then after the new sources have been calculated. For the first two, a call to clean_state is strictly unnecessary because there's already a call to clean_state after the hydro is applied, and neither the old source calculation nor the hydro calculation depends on the value of S_new. So this should not change results, except to roundoff error (or the iteration tolerance in the EOS). For the third, we can simply move the clean_state call explicitly into the CTU advance. (We do the same for the calls to apply_source_to_state in the reflux.)


maxpkatz:
Gated by #535

maxpkatz:
Ready for review; please run the test suite, as I expect this will change answers for problems with reactions.

zingale:
here's the results:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2019-03-04-001/index.html

note: not just reacting CTU problems have different answers, but gravity problems too, like StarGrav.  And in particular, the error in StarGrav is quite large.

zingale:
so I think it changes problems with sponges.  Since the sponge alters the thermodynamics, shouldn't we do a clean_state after applying the sponge?

maxpkatz:
I don't understand fully yet what explains the diff in StarGrav, and that makes me nervous. I'm reverting that specific change until we do understand it. So this PR now only changes interfaces without changing results.

Issue #324, opened 2018-04-26
state_nghost gives different answers


zingale:
Run `reacting_bubble` with `inputs_2d_test` as:

```
 ./Castro2d.gnu.MPI.ex inputs_2d_test castro.state_nghost=0 max_step=10
```
and then

```
./Castro2d.gnu.MPI.ex inputs_2d_test castro.state_nghost=1 max_step=10
```

and there are nontrivial differences:

```
            variable name            absolute error            relative error
                                        (||A - B||)         (||A - B||/||A||)
 ----------------------------------------------------------------------
 level =  1
 density                           0.8425712585E-03          0.3714247577E-12
 xmom                               320278.7954              0.2100573021E-05
 ymom                               267636.6177              0.5798493429E-08
 zmom                               0.000000000               0.000000000    
 rho_E                             0.1145691116E+16          0.2984930291E-12
 rho_e                             0.1145691116E+16          0.2984930291E-12
 Temp                              0.5298435688E-01          0.7741032661E-10
 rho_C12                           0.2527236938E-03          0.3713546910E-12
 rho_O16                           0.5886554718E-03          0.3707040693E-12
 rho_Mg24                          0.4141510243E-05          0.8551035646E-06
 phiGrav                            0.000000000               0.000000000    
 grav_x                             0.000000000               0.000000000    
 grav_y                             0.000000000               0.000000000    
 grav_z                             0.000000000               0.000000000    
 omegadot_C12                      0.1932632789E-11          0.4326742623E-06
 omegadot_O16                      0.3865265579E-11           2.000000000    
 omegadot_Mg24                     0.9663164850E-11          0.2163371005E-05
 enuc                               4456347.807              0.1781261133E-05
 rho_enuc                          0.6370308828E+16          0.1823618222E-05
 pressure                          0.6506360057E+15          0.4507366116E-12
 kineng                             6121024640.              0.7837523051E-08
 soundspeed                        0.5171298981E-03          0.5599428806E-12
 Gamma_1                           0.1355582313E-11          0.1003833509E-11
 MachNumber                        0.2027028380E-12          0.5056993624E-08
 entropy                           0.1647248864E-02          0.2676302374E-10
 magvort                           0.9423386951E-10          0.1382169941E-07
 divu                              0.4817392440E-09          0.7182512651E-07
 eint_E                             629504.0000              0.3720496785E-12
 eint_e                             629504.0000              0.3720496785E-12
[...]

Issue #353, opened 2018-07-03
extra memory usage in clean_state


zingale:
`clean_state` is a bit wasteful with memory.  In particular, there are 2 interfaces:

```
Castro::clean_state(int is_new, int ng)
```
and
```
Castro::clean_state(int is_new, MultiFab& state_old, int ng) 
```
The latter uses the MF passed in to compute the `frac_change` from the `enforce_min_density()` call.  The former interfaces copies the desired `StateData` to a temporary `MF` and then calls the other `clean_state` interface with this MF.  

It's not clear we always want to do this copy, since sometimes we are calling `clean_state` without regard to the return value (like in `expand_state`).  In these cases, we should create an option that allows us to skip this copy (perhaps by passing two copies of the same thing into `enforce_min_density`.

zingale:
notes from slack:

* the point of the copy was to avoid aliasing in Fortran
* we can perhaps create a separate interface that works on only one state

maxpkatz:
Noting here that the reason enforce_min_density has a return value is that we use to determine if we need a retry, so we're recording whether any zones had a negative density after the hydro update, and how negative they were. The latter is now not really needed because the dynamic retry capability allows us to just try a timestep half as long and hope it works, and the former can be achieved by just checking if state.min(Density) < 0 after the hydro update and storing it in a bool.

maxpkatz:
This will be addressed in #538

maxpkatz:
Resolved in #538

Issue #536, opened 2019-03-03
S_old should not be changed in expand_state


maxpkatz:
For CTU (and others), S_old is directly updated in the call to expand_state, when we do clean_state. I don't like this: S_old should be strictly static during an update. It is done with good intentions, since FillPatch is sourced from StateData, and as written we have to call clean_state on S_old so that we can get "clean" data in Sborder. I think the way to fix this -- and simultaneously my observation in issue #324 -- is to do the FillPatch first and only then call clean_state, on Sborder directly.

zingale:
Note that the proposed fix here is opposite the approach we took with #721 , where we relied on S_old being clean so we could eliminate the `clean_state` in `initialize_do_advance`

maxpkatz:
I'm not sure what I was thinking when I filed this bug, because indeed we already do call clean_state on S_old in the beginning of the step because some of the data may have been fillpatched into creation just prior to the advance. And it's important that this data be valid.

PR #538, opened 2019-03-04
Require clean_state to directly work on a state MultiFab

maxpkatz:
## PR summary

This change removes the indirect clean_state interface that took is_new as an argument, and now clean_state must be directly handed the MultiFab that it will operate on. This was done to partially address the concern in #536 that expand_state is directly operating on S_old when it should not be doing so. It also addresses the concern from #324 that the behavior in expand_state was resulting in differences depending on whether the state has ghost zones.

This change also removes the interface to clean_state that passes in an old state for comparison. This was not necessary but it is nicer to only have a single interface, and the logic there was a little complicated. This choice necessitated a change in how we do enforce_minimum_density, with two downstream consequences. First, density_reset_method == 3 no longer exists -- since we don't have the old state to compare to, we cannot reset to it. (No test problems are actually using that method.) Second, because we don't have the old state, we have to change the logic for the retry method based on negative densities. Now we look at the change from the negative density to what it is reset to, rather than the density it originally came from. This will change the outcome of that method a bit, but qualitatively it does the same thing, since the hope is that the density we reset to is at least in the same ballpark as the density from the beginning of the step (we reset to a density similar to the density in adjacent zones).

Finally, clean_state has been removed from expand_state for clarity. In places where we requested a clean_state call indirectly through expand_state, these have been replaced by an explicit clean_state call after the expand_state call.

## PR checklist


maxpkatz:
Gated by #537

maxpkatz:
Note: I still need to deal with computeTemp, this is not ready yet.

maxpkatz:
This should also resolve #353

maxpkatz:
Ready for review. The desired test suite output is a hard zero for second order and a small diff for fourth order.

maxpkatz:
Radiation will also have a small diff since the state has a ghost zone.

maxpkatz:
Update: this PR no longer changes behavior, it only cleans up the interfaces.

maxpkatz:
So, since this PR doesn't change results, this PR does not close #324 or #536. Those will be addressed in a future PR.

#353 will be closed by this PR: we no longer make a copy of the state for usage in enforce_min_density.

Source/driver/Castro_advance_ctu.cpp
```diff
@@ -129,14 +128,12 @@ Castro::do_advance_ctu(Real time,
...
-      int is_new = 1;
-      clean_state(is_new, 0);
+      clean_state(S_new, cur_time, 0);
     }
 
 
     // Sync up state after old sources and hydro source.
-    int is_new=1;
-    frac_change = clean_state(is_new, Sborder, 0);
+    frac_change = clean_state(S_new, cur_time, 0);
```
zingale: we were doing Sborder here before -- isn't that a different time?

maxpkatz: In this call, Sborder was the "old" state passed to enforce_min_density so that we could judge how bad the negative density was, if we had one. The actual cleaning was still on S_new.

PR #540, opened 2019-03-07
Radiation groups check

### LINKED ISSUE #528: Radiation should throw a runtime error if using single-group solver but NGROUPS > 1
RadSuOlson accidentally was being given NGROUPS = 2 at compile time in the test suite, resulting in corrupted plotfiles.

harpolea:
This fixes #528


Source/radiation/Radiation.cpp
```diff
@@ -138,6 +138,11 @@ void Radiation::read_static_params()
     do_multigroup = 0;
   }
 
+  // check that we're not using a single group solver if NGROUPS > 1
+  if (!do_multigroup && NGROUPS > 1) {
```
maxpkatz: If we've filled in Radiation::nGroups by now (I think we have?) could you check on that instead?

PR #541, opened 2019-03-09
Refactor diffusion for GPU compatibility

maxpkatz:
This PR refactors the Fortran code for diffusion to be GPU compatible. To do so, the call to calculate the cell-centered conductivity term was split into a separate call from the routine that averages the cell-centered value to edges.

The MGutils files were made dimension agnostic and also set up for GPUing.

The diffusion_test problem was made compatible with running in 1D and 3D. This allows us to do performance testing with the latter.


Source/driver/Derive_nd.F90
```diff
@@ -1400,23 +1400,58 @@ subroutine derdiffterm(diff,u_lo,u_hi,nd, &
...
+    c_hi = hi + dg
+
+    allocate(coeff_c(c_lo(1):c_hi(1), c_lo(2):c_hi(2), c_lo(3):c_hi(3)))
 
     call ca_fill_temp_cond(lo, hi, &
                            state, d_lo, d_hi, &
-                           coeff_x, d_lo, d_hi, &
-                           coeff_y, d_lo, d_hi, &
-                           coeff_z, d_lo, d_hi)
+                           coeff_c, c_lo, c_hi)
```
zingale: this call should be from `c_lo` to `c_hi`

PR #545, opened 2019-03-11
add thermo_src to source_flag

### LINKED ISSUE #543: thermo source missing from source_flag()
It looks like the thermo source, which is the pdivU term applied for MOL integration, is not considered in `Castro::source_flag()`.  This seems to be an oversight.

zingale:
closes #543


zingale:
There was also a sign error, the source should be -p div(U)

Finally, we cannot use `castro.update_sources_after_reflux` (see #549 ).

With these changes, the 2-d sedov looks great:

sedov_2d_sph_in_cyl

zingale:
We also now don't do update_sources_after_reflux for integration methods that don't support it.

Source/sources/Castro_sources.cpp
```diff
@@ -46,6 +46,13 @@ Castro::source_flag(int src)
 	else
 	    return false;
 
+    case thermo_src:
+        if (time_integration_method == CornerTransportUpwind ||
```
maxpkatz: Minor thing, but you could reverse this logic to return true for MOL and SDC and false otherwise? It's better future proofed if we add more integration strategies later.

Source/sources/thermo_nd.F90
```diff
@@ -41,19 +61,21 @@ subroutine ca_thermo_src(lo, hi, &
     do k = lo(3), hi(3)
        do j = lo(2), hi(2)
           do i = lo(1), hi(1)
-             src(i,j,k,UEINT) = FOURTH*((old_state(i+1,j,k,UMX)/old_state(i+1,j,k,URHO) + &
-                                         new_state(i+1,j,k,UMX)/new_state(i+1,j,k,URHO)) - &
-                                        (old_state(i-1,j,k,UMX)/old_state(i-1,j,k,URHO) + &
-                                         new_state(i-1,j,k,UMX)/new_state(i-1,j,k,URHO)))/dx(1)
+
+             ! compute -div{U}
+             src(i,j,k,UEINT) = -FOURTH*((old_state(i+1,j,k,UMX)/old_state(i+1,j,k,URHO) + &
```
maxpkatz: Unrelated to this PR, but is this calculation correct in non-Cartesian coordinates?

Source/sources/Castro_thermo.cpp
```diff
@@ -7,7 +7,8 @@ void
 Castro::construct_old_thermo_source(MultiFab& source, MultiFab& state, Real time, Real dt)
 {
   // we only include p divU in method of lines integration
-  if (time_integration_method == CornerTransportUpwind) return;
+  if (time_integration_method == CornerTransportUpwind ||
+      time_integration_method == SimplifiedSpectralDeferredCorrections) return;
```
maxpkatz: Same comment here about returning the logic: I would prefer
```
  if (!(time_integration_method == MethodOfLines ||
       time_integration_method == SpectralDeferredCorrections)) return;
```

maxpkatz:
Test suite results here: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2019-03-11-003/index.html

The SDC changes look fine but do you know why the Sedov MOL problems didn't change?

zingale:
> Test suite results here: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2019-03-11-003/index.html
> 
> The SDC changes look fine but do you know why the Sedov MOL problems didn't change?

I think it is simply because Sedov never needed to use internal energy.

Source/sources/thermo_nd.F90
zingale: this is now fixed

Source/sources/Castro_thermo.cpp
zingale: done

Source/sources/Castro_sources.cpp
zingale: done

Source/sources/thermo_nd.F90
```diff
@@ -35,25 +56,50 @@ subroutine ca_thermo_src(lo, hi, &
...
+
+             else if (coord_type == 1) then
+                ! axisymmetric
+                src(i,j,k,UEINT) = -FOURTH*(rp*(old_state(i+1,j,k,UMX)/old_state(i+1,j,k,URHO) + &
+                                                new_state(i+1,j,k,UMX)/new_state(i+1,j,k,URHO)) - &
+                                            rm*(old_state(i-1,j,k,UMX)/old_state(i-1,j,k,URHO) + &
+                                                new_state(i-1,j,k,UMX)/new_state(i-1,j,k,URHO)))/(r*dx(1))
+
+             else if (coord_type == 2) then
+                ! axisymmetric
```
maxpkatz: spherical

zingale: fixed

PR #546, opened 2019-03-11
Overhaul dimension agnostic BCs for GPU compatibility

maxpkatz:
This PR includes quite a few changes centered around making physical boundary conditions compatible for running on GPUs. Summary:

- The dimension agnostic ca_*fill interfaces have been moved up to C++, leaving the actual routines that do the work (*fill) in Fortran.

- The Fortran routines now take (lo, hi) as their first two arguments. They also now accept `time` by value rather than by reference.

- The declarations have been split into separate headers (some in `driver/`, some in `problems/`).

- We no longer do CUDA profiling on the initialization.

- For CUDA, the number of registers when running on the GPU is capped to 128 when doing BCs. This is because amrex_filccn needs to call `syncthreads()` as it works to ensure synchronization, and because of the way it is designed, can need as many as 8 threads per dimension == 512 total, which is too many registers required if 255 registers per thread are allowed. This requires some custom GNU Make rules in `Make.Castro`.


Exec/gravity_tests/DustCollapse/bc_fill_nd.F90
```diff
@@ -47,9 +45,9 @@ subroutine ca_hypfill(adv, adv_lo, adv_hi, &
     end if
 
     ! XLO
-    if ( bc(1,1,1) == FOEXTRAP .and. adv_lo(1) < domlo(1)) then
-       do k = adv_lo(3), adv_hi(3)
-          do j = adv_lo(2), adv_hi(2)
+    if ( bc(1,1,1) == FOEXTRAP .and. lo(1) < domlo(1)) then
+       do k = lo(3), hi(3)
```
zingale: does this change mean that you are not filling corners?

maxpkatz: It's not a functional change == `lo == adv_lo`, `hi == adv_hi`.

maxpkatz: The general direction we've been moving towards is that routines operate on (lo, hi) always. I could add a comment to this effect but I wonder if the issue is more broadly that we need better education/documentation on this new strategy?

PR #548, opened 2019-03-11
Offload gravity source terms onto the GPU

maxpkatz:


maxpkatz:
I am doing this outside of the context of #468 because it will take a bit of time to sync that branch with development, and we only need ConstantGrav offloaded to get going on some science problems.

zingale:
all tests pass

zingale:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2019-03-12-001/index.html

PR #551, opened 2019-03-13
Rad2TShock support to run 3d length wise in x y or z.

ChrisDeGrendele:
Idir specifies the direction, then the initial and boundary conditions will set up based on this. Also added inputs_y,z.M5 which is the problem inputs.M5 set up in y and z as well. 

[Testing notebook.pdf](https://github.com/AMReX-Astro/Castro/files/2962200/Testing.notebook.pdf)


ChrisDeGrendele:
The pdf attached is testing the analytical solution vs the 3 directions, as well as the initial set up.

guadabsb15:
So I have a few comments:
1) In Prob_3d.f90, are you using at all length_min?
2) Comment that might be silly, but just wanted to double check. When you ran the comparisons that you are showing on the pdf, you changed manually the idir parameter in the probin.M5?
3) on bc_fill_3d, in line 58-61 maybe is cleaner if you delete those code lines rather than commenting

Besides those minor things, it looks good to me

guadabsb15:
Another thing, could you please change back inputs.M5? 
to geometry.prob_hi   =    2000. 187.5. 187.5

Exec/radiation_tests/Rad2Tshock/Prob_3d.f90
```diff
@@ -79,38 +79,73 @@ subroutine ca_initdata(level,time,lo,hi,nscal, &
...
+           !Provides the simulation to be run in the x,y,or z direction
+           !where length direction is the length side in a square prism  
+           if (idir == 1) then
+              length_cell = xmin + delta(1) * (dble(i) + 0.5e0_rt)
+           else if (idir == 2) then
+              length_cell = ymin + delta(2) * (dble(j) + 0.5e0_rt)
+           else if (idir == 3) then
+              length_cell = zmin + delta(3) * (dble(k) + 0.5e0_rt)
+           else
+              stop "Invalid Direction, input integer int_dir = [1,3] ."
```
maxpkatz: Please use amrex_error("Invalid direction...") rather than the Fortran stop statement.

Exec/radiation_tests/Rad2Tshock/Prob_3d.f90
```diff
@@ -79,38 +79,73 @@ subroutine ca_initdata(level,time,lo,hi,nscal, &
...
+           endif
+
+          ! Old statement, replaced with conditional above. 
+          ! length_cell = length_min  + delta(1) * (dble(i) + 0.5e0_rt)
```
maxpkatz: Same comment as @guadabsb15 -- it's good to remove commented out old code, we always can revert back to an earlier version with git if we need to see what things used to be.

Exec/radiation_tests/Rad2Tshock/bc_fill_3d.F90
```diff
@@ -52,13 +52,13 @@ subroutine ca_hypfill(adv,adv_l1,adv_l2,adv_l3,adv_h1,adv_h2,adv_h3, &
...
        eint1 = rho1 * eos_state % e
-       etot1 = eint1 + 0.5*rho1*v1**2         
+       etot1 = eint1 + 0.5*rho1*v1**2
     end if
 
-    do n = 1,NVAR
-       call filcc(adv(adv_l1,adv_l2,adv_l3,n),adv_l1,adv_l2,adv_l3,adv_h1,adv_h2,adv_h3, &
-                  domlo,domhi,delta,xlo,bc(1,1,n))
-    enddo
+    !do n = 1,NVAR
```
maxpkatz: Since this looks to be the same code block as in the conditionals below, to reduce redundancy, can we uncomment this and then remove it from each of the cases below?

Exec/radiation_tests/Rad2Tshock/bc_fill_3d.F90
```diff
@@ -67,51 +67,154 @@ subroutine ca_hypfill(adv,adv_l1,adv_l2,adv_l3,adv_h1,adv_h2,adv_h3, &
...
-                   adv(i,j,k,UFX) = adv(i,j,k,URHO)           
-                end if
-                adv(i,j,k,UEINT) = eint1
-                adv(i,j,k,UEDEN) = etot1
-                adv(i,j,k,UTEMP) = T1
-             end do
-          enddo
-       enddo
-    end if
+    if (idir == 1) then !%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```
maxpkatz: The comment string here is unnecessary.

maxpkatz:
`amrex_probinit` should initialize `idir = 1` before the `fortin` namelist is read (this way we aren't relying only on the probins to set the value).

ChrisDeGrendele:
Lupe - 1. Nope, I'll take it out
2. Yes, I changed manually to run the tests
3. I'll take those out

guadabsb15:
The code doesn't compile anymore. 
In Prob_3d.f90, you don't need to declare idir, since you are already doing "use probdata_module".
In Prob_3d.f90, I believe you need to write "use amrex_error_module" under the subroutines ca_initrad and ca_initdata
After I locally did those changes the code compiles for me. I will run it now

ChrisDeGrendele:
So, for idir, I thought that as well. @maxpkatz mentioned that it is not standard to include default data in this file, and it is best to have it defaulted in the Prob_nd file as well?

guadabsb15:
I believe he meant to have idir = 1 not to declare it again (integer untin,i,idir)

ChrisDeGrendele:
Ah yes sorry that makes sense!

guadabsb15:
no problem! Looks good now

PR #552, opened 2019-03-13
Use sphinx-fortran to generate fortran documentation rather than doxygen

harpolea:
This changes the way the API for Fortran files is generated to use `sphinx-fortran` instead of Doxygen/`breathe`. This allows the Fortran docstrings to be cleaner and generates less-buggy documentation. 

The main things this PR does (other than re-format the Fortran docstrings) is:
- add the `sphinx-fortran` extensions & options to `conf.py` 
- adds `fortran_apidoc` to `make_api.py`, which goes through the Fortran files and creates the `.. f:automodule:: module` boilerplate (as `sphinx-apidoc` does for python packages).


zingale:
can we still use the same syntax for docstrings?

harpolea:
This uses a much simpler syntax, e.g. 
``` f90
subroutine foo(bar)
! this is my docstring

integer, intent(inout) :: bar

...

end subroutine
```

The main thing is that the docstring goes immediately after the prototype *without* a line in between

harpolea:
I'm going through now and changing existing docstrings to that

zingale:
nice job!

Issue #553, opened 2019-03-13
SDC and MOL don't need the sources to be fillpatched


zingale:
Since the source terms are not used in the predictor for hydro, I don't think we need to do a FillPatch on `sources_for_hydro` in either `Castro_advance_sdc.cpp` or `Castro_advance_mol.cpp`.

maxpkatz:
Agreed. The source terms for MOL/SDC only need to be valid on the valid domain, if the hydro is only generating an update on the valid domain (which is currently true for all hydro methods).

maxpkatz:
(Note that the FillPatch is also important for doing srctoprim.)

zingale:
neither of these methods need srctoprim

zingale:
I think there is a bit of an issue on the fillpatch in `do_old_sources`.  We call `do_advance_mol` with the _stage_ time, and that time is passed into `do_old_sources`, and hence used in the `FillPatch` in there.  But since for later stages that will be a intermediate between the old and new time, it will need to interpolate in time, but there is no valid new time-data.  Instead, what we need here, atleast for a single level, is to ensure the fill is only done using the old time (`old_source`) data.

It would actually be a lot cleaner if we removed the ghost fill from `do_old_sources`, since that only really makes sense for CTU integration.  Things are much more complicated for 4th order MOL/SDC.

maxpkatz:
I'm OK with moving the fillpatch out of do_old_sources and explicitly writing it in the CTU advance.

zingale:
The next step is to get the 4th order sources correctly done in MOL and CTU -- some of the fillpatches seem to use the node time instead of the time corresponding to the MF, so the time-interpolation is probably not working properly.

zingale:
fixed in #574

PR #554, opened 2019-03-14
remove the FillPatch from do_old_sources

zingale:
turns out a lot of times we don't need to FillPatch in do_old_sources.  When we do, we can simply do so after the call.  This increases the flexibility of things.


zingale:
This is part of #553

zingale:
all test suite problems pass.

Note: this will be followed by another PR that addresses how the fillpatches are done for SDC and MOL.

PR #557, opened 2019-03-19
eliminate castro.fourth_order.

### LINKED ISSUE #556: castro.fourth_order is redundant with sdc_order=4 or mol_order=4
The best option is to simply remove `castro.fourth_order`, and use the 4th order slopes whenever we run with `castro.mol_order=4` or `castro.sdc_order=4`.  

For all the other cases (including the odd `castro.mol_order=3`), we will continue to honor `castro.ppm_type` to select between PLM and PPM.

zingale:
4th order spatial reconstruction is now automatic based on the sdc or mol order

## PR summary

This eliminates the `castro.fourth_order` runtime parameter, instead inferring whether we need 
fourth order spatial reconstruction from the integrator's order for both MOL and SDC

This closes #556 

## PR checklist

### DIFF
--- a/Source/driver/Castro.cpp
+++ b/Source/driver/Castro.cpp
@@ -384,13 +384,6 @@ Castro::read_params ()
     if (time_integration_method != CornerTransportUpwind && use_retry)
         amrex::Error("Method of lines integration is incompatible with the timestep retry mechanism.");
 
-    // fourth order implies do_ctu=0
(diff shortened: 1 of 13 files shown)


zingale:
going to fix #558 first, so we have some 4th order tests in the suite, then run with this change before we want to merge.

zingale:
the new 4th order tests pass with this change too.

Issue #559, opened 2019-03-20
NaNs in new Castro hydro


asalmgren:
In transyz we define qmo from lo to hi and then in riemannus we use ql from lo to hi (where ql in riemannus is the same array as qmo in transyz) but the lo and hi have changed -- the call to transyz uses lo,hi from "xbx" but the call to cmpflx_plus_godunov (first call after transyz call) uses lo,hi from "cxbx"   cxbx, for idir = 1, has been grown by 1 in the y and z directions.  So riemannus tries to access into parts of the array that were not defined in transyz

zingale:
fixed

PR #560, opened 2019-03-20
fix box side in cmpflx call after transyz

zingale:
## PR summary

We were using the wrong box size in the call to `cmpflx` after `transyz` -- they should box operate on the same box.

## PR checklist


zingale:
closes #559

zingale:
all tests pass

Issue #356, opened 2018-07-05
Force new dt after regrid


maxpkatz:
Amr has a parameter compute_new_dt_on_regrid which is disabled by default. But we want this option to be on for most of the problems that are of interest to Castro, which can evolve rapidly and for which updating the timestep after generating new regrid data is probably wise. So we should force this option to be on by default in Castro, assuming that we can find a way to still allow the user to disable it if they really want to.

zingale:
we should just turn this on by default.

PR #580, opened 2019-03-31
Remove grav_source_type and rot_source_type parameters

### LINKED ISSUE #156: get rid of grav_source_type < 4
The older gravity source types are really not as good.  Certainly types 1 and 3 aren't ever used.  `grav_source_type = 2` might be useful for comparison purposes, but also could probably go.

maxpkatz:
## PR summary

The current default values (4 for both; the conservative energy formulation) are now the only option.

Closes #156

## PR checklist

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,9 @@
 # changes since the last release
 
+  -- The parameters castro.grav_source_type and castro.rot_source_type have
+     been removed, and the previous default values (4 in each case, corresponding
+     to the conservative energy formulation) are now the only choice.
+
   -- The runtime parameter castro.fix_mass_flux has been removed: it is not
      clear what the use case is, and it had no test suite coverage. (#572)
 
--- a/Exec/gravity_tests/evrard_collapse/inputs
+++ b/Exec/gravity_tests/evrard_collapse/inputs
@@ -21,7 +21,6 @@ castro.hi_bc       =  2   2   2
 castro.do_hydro = 1
 castro.do_grav  = 1
 castro.ppm_type = 0
-castro.grav_source_type = 4
 
 castro.small_dens = 1.0e-2
 castro.small_temp = 1.0e0
--- a/Exec/hydro_tests/rotating_torus/inputs_3d
+++ b/Exec/hydro_tests/rotating_torus/inputs_3d
@@ -30,9 +30,6 @@ castro.hybrid_hydro = 1
 
 castro.use_retry = 1
 
-castro.grav_source_type = 4
-castro.rot_source_type = 4
-
 # THERMODYNAMICS
 
 castro.dual_energy_eta1 = 1.0e-3
--- a/Exec/hydro_tests/rotating_torus/inputs_3d.test
+++ b/Exec/hydro_tests/rotating_torus/inputs_3d.test
@@ -30,9 +30,6 @@ castro.hybrid_hydro = 1
 
 castro.use_retry = 1
 
-castro.grav_source_type = 4
-castro.rot_source_type = 4
-
 # THERMODYNAMICS
 
 castro.dual_energy_eta1 = 1.0e-3
--- a/Exec/radiation_tests/RadFront/inputs2d
+++ b/Exec/radiation_tests/RadFront/inputs2d
@@ -59,11 +59,6 @@ castro.do_react       = 0       # reactions?
 castro.small_temp = 1.0e1
 castro.small_dens =1.0e-7 
 
-
-
-castro.grav_source_type = 4
-
-
 # TIME STEP CONTROL
 castro.cfl            = 0.8     # cfl number for hyperbolic system
 castro.init_shrink    = 0.1     # scale back initial timestep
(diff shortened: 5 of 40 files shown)


maxpkatz:
I would like to actually do this because it is my goal to implement fully conservative gravity and rotation options that are applied directly to the fluxes (for both momentum and energy). When I originally did the analysis of these source terms, I didn't see much benefit from doing it this way in terms of conservation properties, but I think it will be much better for code maintainability.

zingale:
`flame_wave` uses `grav_source_type = 2` currently.  We need to see if the problem works well with 4.  I vaguely remember there being issues with HSE, but that was a few years ago.

maxpkatz:
For constant gravity, the source term is very similar in both cases, except that the energy source term evaluates the time-centered (rho*v*g) at zone edges rather than zone centers.

The main case in which they would differ substantially is at a symmetry boundary, because the contribution on the boundary is zeroed out, as the fluxes are zeroed out.

zingale:
Note that the MOL and SDC options still use an explicit source term, not the conservative flux-based approach.  Also, it is not clear how to do this conservative approach to 4th order accuracy.

maxpkatz:
grav_source_type == 4 only matters for the corrector source, this shouldn't change anything for MOL/SDC.

zingale:
agreed, but it should be possible to do this for MOL/SDC second-order, since we have the fluxes at the time we need the source.

maxpkatz:
Yes. That is one of the things I will pursue in a later PR.

zingale:
so `castro.grav_source_type=4` crashes `flame_wave`.  Too many subcycles.

Is this conservative formulation axisymmetric-aware?

zingale:
I suspect the problem here is the hydrostatic boundary.  We need to figure out how to do an HSE boundary condition with conservative gravity.

maxpkatz:
Yes, the conservative formulation works for axisymmetric/spherical geometries.

maxpkatz:
@zingale can you provide a specific reproducer for the flame_wave issue?

zingale:
Running `flame_wave` with the following causes the "too many subcycles" error.

Works fine with `grav_source_type = 2`.

[inputs.noboost.500Hz.txt](https://github.com/AMReX-Astro/Castro/files/4004083/inputs.noboost.500Hz.txt)
[probin.noboost.500Hz.txt](https://github.com/AMReX-Astro/Castro/files/4004085/probin.noboost.500Hz.txt)

maxpkatz:
I was able to reproduce this after 3234 steps with

```
jsrun -n 8 -a 1 -g 1 ./Castro2d.pgi.MPI.CUDA.ex inputs.noboost.500Hz castro.grav_source_type=4 amr.max_grid_size=1152
```

The code is crashing because of CFL violations on the fine grids. They are so large that the default number of retry subcycles cannot handle it, hence the abort.

Examining the plotfile at the end of the previous step shows that all of the velocities are reasonable. The large velocity is not present there on any level. Rather, the large velocity is being generated during a regrid. Unfortunately this is diffcult to avoid: even if the regrid step is extremum-preserving, the extrema preserved are the momenta and density, not velocity (which is not a state variable), and so it's still possible to generate a large velocity as a result of an unlucky combination of (rho u) and (rho). I verified that if I set `amr.regrid_int=-1` and `amr.regrid_on_restart=0` on the restart then the issue went away.

My recommendation is that you use `amr.compute_new_dt_on_regrid=1` so that, if large velocities are generated by a regrid scheme, we catch this and drop the timestep to compensate for it. Actually, we do have issue #356 about making it on by default, so that would be another way to resolve the issue. I checked that this option resolves the crash for this case, and the timestep quickly returns to its original value thanks to change_max.

I don't know for sure why this happens only with grav.source_type = 4. But it's most likely just bad luck; this could happen regardless of other parameters, and it's the reason I have had `compute_new_dt_on_regrid` enabled in wdmerger for a while.

zingale:
There are still issues with flame_wave.  Using the above dt fix, the problem runs out longer but eventually hits a dt of 9e-15 (sometime after 27000 steps).

zingale:
The removal of the `hard_cfl_limit = 0` option in this run resolves that timestep crash (#723)

But the timestep is still about 4x smaller than the `grav_source_type = 2` version.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-05-27-002/index.html
lots of gravity compilation failures

zingale:
we still have the timestep issue with this change for `flame_wave` -- the timestep is about a factor of 2x smaller with `castro.grav_source_type = 4` as compared with `castro.grav_source_type = 2`

zingale:
Here's a comparison of the flame_wave with different `castro.grav_source_type`:

`castro.grav_source_type = 2`:

flame_wave_pslope_gst2

`castro.grav_source_type = 4`:

flame_wave_pslope_gst4

The biggest difference appears to be in the temperature structure at the top of the atmosphere.  This is where the density is going toward 0, so it makes sense that the mass fluxes might be doing something odd here which changes the energy.

zingale:
The `castro.grav_source_type = 4` run has more retries and has them due to both "timestep validity check failed;" and "invalid density; proceeding to a retry.".  The `castro.grav_source_type = 2` run only has it due to timestep validity.  And has fewer retries overall despite evolving for 4.8x longer (in simulation time)

zingale:
Note this is with CFL = 0.8, so the timestep validity retries are not unexpected.

PR #585, opened 2019-04-03
a bunch of optimizations:

zingale:
Still need to test convergence with all of this

## PR summary

 * we no longer recreate the sources for m = 0 on the later iterations

  * we skip center <-> average conversions if the state is 0

  * we no longer do the final source computation except for the last iteration

  * we use Sborder for the initial burn instead of filling Sburn -- this changes
    answers since we do a clean state on Sborder and were not doing one on
    Sburn

## PR checklist


zingale:
this is ready for review.  There are some small changes with this, but I previously confirmed that I still converge to 4th order in reactions.

Source/hydro/fourth_order.F90
```diff
@@ -448,24 +448,30 @@ subroutine ca_make_cell_center(lo, hi, &
     endif
 
     do n = 1, nc
-       do k = lo(3), hi(3)
-          do j = lo(2), hi(2)
-             do i = lo(1), hi(1)
 
-                lap = U(i+1,j,k,n) - TWO*U(i,j,k,n) + U(i-1,j,k,n)
+       if (maxval(abs(U(U_lo(1)-1:U_hi(1)+1, U_lo(2)-dg(2):U_hi(2)+dg(2), U_lo(3)-dg(3):U_hi(3)+dg(3), n))) == ZERO) then
```
maxpkatz: Is this change for performance or correctness? If it's performance, I'm not sure it's actually worth it, I suspect we won't save that much since we'll still have to pull in all of the data from memory and there's only ~3 flops/byte in this conversion, so this calculation is probably always memory bandwidth bound.

zingale: the idea was performance, but I had the same concern as you.  I'll revert that one.

zingale: I removed them all -- tests showed no performance difference

zingale:
I think this is ready for merge...

PR #562, opened 2019-03-21
Mol cuda

zingale:
## PR summary

This rewrites the CPU second-order (PLM/PPM) MOL hydro to be GPU friendly, and eliminates the need for a separate code path for the CUDA MOL.

## PR checklist

- [ ] test suite needs to be run on this PR
- [ ] this PR will change answers in the test suite
- [ ] all functions have docstrings as per the coding conventions

PR #588, opened 2019-04-04
sdc do_hydro

zingale:
## PR summary

This implemented the `castro.do_hydro` parameter for the SDC 2nd-order solver

## PR checklist


zingale:
blocked by #562 

also... still need to implement the 4th order version.

zingale:
answers will change for SDC diffusion problems, since this fixes how the diffusion conductivities are computed.

zingale:
this is ready for review

maxpkatz:
Is it not possible to do this more cleanly by just not calling construct_mol_hydro_source if !do_hydro?

zingale:
well... we still need to put the source term into the update.  But indeed, I suppose I could just copy the source over into the update in that case...

zingale:
actually though, I want to experiment with doing diffusion via fluxes, so I still will want to get through this part of the code path.

maxpkatz:
If we do diffusion via fluxes, it seems like there'd be no realistic way to do that without essentially doing the hydro update, since you still need to do the Riemann solve to have the data on edges. It would still be possible to disable do_hydro in that case, perhaps by calculating the hydro update but not applying it to the state. But that would still be done from the advance routine.

Issue #573, opened 2019-03-24
Separate the tracing of the passive and thermodynamic variables


maxpkatz:
We have to carry around NQ variables in the construction of the final interface states, for Ip/Im (and similarly, NQSRC variables for Ip_src/Im_src). By splitting the thermodynamic and passive variables into separate arrays, there's two potential optimizations. First, we only require one wave (the middle wave) for the passive variables, so the second-to-last dimension can be size 1, not size 3. Second, we could configure the PPM reconstruction and tracing to be done one passive variable at a time, so that the passive Ip/Im could be size 1 in the last dimension, not size npassive. This optimization is necessary for being able to scale the PPM algorithm for large networks.

zingale:
The strategy here is to break q into 3 groups:

  * `q_core`: `NQC` comps, `NQCSRC` source comps
  * `q_passive`: `NQP` comps, `NQPSRC` source comps (which may be 0 if we are not considering passive sources)
  * `q_rad`: `NQR` comps (there are no sources for this).

Eventually, we will rename `q_core` -> `q`

zingale:
For the 4th order, we would need a q_bar version of each.

zingale:
important: as part of this, we need to take the velocities out of `qpass_map`

zingale:
there is a better way to save memory without the complexity of separating the primitive state.  This is to do the parabolic reconstruction and intergration "on demand" as done now on the `actual_ppm` branch.

PR #589, opened 2019-04-04
replace primitive state q with q_core, q_pass, and q_rad

zingale:
## PR summary

The splits the primitive state variable q into pieces to help cut down memory overhead as discussed
in #573 

## PR checklist


zingale:
Most things work, except old SDC

Still need to do:

- [ ] remove the extra waves from Im_pass, Ip_pass and the sources (except if we require them for ppm_temp_fix)
- [ ] in the passive tracing, we don't need to use `qpass_map` anymore, we can just loop over `NQP`

zingale:
there are some tiny roundoff diffs for 2d problems with rotaiton (O(1.e-19)), but otherwise, all tests pass.  This is ready for review

zingale:
also note that as part of this PR, we no longer include QU, QV, and QW in the qpass_map.

zingale:
we decided against this route in favor of doing the parabolic stuff on demand.

PR #593, opened 2019-04-08
SDC tols

zingale:
## PR summary

This splits the tolerance that the nonlinear solve for SDC reactions uses to have separate values for density, species, and energy.

## PR checklist


zingale:
blocked by #585

zingale:
this also now implements an absolute tolerance that helps a lot with species integration

zingale:
reacting_convergence still gets 4th order with these changes

zingale:
this is ready for review.  I am happy with the convergence with these changes.

Source/driver/sdc_util.F90
```diff
@@ -248,8 +265,22 @@ subroutine sdc_solve(dt_m, U_old, U_new, C, sdc_iteration)
...
        endif
 
+       ! relative tolerances
+       rtol(0) = tol_dens
+       rtol(1:nspec_evolve) = tol_spec
+       rtol(nspec_evolve+1) = tol_ener
+
+       ! absolute tolerances
+       atol(0) = sdc_solver_atol * U_old(URHO)
+       atol(1:nspec_evolve) = sdc_solver_atol * U_old(URHO)   ! this way, atol is the minimum x
```
maxpkatz: I don't follow this choice or the comment -- can you elaborate?

zingale: we are integrating (rho X), so I need to turn what I want the minimum value of X to track into the atol for (rho X).

maxpkatz: OK. But since sum(rho X) = rho, isn't this scaling based on X == 1?

maxpkatz: OK, I get it now.

PR #596, opened 2019-04-11
Sdc hybird

zingale:
## PR summary

This adds the ability to try Newton solve first and if a zone doesn't converge, it retries with VODE

## PR checklist


zingale:
blocked by #593

zingale:
This is ready for review.  It does 3 things:

  * adds a hybrid solver for the nonlinear SDC solve, that tries Newton first and then VODE if that fails
  * uses a better initial guess for T in all the solves
  * adjusts the flame problem setup and adds a enuc limiter

PR #599, opened 2019-04-13
Fix bug in post-reflux position-dependent source terms

maxpkatz:
## PR summary

The algorithm for recalculating source terms after an AMR reflux did not set some data needed to correctly calculate positions on the fine levels, so position-dependent source terms (rotation, hybrid momentum) were being calculated incorrectly. This caused some strange effects in AMR simulations with rotation, such as binary star orbits getting wider with time and drifting relative to the system center of mass. This has now been fixed.

The issue is that the position() function implicitly needs to know what AMR level we're on to use the correct dx spacing for that level, so we need to call ca_set_amr_info() whenever we're looping over levels.

## PR checklist


zingale:
is the main bit of information that was needed from the `ca_set_amr_info()` call the level?

maxpkatz:
Yes, we only need the level (that tells us what dx to use).

PR #603, opened 2019-04-19
Enforce a speed limit in the Riemann solver

maxpkatz:
## PR summary

Castro is a Newtonian hydrodynamics code and has no knowledge of special relativity. This means that spuriously large, unphysical velocities can be generated. This commonly occurs in the wdmerger problem at poorly resolved density gradients near the edge of the stars. Although a proper solution to this problem involves adding SR to the code, a simple approach to improving the situation is to enforce a speed limit in the Riemann solve. By default this speed limit is c, but the user can lower it (for example, to a value like 0.1c which is still in the relativistic regime where Castro results should be distrusted) if they know the range of realistic values taken by the velocity for their problem.

## PR checklist

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,12 @@
 # changes since the last release
 
+  -- The velocity calculated for the interface in the Riemann solve for
+     the CGF/CG Riemann solvers can no longer exceed the speed of light.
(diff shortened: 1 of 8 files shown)


maxpkatz:
Not sure if it makes sense to do this for HLLC.

Issue #369, opened 2018-07-16
include diffusion in fluxes correction for MOL?


zingale:
For the MOL (and SDC) integration, it is possible to include the diffusive flux in the flux registers that get used for the flux correction.  This may not be a big deal, since we are typically refined where diffusion is acting the most (e.g. flames).

zingale:
although... I think the Laplacian operator we are using currently knows about C-F interfaces directly

zingale:
this will be taken care of as part of #495

zingale:
this is being implemented in `sdc_diffusion`, along with 4th order accurate diffusion.

zingale:
this is implemented in #610

PR #610, opened 2019-05-10
Implement diffusion (including 4th order) for MOL and SDC + SDC solver fixes

zingale:
## PR summary

This implements a flux-based thermal diffusion term for MOL and SDC in both 2nd and 4th order.  It also cleans up the SDC solvers, uses a better error estimate for Newton, adds a multi-newton option, and various bug fixes.

## PR checklist


Source/driver/Castro_advance_sdc.cpp
```diff
@@ -203,8 +203,15 @@ Castro::do_advance_sdc (Real time,
     // update to
     if (m < SDC_NODES-1) {
       do_sdc_update(m, m+1, dt); //(dt_sdc[m+1] - dt_sdc[m])*dt);
+
+      // we now have a new value of k_new[m+1], do a clean_state on it
+      MultiFab::Copy(S_new, *(k_new[m+1]), 0, 0, S_new.nComp(), 0);
```
maxpkatz: We can't call clean_state directly on k_new[m+1]?

Source/driver/Castro.cpp
```diff
@@ -1075,39 +1075,79 @@ Castro::initData ()
...
-             const Box& box     = mfi.validbox();
+             const Box& box = mfi.validbox();
 
              ca_make_fourth_in_place(BL_TO_FORTRAN_BOX(box),
                                      BL_TO_FORTRAN_FAB(Sborder[mfi]));
            }
 
-         // now copy back the averages
-         MultiFab::Copy(S_new, Sborder, 0, 0, NUM_STATE, 0);
+         // now copy back the averages for UEDEN and UTEMP only
```
maxpkatz: It looks like this comment should be corrected, UEINT and UTEMP only

Source/driver/Castro_util.F90
```diff
@@ -128,6 +128,63 @@ subroutine ca_enforce_consistent_e(lo,hi,state,s_lo,s_hi) bind(c,name='ca_enforc
...
+             eos_state % e = state(i,j,k,UEINT) * rhoInv - HALF * (u*u + v*v + w*w)
+             eos_state % xn(:) = state(i,j,k,UFS:UFS-1+nspec) * rhoInv
+
+             call eos(eos_input_re, eos_state)
+
+             state(i,j,k,UTEMP) = eos_state % T
+
+             state(i,j,k,UEINT) = eos_state % rho * eos_state % e
+
+             !state(i,j,k,UEDEN) = state(i,j,k,UEINT) + &
```
maxpkatz: If this is not done, can you remove the comment?

Source/driver/Castro_util.F90
```diff
@@ -128,6 +128,63 @@ subroutine ca_enforce_consistent_e(lo,hi,state,s_lo,s_hi) bind(c,name='ca_enforc
...
+    ! Local variables
+    integer  :: i,j,k
+    real(rt) :: u, v, w, rhoInv
+
+    type(eos_t) :: eos_state
+
+    !$gpu
+
+    !
+    ! Enforces (rho E) = (rho e) + 1/2 rho (u^2 + v^2 + w^2)
```
maxpkatz: This comment should be removed

Source/driver/sdc_util.F90
```diff
@@ -42,19 +51,133 @@ module sdc_util
...
+    integer, intent(in) :: sdc_iteration
+
+    integer :: ierr
+
+
+    if (sdc_solver == NEWTON_SOLVE) then
+       ! we are going to assume we already have a good guess for the
+       ! solving in U_new and just pass the solve onto the main Newton
+       ! solve
+       call sdc_newton_solve(dt_m, U_old, U_new, C, sdc_iteration, ierr)
```
maxpkatz: Similar to what we do for the CTU subcycling, it should be possible to have sdc_newton_subdivide be the main entry routine, which starts with nsub=1 and then takes more as needed. This makes the code slightly cleaner.

Source/driver/sdc_util.F90
```diff
@@ -103,248 +222,144 @@ subroutine sdc_solve(dt_m, U_old, U_new, C, sdc_iteration)
...
 
-          istate = 1
-          iopt = 1
+       dU_react(:) = f_rhs(:)
 
-          iwork(:) = 0
+       ! how much of dU_react should we apply?  We don't want species
+       ! going negative.  Here, the 1/2 is a safety factor
+       eta = ONE
+       ! do n = 1, nspec_evolve
```
maxpkatz: Is this part of the code still a work in progress for future PRs, or do you think we're doing the right thing here now?

Source/hydro/Castro_mol_hydro.cpp
```diff
@@ -99,9 +100,7 @@ Castro::construct_mol_hydro_source(Real time, Real dt, MultiFab& A_update)
           }
 
 #if (AMREX_SPACEDIM <= 2)
-          if (!Geom().IsCartesian()) {
-            pradial.resize(amrex::surroundingNodes(bx,0),1);
-          }
+          pradial.resize(amrex::surroundingNodes(bx,0),1);
```
maxpkatz: Was this removal of the !IsCartesian() guard intentional?

Source/driver/Castro_advance_sdc.cpp
zingale: huh... I think you are right.  Maybe I was thinking we needed a ghost fill, but that'll happen inside clean_state.

zingale: fixed.

Source/hydro/Castro_mol_hydro.cpp
zingale: yes, the code has valgrind errors otherwise.

Source/driver/sdc_util.F90
zingale: I think that was an experiment that I no longer need

Source/driver/sdc_util.F90
zingale: yes, that makes sense.

zingale: this has been fixed

zingale:
I think everything has been fixed.  I will run the CPU and GPU suite once more on this branch and then merge.

zingale:
this closes #369

Source/hydro/Castro_mol_hydro.cpp
maxpkatz: It seems the difference here is that in CTU we only use pradial in the C++ hydro inside !IsCartesian() regions, but we're passing it ca_fourth_single_stage either way.

PR #611, opened 2019-05-12
Tagging

maxpkatz:
## PR summary

The tagging routines have been reformulated so that they run on the GPU. Since the tags_and_untags routine in AMReX is not GPU-accelerated, we opt to directly fill in the TagBoxArray in the tagging routines. Since these are based on char data, which doesn't seem to work well with PGI CUDA Fortran, we pass the TagBox to Fortran as an int8_t.

Castro_prob_err_list.H and other related files have been deleted as they were not actually used anywhere.

Castro_error.cpp is now removed and there is no further support for writing custom tagging routines. The set of variables that we check for tagging is hard-coded in Castro and can be expanded as needed. Problem-specific tagging should be done through the set_problem_tags functionality.

## PR checklist


maxpkatz:
Ready for review

maxpkatz:
Defer until after 19.06 release

zingale:
is this good for review now?

zingale:
all tests pass

PR #614, opened 2019-05-21
Prepare Castro for upcoming changes in AMReX

WeiqunZhang:
There are some upcoming changes in AMReX's Geometry class. This pull-request makes Castro work with the current AMReX development branch and the future version.  I have tested this in CCSE's regression test suite.

### DIFF
--- a/Source/driver/Castro.cpp
+++ b/Source/driver/Castro.cpp
@@ -278,18 +278,20 @@ Castro::read_params ()
         phys_bc.setHi(i,hi_bc[i]);
     }
 
+    const Geometry& dgeom = DefaultGeometry();
+
     //
     // Check phys_bc against possible periodic geometry
     // if periodic, must have internal BC marked.
     //
-    if (Geometry::isAnyPeriodic())
+    if (dgeom.isAnyPeriodic())
     {
         //
         // Do idiot check.  Periodic means interior in those directions.
         //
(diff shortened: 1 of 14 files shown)


zingale:
I'll run out suite and then merge

WeiqunZhang:
I updated `Util/ConvertCheckpoint`, but I didn't test it.  However, it seems that it has not been used for a long time.  It didn't actually compile before my change.  It compiles now.

Source/driver/Castro.cpp
```diff
@@ -753,12 +755,12 @@ Castro::initMFs()
 	mass_fluxes[dir].reset(new MultiFab(get_new_data(State_Type).boxArray(), dmap, 1, 0));
 
 #if (BL_SPACEDIM <= 2)
-    if (!Geometry::IsCartesian())
+    if (!Geom().IsCartesian())
```
maxpkatz: What's the difference between DefaultGeometry() and Geom()?

WeiqunZhang: `Geom()` is a member function of `AmrLevel`, whereas `DefaultGeometry` is a free function in amrex namespace.  For `IsCartesian()`, both are fine.  But for things like `CellSize`, one has to use `Geom()` or `Geom(int level)` in `Amr` class.

PR #616, opened 2019-06-05
Development

taehoryu:
Planet code update (radiation)


zingale:
you have some conflicts, so you should merge Castro's development into your fork and then push to resolve them.  There are a lot of changes to particles here, which hopefully will go away when you resolve the differences.  We want this PR just to be about your problem setup.

PR #617, opened 2019-06-05
planet update (radiation)

taehoryu:


Exec/science/planet/Prob_2d.f90
```diff
@@ -175,7 +178,7 @@ subroutine ca_initdata(level,time,lo,hi,nscal, &
   do j = lo(2), hi(2)     
      do i = lo(1), hi(1)   
         
-        state(i,j,UEDEN) = state(i,j,URHO) * state(i,j,UEINT) 
+        state(i,j,UEDEN) = state(i,j,URHO) * state(i,j,UEDEN) 
         state(i,j,UEINT) = state(i,j,URHO) * state(i,j,UEINT)
```
zingale: I think this change is wrong.  I'll fix it after the merge

PR #624, opened 2019-06-25
eliminate arrays holding ppm integrals

zingale:
## PR summary

This gets rid of the multidimension `Ip` and `Im` arrays that hold the integrals under the parabolas for each characteristic wave.  Now, instead, we do the reconstruction and integration "on demand" in the trace routine.  This saves considerable memory.

## PR checklist


zingale:
All tests pass.  There was a failure with the `ppm_temp_fix = 3` test, but that was actually a problem on `development`, and it should be good now.

PR #627, opened 2019-06-28
split source fill into its own routine

zingale:
and we now use the source for ... sources!  This makes it
easier to override them.  Also, add checks for external BCs
and set to outflow for all quantities.

## PR summary

We were using the `generic_fill` stuff for filling the source terms,  But this is potentially bad, since we also use those routines for other state data.  This makes source filling its own set of routines.  I've also added checks on EXT_DIR and set to first-order extrapolate there to be sure we handle that case, e.g., with HSE BCs.

## PR checklist


zingale:
tests passed, so I am going to merge

PR #631, opened 2019-07-11
move interpolate_sub into model_parser_module

### LINKED ISSUE #615: consolidate the function and subroutine interoplate() routines
We have both a function and subroutine interface for the interpolation in `Source/driver/interpolate.F90`.  The latter is probably more general and written for GPU offloading, so we should convert all of the problem setups to use the subroutine version and then ditch the function.

zingale:
## PR summary

closes #615 

## PR checklist


zingale:
all tests passed.  This is ready for review

PR #632, opened 2019-07-12
change the species tracing

### LINKED ISSUE #623: simplify species tracing with ppm?
We currently trace the species as:

```
if (un > ZERO) then
   qp(i,j,k,n) = q(i,j,k,n)
else
   qp(i,j,k,n) = Im(2,n)
end if
if (n <= NQSRC) qp(i,j,k,n) = qp(i,j,k,n) + HALF*dt*Im_src(2,n)
```

but do we really need that if-test?  I think we already do that check in the integration / construction of Im

zingale:
we now just use the limit of the parabola if the wave moves away from the interface

## PR summary

This simplifies the species tracing in the PPM code.  Since we already take the limit of the parabolic interpolant if the wave is not moving toward the interface in the ppm_integrate routine, there is no reason to look again at the direction of the wave in the tracing routine.  This is more consistent with how we treat the other variables.

closes #623 

## PR checklist


zingale:
tests are here:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2019-07-11-001/index.html

PR #633, opened 2019-07-12
switch to interior stencils at physical boundaries

### LINKED ISSUE #293: use modified stencil at physical boundaries for higher order solvers
For higher order solvers, we should modify the stencil at physical boundaries to only use interior cells.  See McCorquodale & Colella 2011 for an example.

zingale:
## PR summary

This adopts the one-sided stencils from McCorquodale and Colella at physical boundaries.  With this change we get 4th order convergence for hydrostatic atmospheres with reflecting BCs.

This closes #293 

## PR checklist


zingale:
tests are good.  Only 3 SDC problems change, and are expected.  This is ready for review.

Source/hydro/fourth_order.F90
```diff
@@ -56,375 +59,849 @@ subroutine states(idir, &
...
+    real(rt) :: lap
+
+    lapx = ZERO
+    lapy = ZERO
+    lapz = ZERO
+
+    ! we use 2nd-order accurate one-sided stencils at the physical boundaries
+
+    if (i == domlo(1) .and. physbc_lo(1) /= Interior) then
+       lapx = 2.0_rt*a(i,j,k,n) - 5.0_rt*a(i+1,j,k,n) + 4.0_rt*a(i+2,j,k,n) - a(i+3,j,k,n)
```
maxpkatz: This one-sided second-order stencil looks correct to me, but I notice that McCorquodale and Colella suggest a different approach: if I read it correctly, set the Laplacian here to just be the Laplacian at domlo(1) + 1. I like what you've done more. Any thoughts here? Can you add a comment explicitly saying why you did it differently than the paper so we don't trip up ourselves up on this later?

zingale: I basically didn't like MC's approach here.  I say so in the SDC paper.  I think my approach is better, but I agree, I should add a comment noting the difference.

Source/hydro/fourth_order.F90
```diff
@@ -56,375 +59,852 @@ subroutine states(idir, &
...
-                do i = lo(1)-3, hi(1)+3
-                   d2ac(i,j,k) = a(i-1,j,k,n) - 2.0_rt*a(i,j,k,n) + a(i+1,j,k,n)
-                enddo
-             enddo
-          enddo
+
+                ! reset the left state at domlo(1) if needed -- it is outside the domain
+
+                if (physbc_lo(1) == Outflow) then
+                   al(domlo(1),j,k,:) = ar(domlo(1),j,k,:)
```
maxpkatz: Is this fourth order accurate?

zingale: I haven't tested the convergence of this, only the reflecting.  reflection was 4th order accurate.  I suspect this is as well, since we are basically telling the Riemann problem that the state is outflow.

maxpkatz: We've decided to defer this to a future release.

PR #637, opened 2019-07-13
Castro error  wrapper

### LINKED ISSUE #244: Create Castro-owned error wrapper
Rather than calling bl_error directly, we should create a castro_error function that by default wraps bl_error. This will allow us to more easily handle these error calls when we switch to CUDA: at first we can simply ifdef out the bl_error call in one central location, until we find a better solution.

zingale:
## PR summary

This adds a `castro_error()` routine that simply calls `amrex_error()`.  This will allow us to hook into the error routine for GPUs.

closes #244 

## PR checklist

### DIFF
--- a/Exec/hydro_tests/Sod/Prob_nd.F90
+++ b/Exec/hydro_tests/Sod/Prob_nd.F90
@@ -118,7 +118,7 @@ subroutine ca_initdata(level, time, lo, hi, nscal, &
                        state, s_lo, s_hi, &
                        dx, xlo, xhi)
 
-  use castro_error_module, only: amrex_abort
+  use castro_error_module, only: castro_error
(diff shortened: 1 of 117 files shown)


Exec/hydro_tests/Sod/Prob_nd.F90
```diff
@@ -118,7 +118,7 @@ subroutine ca_initdata(level, time, lo, hi, nscal, &
                        state, s_lo, s_hi, &
                        dx, xlo, xhi)
 
-  use amrex_error_module, only: amrex_abort
+  use castro_error_module, only: amrex_abort
```
maxpkatz: I think your sed-fu didn't quite work out here :-)

PR #641, opened 2019-07-16
remove the transverse velocities from the passive map

zingale:
## PR summary

This removes QV and QW from the pass_map stuff for looping over passives.  This makes the code overall more consistent.  Note: in some places were we doing the same thing twice (overwriting the previous), with different floating point ops (multiplying by 1/rho instead of dividing by rho).

## PR checklist


zingale:
closes #641

zingale:
this is ready for review.  Some problems with rotation and 2-d will have roundoff diffs.  There is one optimization commented out, in ctoprim, that we can restore, which reuses rhoinv instead of dividing by rho for the species.

PR #642, opened 2019-07-17
some PPM optimizations

zingale:
## PR summary

This implements various optimizations in the PPM solver

## PR checklist


zingale:
this is updated now -- I still think it's a good idea.

Source/hydro/riemann_nd.F90
```diff
@@ -1358,57 +1358,32 @@ subroutine riemannus(ql, ql_lo, ql_hi, &
...
-                gamco_g = gamcgl
-#endif
-
-             else if (ustar < ZERO) then
-                ro = rr
-                uo = ur
-                po = pr
-                reo = rer
-                gamco = gamcr
+             sgnm = sign(ONE, ustar)
```
maxpkatz: I don't think this works correctly if ustar == 0.

Source/hydro/trace_ppm.F90
```diff
@@ -522,31 +514,16 @@ subroutine trace_ppm_rhoe(lo, hi, &
...
                 ! paper (except we work with rho instead of tau).  This is
                 ! simply (l . dq), where dq = qref - I(q)
 
-                alpham = HALF*(dptotm*(ONE/(rho_ev*cc_ev)) - dum)*(rho_ev/cc_ev)
-                alphap = HALF*(dptotp*(ONE/(rho_ev*cc_ev)) + dup)*(rho_ev/cc_ev)
-                alpha0r = drho - dptot/csq_ev
-                alpha0e_g = drhoe_g - dptot*h_g_ev  ! note h_g has a 1/c**2 in it
+                alpham = HALF*(dptotm*rho_ref_inv*cc_ref_inv - dum)*rho_ref*cc_ref_inv
+                alphap = HALF*(dptotp*rho_ref_inv*cc_ref_inv + dup)*rho_ref*cc_ref_inv
+                alpha0r = drho - dptot/csq_ref
```
maxpkatz: Since we're doing this divide by csq_ref twice now, should it have its own inverse variable too?

Source/hydro/riemann_nd.F90
zingale: you're right.  I added a fix there.

PR #643, opened 2019-07-22
NERSC Hackathon work

jaharris87:
* Mostly bug fixes
* Some updates to reflect changes in thornado/weaklib interface
* GNU Makefile updates to allow GPU build of thornado/weaklib w/o GPU Castro/AMReX


Source/driver/meth_params.F90
```diff
@@ -801,8 +886,6 @@ subroutine ca_set_castro_method_params() bind(C, name="ca_set_castro_method_para
        gravity_type_int = 1
     else if (gravity_type == "PoissonGrav") then
        gravity_type_int = 2
-    else if (gravity_type == "PrescribedGrav") then
-       gravity_type_int = 3
```
zingale: why is this removed?

jaharris87: I'm not sure---I had made a change to `parse_castro_params.py`, and re-ran it with the change. Perhaps I didn't provide correct arguments to the script?

maxpkatz: Re-run the mk_params.sh script which invokves parse_castro_params.py for you.

zingale: okay... maybe someone made the change directly to that file and not the template in `development`.  Let me take a look.

jaharris87: > Re-run the mk_params.sh script which invokves parse_castro_params.py for you.

That's what I did initially.

maxpkatz: I updated the meth_params.template very recently, so perhaps you just need to sync again with development.

jaharris87:
Merged `development` and re-ran the `mk_params.sh` script. Does this look okay now?

zingale:
now it says there are 209 files changed -- is that what you expect?

PR #645, opened 2019-07-27
Fix the double Mach reflection problem so it compiles for GPUs

maxpkatz:
## PR summary

The double Mach reflection problem did not compile with PGI due to the use of real parameter arrays (it complained about implied do loops). I replaced them with expressions calculated in the loop. I also fixed issues in the problem-specific bc_ext_fill_nd.F90 which prevented it from compiling with CUDA.

## PR checklist

### DIFF
--- a/Exec/hydro_tests/double_mach_reflection/Prob_nd.F90
+++ b/Exec/hydro_tests/double_mach_reflection/Prob_nd.F90
@@ -21,8 +21,7 @@ subroutine amrex_probinit (init,name,namlen,problo,probhi) bind(c)
 
   namelist /fortin/ p_l, u_l, v_l, rho_l, rhoe_l, &
                     p_r, u_r, v_r, rho_r, rhoe_r, &
-                    T_l, T_r, frac, idir, &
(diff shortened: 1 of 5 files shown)


Exec/hydro_tests/double_mach_reflection/Prob_nd.F90
```diff
@@ -150,36 +170,33 @@ subroutine ca_initdata(level, time, lo, hi, nscal, &
...
-              y = ycen + 0.5d0*delta(2)*gp(jj)
+           do jj = -1, 1
+              if (jj == 0) cycle
+              y = ycen + HALF * delta(2) * (dble(jj) / sqrt(THREE))
 
-              do ii = 1, 2
-                 x = xcen + 0.5d0*delta(1)*gp(ii)
+              do ii = -1, 1
+                 if (ii == 0) cycle
+                 x = xcen + HALF * delta(1) * (dble(ii) / sqrt(THREE))
```
zingale: why did this change?

PR #649, opened 2019-08-03
some cleanups of the fourth order reconstruction

zingale:
## PR summary

This does some basic cleanups of the fourth-order reconstruction to eliminate arrays in prep for GPUing.  It also reuses the shk and flatn computation done in C++ for MOL.

## PR checklist


zingale:
The fourth-order tests in the suite all pass with these changes

zingale:
I'm also using this branch to work on the 4th order outflow BCs.

Source/hydro/fourth_order.F90
```diff
@@ -93,6 +93,28 @@ subroutine states(idir, &
...
+                      ! now use the 1-sided stencil from above with
+                      ! this extrapolated value
+                      a_int(i,j,k) = (1.0_rt/12.0_rt)*(25.0_rt*ac - 23.0_rt*a(i+1,j,k,n) + &
+                                                       13.0_rt*a(i+2,j,k,n) - 3.0_rt*a(i+3,j,k,n))
+
+                   else if (i == domhi(1)+2 .and. physbc_hi(1) == Outflow) then
+                      ! extrapolate to the domhi(1)+1 cell using a
+                      ! conservative cubic polynomial averaged over
+                      ! the cell
+                      ac = 4.0_rt*a(domhi(1),j,k,n) - 6.0_rt*a(domhi(1)-1,j,k,n) + 4.0_rt*a(domhi(1)-2,j,k,n) - a(domhi(1)-3,j,k,n)
```
maxpkatz: The check here is on `i == domhi(1)+2` but we're using data starting at `domhi(1)`, not `domhi(1)+1` -- is this right?

Source/hydro/fourth_order.F90
```diff
@@ -532,96 +554,84 @@ subroutine states(idir, &
           ! the limiting loops are now over zones
           if (limit_fourth_order == 1) then
 
+             ! this is a loop over cell centers, affecting
+             ! j-1/2,R and j+1/2,L
```
maxpkatz: This comment needs to be fixed since this is the k loop.

Source/hydro/fourth_order.F90
zingale: Yes, but this is a little tricky.  When we limit, then to get the right state for interface domhi+1, we need to have both of the interface states that cell domhi+1 sees, so this is domhi+1/2.R and domhi+3/2,L, which would have indices in the code domhi+1 and domhi+2 respectively.  So for outflow, I decided to extrapolate using a 4th order conservative interpolant to get that outer interface.

Source/hydro/fourth_order.F90
zingale: fixed

zingale:
This has been updated and is ready for review and merge

Issue #652, opened 2019-08-19
For PLM MOL/SDC, handle reflect BCs with the interface states


zingale:
For the PLM reconstruction used with second-order SDC, we can enforce the reflecting boundary on the interface states, like we do with the fourth-order reconstruction.

zingale:
this is implemented on `xrb_sdc`

Issue #598, opened 2019-04-11
implement react_T_min/max and react_rho_min/max for SDC


zingale:
If we are outside of these bounds, we can skip the nonlinear SDC solve and just do the advective update.

zingale:
the main difficulty here is that for the 4th order, we go into the reaction solve with a cell-centered "C", so we need a way to just bail out and do the update with the reactions set to 0.  It might be fine to just set R = 0, since we redo the conservative update.  There will still be the cost of the <C> -> C transformation, but we'll skip the whole nonlinear solve.

zingale:
to be consistent, we also need to do this in `construct_old_react_source`

zingale:
we also need this logic in `ca_instantaneous_react`

zingale:
maybe it would be useful to have a function `okay_to_burn(U)` that returns true if we should do reactions and false otherwise.  This way we don't need to replicate this logic all over the place.

Issue #294, opened 2018-02-10
add well-balanced scheme to MOL solvers


zingale:

zingale:
the Kappeli well-balanced scheme has been added to the PLM reconstruction for MOL/SDC on the `xrb_sdc` branch in PR #654 

For the 4th order SDC, it is not necessary, since we do really well on our own.

PR #654, opened 2019-08-19
some SDC improvements

zingale:
## PR summary

* Implements reflecting BCs for PLM reconstruction with SDC (issue #652)

* Implements the density and temperature limits to burning for SDC (issue #598)

* Fixes the weighting of the flux registers for SDC (#658)

* Implements a well-balanced scheme for 2nd order PLM reconstruction (#294)

* Also adapts the bubble_convergence problem to work as a test for the well-balanced method and can now do second-order or fourth order initial model generation.

* Changes the meaning of the `plm_iorder` flag to now do 1 = piecewise constant, 2 = 2nd order MC limiter,  and >= 3 for 4th order MC limiter  

* Adds support for HSE boundary conditions in +X and +Y

## PR checklist


zingale:
this also addresses #658

zingale:
also addresses #294

zingale:
All tests pass

Source/driver/_cpp_parameters
```diff
@@ -104,7 +104,12 @@ ppm_predict_gammae           int           0                  y
 ppm_reference_eigenvectors   int           0                  y
 
 # for piecewise linear, reconstruction order to use
-plm_iorder                   int           2                  y
+# 1 = piecewise constant, 2 = MC limiter, >2 = 4th order MC limiter
+plm_iorder                   int           3                  y
```
maxpkatz: I don't believe we should change the default PLM order, this is a breaking change without a good reason for it.

zingale: I'm not changing it
I am just changing what the numbers mean.  So 3 is the new 2

zingale: the idea is 1 is first order (piecewise constant), 2 is second order slopes, >=3 is the 4th order limited slopes

maxpkatz: If someone has hardcoded plm_iorder in their inputs script, that inputs script will now do something different though, as a result of this change.

zingale: yes, I know.  I am open to suggestions.  It just seemed odd to me to have 1 = first order, 2 = 4th order, 3 = 2nd order...

zingale: okay, I've changed how this works.  `plm_iorder` remains the same as before now.  The choice of limiter when using linear slopes is now determined by `plm_limiter`.  I think this is actually a more natural way to do it, since both limiter choices are still linear slopes and therefore second-order.

Source/hydro/Castro_mol_hydro.cpp
```diff
@@ -481,7 +486,14 @@ Castro::construct_mol_hydro_source(Real time, Real dt, MultiFab& A_update)
 	// Store the fluxes from this advance -- we weight them by the
 	// integrator weight for this stage
 
-        for (int idir = 0; idir < AMREX_SPACEDIM; ++idir) {
+        // For SDC, we store node 0 the only time we enter here (the
+        // first iteration) and we store the other nodes only on the
+        // last iteration.
+        if (time_integration_method == MethodOfLines ||
```
maxpkatz: This MethodOfLines integration method is now obsolete

Source/hydro/Castro_mol_nd.F90
```diff
@@ -59,66 +60,238 @@ subroutine ca_mol_plm_reconstruct(lo, hi, &
...
+
+       ! special care for reflecting BCs
+       if (idir == 1) then
+          if (lo(1) == domlo(1)) then
+
+             do k = lo(3)-dg(3), hi(3)+dg(3)
+                do j = lo(2)-dg(2), hi(2)+dg(2)
+
+                   ! reset the left state at domlo(1) if needed -- it is outside the domain
+                   if (physbc_lo(1) == Symmetry) then
```
maxpkatz: Perhaps we should do these if checks outside the loop?

zingale: I fixed this

Issue #638, opened 2019-07-13
remove Strang + MOL (time_integration_method = 1)


zingale:
The new SDC method is merged and does proper coupling of hydro and reactions, making the Strang + MOL method not necessary.  If we want to do Strang, then CTU is better.  If we want to do a MOL-style integration, then the SDC integration, even 2nd order in time, should be better than the MOL.  Furthermore, the SDC method extends more easily to AMR + subcycling.

I propose we remove the pure MOL (RK time integration) method, since we are not likely to use it and don't want to support it.

PR #657, opened 2019-08-20
remove the RK-based MOL time integration method

zingale:
## PR summary

it is not tested, still uses Strang, and doesn't have an easy path to AMR
SDC is the better option.

addresses #638 

## PR checklist


Source/driver/Castro.cpp
```diff
@@ -454,13 +442,11 @@ Castro::read_params ()
       Radiation::read_static_params();
     }
 
-    // The CUDA MOL implementation doesn't currently do radiation.
-#ifdef AMREX_USE_CUDA
+    // radiation is only supported with CTU
     if (do_radiation && time_integration_method != CornerTransportUpwind) {
-        amrex::Error("Radiation is currently unsupported for MOL when using CUDA.");
+        amrex::Error("Radiation is currently unsupported for CTU time advancement.");
```
harpolea: I think this should say `Radiation is currently only supported for CTU time advancement`?

zingale: fixed

Source/driver/Castro_setup.cpp
```diff
@@ -1089,91 +1089,44 @@ Castro::variableSetUp ()
...
+    // Gauss-Lobatto (Simpsons)
     SDC_NODES = 3;
 
     dt_sdc.resize(SDC_NODES);
     dt_sdc = {0.0, 0.5, 1.0};
 
+    //node_weights.resize(SDC_NODES);
+    //node_weights = {1.0/6.0, 4.0/6.0, 1.0/6.0};
+
+  } else if (sdc_order == 20) {
```
maxpkatz: Can we remove this Radau code from this PR please, and defer it to #666?

zingale: done

PR #670, opened 2019-09-15
change how we store the reaction terms for SDC

zingale:
## PR summary

now we do an average over the timestep, instead of just the instantaneous
value at the last time node.

## PR checklist


zingale:
this still shows 4th order convergence in the reaction source terms in the plotfile

zingale:
this is not ready yet... SDC tests are failing

zingale:
okay, tests work as expected now.  The reaction source terms will differ in the tests, since that's the purpose of this PR, but all the other variables are unchanged

zingale:
I'm no longer convinced this is the approach we want to take

PR #673, opened 2019-09-23
automatically create the probin parameters and namelist

### LINKED ISSUE #234: switch problem parameters to a script
We should list the problem-specific parameters in a file like we do with the extern parameters (like `_parameters`), and then parse this at compile time to build the probin module and the namelist read mechanism called from `probinit()`.

This will be cleaner and easier to generalize to the GPUs.  We should switch all problems to this system.  This is similar to what Maestro currently does.  We can build from the infrastructure that exists for extern already.

### LINKED ISSUE #619: probin stuff is not stored in job_info
Currently, the problem-specific variables set in `probin` are not stored in `job_info`.  We should fix this.  The easiest way to do it would be to implement #234

zingale:
## PR summary

The automates the creation of the probdata variables, the namelist for reading them, setting them as managed for CUDA, and adds the ability to output the values to a file (like job_info)

This feature is opt-in.  You need to set `USE_PROB_PARAMS` in your `GNUmakefile` and then
define the problem parameters in a file `_prob_params` in the problem directory.

For problems that opt-in, the values of the problem parameters are now output into the `job_info` file.

closes #234
closed #619 

## PR checklist


Util/scripts/write_probdata.py
```diff
@@ -0,0 +1,344 @@
...
+
+max_step                            integer            1
+small_dt                            real               1.d-10
+xlo_boundary_type                   character          ""
+octant                              logical            .false.
+
+This specifies the runtime parameter name, datatype, and default
+value.
+
+An optional 4th column can be used to indicate the priority -- if,
```
harpolea: It says here the 4th column should be a number, but in the parameter files for Sedov and flame_wave it’s y/n?

zingale: oh right, that comment is wrong.  I'll fix that.  The 4th column indicates whether the variable should be in the namelist

PR #674, opened 2019-09-25
swtich tagging from integer(1) to c_int8_t

### LINKED ISSUE #625: Fix tagging interfaces
Weiqun points out that integer(1) is not safe as a Fortran conversion from int8_t in C because the Fortran compiler may not treat the (1) as a 1-byte integer. To be safer we should use kind=c_int8_t.

zingale:
## PR summary

This changes the type of the tagging stuff from `integer(1)` to `kind=c_int8_t`.  Closes #625 

## PR checklist


zingale:
* GPU test suite passes
* Compiles without issue on Summit

zingale:
this works, all tests pass

PR #682, opened 2019-10-13
Avoid an unnecessary EOS call in diffusion

maxpkatz:
## PR summary

`ca_fill_temp_cond` has an unnecessary EOS call -- we already call the EOS to get a consistent temperature when we call `clean_state`.

## PR checklist


zingale:
Some versions of the conductivity need the electron chemical potential and other quantities, so I think we need this EOS call here.

maxpkatz:
Fair enough. We can revisit this when we have more granularity over what parts of the EOS we want to evaluate.

PR #689, opened 2019-11-02
Inline pdivu

maxpkatz:
## PR summary

There is no need to store a fab for pdivu, it can just be calculated on demand on a per-zone basis in the update.

## PR checklist


zingale:
I want to run the suite on this before merging

PR #695, opened 2019-11-16
Use the subcycling-enabled retry CTU advance for simplified SDC

maxpkatz:
## PR summary

The retry mechanism can be used for the simplified SDC advance as well as the standard CTU advance.

## PR checklist


maxpkatz:
The test for this should be wdmerger_retry, but compiled with USE_SIMPLIFIED_SDC=TRUE and with castro.time_integration_method = 3.

maxpkatz:
Ready for review, in conjunction with #215 from Microphysics. New test cases are `inputs-collision.testsuite` in the Detonation problem, and `inputs_test_wdmerger_retry` in wdmerger's wdmerger_retry test directory. Both can be added to the test suite for both CTU integration mode and simplified SDC integration mode, with `castro.time_integration_method=3` added to the command line for the latter. The Detonation problem tests retries in the case of burn failures, while the wdmerger problem tests retries in the hydro.

maxpkatz:
I've made some changes to improve this and make it quite simple. Now we simply do the SDC iterations inside the outer while loop that does retries.

I've also made some cosmetic changes to the return values of the CTU advance and react_state to return bools, indicating whether or not they were successful (unsuccessful meaning we either abort or retry).

PR #699, opened 2019-11-26
reduce memory usage for conserved-state sources

### LINKED ISSUE #698: reduce size of Source_Type
The StateData `Source_Type` has a size of `NUM_STATE`, but if the species don't have sources, then we don't need to include them in this memory.  We should have a conserved variable source size that we can use in the same fashion as `NQSRC` to reduce the size of this array.

zingale:
## PR summary

This change removes the species from the conserved state sources.  Closes #698 

## PR checklist

### DIFF
--- a/Source/driver/Castro.H
+++ b/Source/driver/Castro.H
@@ -1657,6 +1657,7 @@ public:
 /// variable counts
 ///
     static int       NQSRC;
+    static int       NSRC;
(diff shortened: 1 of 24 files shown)


zingale:
all CPU tests pass

zingale:
GPU tests pass

Issue #697, opened 2019-11-25
split species tracing off from the main PPM tracing


zingale:
We can greatly reduce the amount of memory needed in the PPM tracing routines if we don't compute the parabolic representations and integrals of all the species at once in the main tracing routine.  Instead we do the tracing separately.  This was done in mini-Castro here:

https://github.com/AMReX-Astro/mini-Castro/commit/d45865a01172b2e504441822237c35625c4e417c

If we go this route, we'd need to get rid of the ppm_temp_fix option where we work in the temperature eigensystem, since we will not have the species on edges to call the EOS.

zingale:
One complication here is that we need to define a `QTHERM` which is the primitive state minus and species and other passive quantities.  But radiation gets in the way.  Look at `_variables`.  We need both `NQSRC` and `NQTHERM` to be contiguous.  But the speices might add to `NQSRC` and not `NQTHERM` while the radiation might do the opposite: add to `NQTHERM` and not `NQSRC`

zingale:
This work is being done on the `species_separate` branch

PR #700, opened 2019-11-27
Trace species separately

zingale:
## PR summary

This change syncs us with mini-Castro by doing the tracing of the species separately from the
main thermodynamic tracing.

## PR checklist


zingale:
This addresses #697

zingale:
as part of this change, the variable parsing script allows for multiple "adds to" variables.

zingale:
To address the concern that radiation and species both add to different variables and we need things contiguous, the next step is to split the species primitives into its own array.

This will be a big change, and affect all of the solvers, but it will give us a lot of flexibility down the road.

zingale:
going to revisit this once the C++ conversion is done.

PR #709, opened 2019-12-13
Add an option for implicit external source terms

maxpkatz:
## PR summary

The external source terms are currently only implemented as an explicit predictor-corrector scheme. This is incompatible with how wdmerger implements its relaxation source term, so wdmerger is currently not working as intended. This PR introduces a new option, castro.ext_src_implicit (defaulting to 0) which, if turned on, changes the handling of the external source terms to allow an implicit solve. This is done by subtracting the full old-time source and adding the full new-time source in the corrector, rather than -0.5 and +0.5 of each, respectively. It would still be up to the individual problem to make sure they are consistent with this scheme if they turn the option on. Even if they don't, the source term ends up being a simple time-advanced source term at n+1 which is not a serious problem.

## PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,14 @@
+# 20.01
+
+   * A new option castro.ext_src_implicit has been added. The external source
+     terms were previously only implemented as an explicit predictor-corrector
+     scheme. The new option, if turned on, changes the handling of the external
+     source terms to allow an implicit solve. This is done by subtracting the
+     full old-time source and adding the full new-time source in the corrector,
+     rather than -0.5 and +0.5 of each, respectively. It is still up to the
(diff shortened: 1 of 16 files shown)


zingale:
all tests pass, but the PR suggested that answers should change, so I will not merge until we understand this.

zingale:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2019-12-12-003/index.html

maxpkatz:
I incorrectly believed that at least one of the wdmerger problems was using the external source terms, but this was not the case. I have now made this change on wdmerger_3D in development. Please run that test on development, reset it, and then re-run the test on this branch.

PR #710, opened 2019-12-13
Add option to apply sources consecutively

maxpkatz:
## PR summary

This new option allows you to add sources to the state as they are calculated, rather than grouping them together at the end. Its purpose is to allow sources that come at the end, in particular the sponge and the external source term, to act on the state after the actual physics source terms have already been applied.

## PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 # 20.01
 
+   * A new option castro.apply_sources_consecutively has been added. By default
+     we add all source terms together at once. This option, if enabled, adds the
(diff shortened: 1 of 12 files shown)


zingale:
this breaks SDC.  We need an interface where we can just get the sources without applying them.

maxpkatz:
The SDC-related failures are here.

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2019-12-14/index.html

I have updated the interfaces and disabled applying the sources for SDC.

zingale:
SDC tests pass now

PR #719, opened 2019-12-29
Remove self-gravity define, use GRAVITY always

### LINKED ISSUE #157: SELF_GRAVITY ifdef is inconsistent and confusing
currently, `SELF_GRAVITY` is separate from `GRAVITY` in delineating code sections, but things are not always consistent.  We need to clean this up.

maxpkatz:
## PR summary

This PR closes #157 by removing the SELF_GRAVITY ifdef -- we revert back to the previous behavior where GRAVITY is the only relevant define.

Also, I'm including a centralization on the approach for grav_source_type = 4 based on g, not phi. The mechanism based on phi is slightly better at conservation in the presence of coarse-fine boundaries, but the difference between the two methods is quite small in absolute terms -- both are very good -- and it is better to only have one method so that we don't have to worry when switching between different gravity implementations (not all of them define phi).

## PR checklist

### DIFF
--- a/Source/gravity/gravity_sources_nd.F90
+++ b/Source/gravity/gravity_sources_nd.F90
@@ -9,10 +9,8 @@ contains
   subroutine ca_gsrc(lo, hi, &
                      domlo, domhi, &
                      uold, uold_lo, uold_hi, &
-#ifdef SELF_GRAVITY
                      phi, phi_lo, phi_hi, &
                      grav, grav_lo, grav_hi, &
-#endif
                      source, src_lo, src_hi, &
                      dx, dt, time) bind(C, name="ca_gsrc")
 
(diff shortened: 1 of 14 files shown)


zingale:
test suite output is here:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2019-12-28/index.html

some diffs are big (like evrard_collapse)

maxpkatz:
evrard_collapse and wdmerger_3D are both sensitive to the energy term due to their large kinetic energies on the grid, so it is not surprising that small differences in how it is applied have noticeable effects. I don't think either case is more right than the other, they're both accurate to second order, so I'm OK with that diff.

Source/gravity/gravity_sources_nd.F90
```diff
@@ -77,12 +67,6 @@ subroutine ca_gsrc(lo, hi, &
     src(:) = ZERO
     snew(:) = ZERO
 
-    ! For constant gravity, we can just initialize the gravitational acceleration here.
-
-#ifndef SELF_GRAVITY
-    Sr(dim) = const_grav
```
zingale: why is this no longer needed?

Source/gravity/gravity_sources_nd.F90
```diff
@@ -104,11 +88,7 @@ subroutine ca_gsrc(lo, hi, &
 
              old_ke = HALF * sum(snew(UMX:UMZ)**2) * rhoInv
 
-#ifdef SELF_GRAVITY
              Sr = rho * grav(i,j,k,:)
-#else
-             Sr(dim) = rho * const_grav
```
zingale: likewise, why is this no longer needed?

Source/gravity/gravity_sources_nd.F90
maxpkatz: It was never needed. We fill the StateData for g when we're using ConstantGrav. That code was put in there so the version without SELF_GRAVITY would compile, but it was not necessary for correct results.

PR #720, opened 2019-12-29
Set compute_new_dt_on_regrid by default

### LINKED ISSUE #356: Force new dt after regrid

maxpkatz:
## PR summary

The AMR parameter `amr.compute_new_dt_on_regrid` is now on by default; closes #356. This avoids crashes that result from the CFL number being too large after regridding, because we update the timestep after seeing that larger velocity.

The user can still opt to set this off if they want to, since the location where we're setting it in Castro occurs before the Amr ParmParse of this parameter.

## PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # 20.01
 
+   * The AMR parameter amr.compute_new_dt_on_regrid is now on by
+     default. This avoids crashes that result from the CFL number
+     being too large after regridding, because we update the
+     timestep after seeing that larger velocity. You can still opt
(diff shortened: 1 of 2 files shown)


zingale:
tests are here:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2019-12-29/index.html

PR #722, opened 2020-01-02
Replace do_real_eos = 0 with an actual EOS

maxpkatz:
## PR summary

This PR adds an EOS, `rad_power_law`, which replaces the functionality previously provided by `radiation.do_real_eos = 0`. The EOS defines `c_v` as before, and can only be called in (rho, T) and (rho, e) modes. It does not define hydrodynamic variables like pressure, so it cannot be used in conjunction with `castro.do_hydro = 1`. This sets us up for merging the radiation and non-radiation EOS interfaces.

## PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # 20.02
 
+   * The parameter radiation.do_real_eos = 0 has been removed, and its
+     functionality is now enabled with a new equation of state called
+     rad_power_law. This new EOS is only compatible with the pure
+     radiation-diffusion tests, not with castro.do_hydro = 1. (#722)
+
    * Support for neutrino radiation diffusion has been removed.
 
 # 20.01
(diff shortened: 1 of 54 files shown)


zingale:
tests are here:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-01-02-003/index.html

note that RadSuOlsen crashes with negative rho e

zingale:
answers change a lot in the current RadSuOlson test

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-01-02-004/rad-SuOlson.html

maxpkatz:
I looked at the old code. We were previously setting T and e to 0 in ca_initdata. What was happening in the computeTemp call after ca_initdata is that the old EOS's approach of flooring the data by adding 1e-50 meant that it was treating rho*e as 1.e-50, and then calculating T from there. So the original data was completely arbitrarily determined by that floor.

Since the initial values don't really matter, I've set these to be the same as they effectively were before (1.e-50) so that results don't change much.

PR #725, opened 2020-01-03
Move power law opacity into its own module

maxpkatz:
## PR summary

This PR unifies the approach to opacities in the radiation code, so that ca_compute_planck, ca_compute_rosseland, and ca_compute_scattering are the only entry points to getting the opacities. These work for single-group or multi-group problems. To do this, we moved the power-law opacity to its own module in Castro/Microphysics, `rad_power_law`, replacing the `powerlaw-example` case, and now have the problems point to the new opacity directory rather than `null`. (The `null` directory has been removed, to prevent people from accidentally trying to continue using the old setup.) All of the corresponding inputs parameters for the power law opacity have been moved to the probin file (in the extern namelist).

We now always expect the user to pick a specific opacity implementation, so the parameter radiation.use_opacity_table_module has been removed.

The multigroup solver was previously forcing the Rosseland opacity to be equal to the Planck opacity if the latter was set but the former was not. There was similar unintuitive behavior for the behavior of the scattering parameter. Now you will get exactly what you ask for in the probin file, given the defaults in the _parameters file for the rad_power_law opacity. By default the constant coefficients for both are negative, which is invalid, so both must be set to a non-negative value for the code to work. Problems that were previously setting const_kappa_p but not const_kappa_r should set the latter equal to the former to maintain the same code behavior. The analogous thing should be done for the exponents (kappa_p_exp_m, kappa_p_exp_n, and kappa_p_exp_p).

## PR checklist


Exec/radiation_tests/RadSourceTest/inputs-cooling
```diff
@@ -139,51 +139,6 @@ radiation.hi_bcflag = 0 0 0
...
-# to be meaningful.  It does not affect conservation.
-
-radiation.temp_floor    = -1.e-20
-
-# As kappa_r becomes low the Marshak boundary condition becomes
-# ill-conditioned.  The following floor is enforced regardless of the
-# mechanism used to set kappa_r.  (In an ideal code this floor would
-# be nondimensionalized using a characteristic length (domain size?).
-# In the present implementation the Marshak b.c. includes a factor of
-# delta x which can cause ill-conditioning if delta x is not O(1).)
```
zingale: There's some useful info in these comments that is getting deleted (not just this problem).  Perhaps we can move them to a README?

Exec/radiation_tests/RadSuOlsonMG/probin.common
```diff
@@ -16,4 +16,14 @@
   eos_c_v_exp_m = 0.0d0
   eos_c_v_exp_n = -3.0d0
 
+  const_kappa_r = 1.9801980198019803d-31
```
zingale: this was not set originally

maxpkatz: That's right, but I'm preserving the previous behavior, which was that the code was forcing kappa_r to be equal to kappa_r if kappa_p was set but kappa_r was not.

Issue #729, opened 2020-01-09
optimized CUDA run crashes on Summit / debug version works


zingale:
Running on Summit using `rotating_star` compiled with:
```
make COMP=pgi USE_CUDA=TRUE USE_SIMPLIFIED_SDC=TRUE
```
using `inputs_2d` and `probin` from that directory, I can run fine if I compile with `DEBUG=TRUE`, but if I run without debug, I get `too many subcycles` from already in step 0.

So it seems that there is either a bug in our code or in PGI.  

Note: this is with PGI 19.10

zingale:
Note: I made one change to the `inputs_2d`:
```
castro.init_shrink    = 0.001
```
and one change to the `probin`:
```
jacobian = 2
```

zingale:
I'll try PGI 19.4

zingale:
okay... oddly, with 19.4 it runs, and also takes far fewer timesteps.  So there is definitely something odd going on...

zingale:
actually, it is not _far_ fewer.

With 19.10 in debug mode I get:
```
STEP = 169 TIME = 0.001914485266 DT = 0.0001707639119
```

With 19.4 in non-debug mode:
```
STEP = 169 TIME = 0.002840796515 DT = 0.0002411334111
```

maxpkatz:
This is caused by a bug introduced by #695. The simplified SDC burn only fills the "old" Reactions_Type data (I do not remember why it is the old data and not the new data), but the check_timestep routine uses both the "old" and "new" data for computing whether we violated dtnuc_e or dtnuc_X, since it was written for CTU + Strang splitting. So we are computing on uninitialized data.

zingale:
Note: the issue is in both `ca_check_timestep` and `ca_estdt_burning`.  The easiest solution is to simply define the "new" burn state to be the same as the old for SDC

PR #733, opened 2020-01-17
fix nuclear burning dt estimator with simplified SDC

zingale:
## PR summary

The simplified-SDC solver didn't fill the new time reactions, so the ca_check_timestep routine was accessing invalid data when computing the nuclear burning timestep.

## PR checklist


zingale:
this fixes #729

PR #734, opened 2020-01-18
Compute geometry factors on the fly in CTU hydro

maxpkatz:
## PR summary

For the places in the CTU hydro where we need area or volume (or the geometry source term factor) we compute it on the fly. This slightly reduces memory pressure.

## PR checklist


Source/driver/Castro_util.F90
```diff
@@ -679,6 +680,64 @@ end function area
...
+
+    dlogarear = ZERO
+
+    if (coord_type .eq. 1 .and. dir .eq. 1) then
+
+       ! Cylindrical (2D only)
+
+       ! Get edge-centered position
+
+       loc = position(i,j,k,cc(1),cc(2),cc(3))
```
zingale: don't you need to do `cc(idir) = .false.` here?

maxpkatz: I would have thought so too, but I looked at the dLogArea calculation in AMReX and it seems to be using cell-centered indexing, which I think is consistent with how we're computing the geometry source terms.

maxpkatz: I'll fix the comment to be consistent.

zingale:
here's the test output for this PR:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-01-21-003/index.html

zingale:
this seems to really slow down the CPU code

maxpkatz:
I have two hypotheses for that: (1) we're getting hurt because these functions are not inlined; (2) the position/area/volume functions have a bunch of conditional checks in them, so they're not especially optimized code. The first one could be tested using mega_fortran -- is it possible to do a test suite run with that?

maxpkatz:
Closing. Aside from the issue noted above, the primary motivation for this was to minimize the number of Fabs being passed into the kernels to deal with the register pressure issue. But we are addressing this indirectly by our long-term transition to C++.

PR #738, opened 2020-01-19
Fourth cleanup 3

zingale:
## PR summary

The merges the calls to normalize_species_fluxes between the 2nd and 4th order code paths + adds the density flux limiter to the 4th order (even though it won't be 4th order).  Finally, it reorganizes the construction of the fluxes so that each direction's code is together.  This will make it easy to move this to C++ as a single loop over dimension.

## PR checklist


zingale:
tests pass.

PR #747, opened 2020-01-26
switch state indices to parameters / const int

zingale:
## PR summary

This switches the variable keys to Fortran paramerters (in state_indices.F90) and C++ const ints (in state_indices.H).  This change will help with the GPUing and lambda captures.  A lot of the routines that were used to pass these indices back and forth between Fortran and C++ are now gone.  

Note: this change requires the network to have a simple C++ header file giving the network and aux sizes.

## PR checklist


zingale:
depends on #746

zingale:
all CPU tests pass

GPU tests have roundoff diffs (must be optimization?):
http://groot.astro.sunysb.edu/Castro/test-suite/gpu/

zingale:
this is replaced by #754 and #755

Issue #741, opened 2020-01-20
2nd order MOL can be memory optimized


zingale:
We currently precompute the x, y, and z interface states at once and then have a loop over direction to do the Riemann, etc.  We should dimension qm, qp to be just single direction and fill the states in the loop over direction.  This will save a lot of memory.

I think some of the boxes are bigger than needed as well.

PR #749, opened 2020-01-27
2nd order MOL/SDC cleaning

zingale:
## PR summary

This reduced the memory needed for the interface states by doing all of the state reconstruction / flux solve in a loop over dimensions.  Addresses #741 

## PR checklist


zingale:
All tests now pass (there are 3 in the test suite that do 2nd order SDC)

Source/hydro/Castro_mol_hydro.cpp
```diff
@@ -439,62 +439,60 @@ Castro::construct_mol_hydro_source(Real time, Real dt, MultiFab& A_update)
...
-                 AMREX_REAL_ANYD(dx));
-            }
+            if (do_hydro) {
 
-          }
+              if (ppm_type == 0) {
 
-          // compute the fluxes and add artificial viscosity
+                dq.resize(obx, NQ);
+                Elixir elix_dq = dq.elixir();
```
maxpkatz: This won't work, but we can resolve that when we do the final pass for GPUs.

zingale: okay, but can you briefly share why it doesn't work?

PR #757, opened 2020-02-08
sync the conserved state names between C++ and Fortran

zingale:
## PR checklist


zingale:
This now sets the only state names for backwards compatibility.

Source/driver/Castro.cpp
```diff
@@ -3281,7 +3281,7 @@ Castro::computeTemp(MultiFab& State, Real time, int ng)
 
       ca_compute_lap_term(BL_TO_FORTRAN_BOX(bx0),
                           BL_TO_FORTRAN_FAB(Stemp[mfi]),
-                          BL_TO_FORTRAN_ANYD(Eint_lap[mfi]), &Eint,
+                          BL_TO_FORTRAN_ANYD(Eint_lap[mfi]), UEINT,
```
maxpkatz: I think you need to keep the reference here, to make the Fortran interface work. Same for the next change.

PR #758, opened 2020-02-08
Handle the plotfile case where a derived variable has multiple components

maxpkatz:
## PR summary

In this case, the plotfile logic would only plot the first component of the derived
variable. Now, if a derived variable has multiple components, we plot all components.
The name of each is appended with an integer variable indicating which one it is.

This issue also exists in AMReX, where this plotfile code presumably originally came from.

## PR checklist


maxpkatz:
The most common case affected by this will be the StateErr derived variable, which is part of the default set of Castro derives. Now each plotfile will have "StateErr_0", "StateErr_1", and "StateErr_2", so the regression suite will need to be updated.

zingale:
the test suite never completes with this PR.  The Detonation test has been running for ~12 hours.  Going to kill it now.

maxpkatz:
Fixed a bug, I was accidentally trying to plot too many variables.

PR #759, opened 2020-02-09
Add analytic solution for the radiating sphere to the plotfile

maxpkatz:
## PR summary

The Python analysis routine is also updated to take advantage of this, rather than relying on a fixed output (which could only be used at a specific simulation time and distance from the origin).

## PR checklist


zingale:
the analysis script fails -- there is no -v option

zingale:
looks like I needed a `make realclean` in the diagnostics dir

Issue #651, opened 2019-08-11
Make AMReX and Microphysics submodules of Castro


maxpkatz:
AMReX and Microphysics should be git submodules of Castro for a lower barrier to entry for new users.

I've sketched out what this looks like in mini-Castro already, for AMReX. We would create directories called amrex and Microphysics in the top-level Castro directory. Then inside those we set a specific commit of each of those codes to check out when you download Castro (which would presumably be the most recent monthly release of each of those codes -- this enforcement of version parity is a side benefit of going this route). A new user getting the code for the first time just has to do
```
git clone --recursive
```
and a user that already has the Castro code just has to do 
```
git submodule init ---update
```
to get the submodules downloaded.

This will not break the workflow of those of us who maintain those dependencies separately, since in the Makefile we will do
```
AMREX_HOME ?= $(TOP)/amrex
```
and if you already have `AMREX_HOME` in your environment it will pick up your other copy of it.

We discussed this a few years ago and decided not to do this, but we should revisit this because my proposed mechanism gives a person the freedom to use the code either way. Also, since it is now very common for large HPC codes to have git submodule dependencies like this now (e.g. most Kokkos- and RAJA-based codes use Kokkos and RAJA as git submodules), I don't think this will be a confusing workflow for users.

A major benefit of this approach for Microphysics is that we would not have to maintain redundant copies of the Microphysics interfaces (and the gamma law EOS) in Castro's Microphysics directory. We could just have Microphysics supply that directly. I recall an argument that was made at the time was that we don't want users to be confused about extra requirements, but aside from my point above that I think most users are familiar with this idea by now, we can also have this be friendly by detecting if either MICROPHYSICS_HOME is set or if the Microphysics directory doesn't yet exist in the top-level directory (because they forgot to do git clone --recursive) and then printing out a helpful error message telling them what to do in that case. I've done this in mini-Castro for AMReX, as follows:
```
# Include the AMReX make rules. Throw an error
# if we don't have AMReX.
ifeq ("$(wildcard $(AMREX_HOME)/Tools/GNUMake/Make.defs)","")
  $(error AMReX has not been downloaded. Please run "git submodule update --init" from the top level of the code)
endif
include $(AMREX_HOME)/Tools/GNUMake/Make.defs
```

adam-m-jcbs:
tl;dr: 
1) Excellent use-case for submodules
[...]

Issue #690, opened 2019-11-03
remove Microphysics directory


zingale:
We should just require the Microphysics repo instead of locally reproducing some of the functionality in Castro.  This would make it easier to manage the EOS and burner interfaces.

PR #760, opened 2020-02-09
Replace Castro microphysics with starkiller-astro Microphysics

maxpkatz:
## PR summary

This implements the Microphysics part of #651. We replace the local Castro Microphysics directory with a dependence on the main starkiller-astro Microphysics directory. This means that obtaining Microphysics is now a requirement for using Castro. If the user sets `MICROPHYSICS_HOME` as before, then their workflow does not change. A new path is that Microphysics is a git submodule of Castro, so if the user does `git submodule update --init --recursive` then they will obtain the version of Microphysics that we specify in the repo, and if they do not have `MICROPHYSICS_HOME` set then this submodule will be used. The repo is now updated with `git pull --recurse-submodules`, so that updates to Microphysics come along when you pull.

For Castro developers, this means that -- at a minimum -- we must update the Microphysics submodule monthly when we issue new releases. The release for Microphysics must be done first. Then navigate to the Microphysics directory, checkout the new tag, and then from the top-level directory of Castro do a `git add` on the Microphysics directory to store the new tag. So, for example, at the beginning of March 2020 we would first issue the `20.03` tag on Microphysics, then do

```
cd $CASTRO_HOME/Microphysics
git pull
git checkout 20.03
cd ..
git add Microphysics
git commit -m "Update Microphysics to release 20.03"
```

Then we can proceed with issuing our own release.

When breaking changes to Microphysics occur in its `development` branch that Castro depends on, we must update the Microphysics submodule on the Castro `development` branch in the same way, replacing the git checkout statement with the latest commit hash on the Microphysics `development` branch. A git submodule always tracks a specific commit/tag on the target repo -- it is not configured to automatically track a particular branch.

In order to ensure that we don't lose track of this coupling between the two codebases, it is my recommendation that we use this git submodule for the Castro regression test suite, rather than relying on a standalone installation of Microphysics. So we should remove any references to MICROPHYSICS_HOME as part of that setup, and then make sure the git commands are appropriately updated.

A minor associated change is that we've standardized on `NETWORK_INPUTS` in all of the makefiles, instead of using `GENERAL_NET_INPUTS` with the hardcoded path.

## PR checklist


zingale:
note: this also implements #690

zingale:
all tests pass here, using the current external Microphysics.  

I'm going to merge and then change up the test suite shortly.  I think we want to test both ways -- using submodules (which is how users will probably use the code, but likely on `master`) and using the separate Microphysics (which is how developers will likely use the code).

PR #767, opened 2020-02-12
move the EOS reset on states outside of the trans loop

zingale:
## PR checklist


zingale:
all CPU tests pass

zingale:
GPU tests look good -- they are here: 
http://groot.astro.sunysb.edu/Castro/test-suite/gpu/2020-02-12-008/index.html

maxpkatz:
Noting that the reason we chose to do this is because it removes the EOS from the transverse routines, which allows us to move the transverse routines to C++ independently of the implementation of the EOS in Microphysics in C++ (which is happening over a longer timescale). However, in the long run when we have got that all done, we will want to look at moving this back to be a zone-by-zone implementation.

PR #768, opened 2020-02-13
merge the transverse routines

zingale:
## PR summary

This consolidates the various transverse routines, using mini-Castro as a guide.
 

## PR checklist


zingale:
This is ready for review.  Some notes:

* 2-d axisymmetric pure hydro will have roundoff diffs because we now include the area factors in the z-direction (previously it was only in the r direction)
* we are slower by about 10% -- this is likely because the compiler cannot do cache optimization.  We will address this with templating when we convert to C++
* for radiation, in 2-d we now use the general geometry form for the radiation pressure in the transverse flux

zingale:
Here are the GPU test results:
http://groot.astro.sunysb.edu/Castro/test-suite/gpu/2020-02-18-001/index.html

zingale:
this is the CPU tests:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-02-18-005/index.html

PR #775, opened 2020-02-17
Extrapolate borders

maxpkatz:
## PR summary

This gets extrapolateBorders dimension agnostic and GPU friendly. The previous approach was not zone-local, so I've rewritten the algorithm such that ghost zones don't depend on other ghost zones.

## PR checklist

### DIFF
--- a/Source/radiation/Rad_nd.F90
+++ b/Source/radiation/Rad_nd.F90
@@ -9,6 +9,188 @@ module rad_nd_module
 
 contains
 
+  subroutine bextrp(lo, hi, &
+                    vlo, vhi, &
+                    f, f_lo, f_hi) &
+                    bind(C, name="bextrp")
+
+    integer,  intent(in   ) :: lo(3), hi(3)
+    integer,  intent(in   ) :: vlo(3), vhi(3)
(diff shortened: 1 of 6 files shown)


maxpkatz:
@zingale please let me know if you see a cleaner (less verbose) way to write this.

Source/radiation/Rad_nd.F90
```diff
@@ -9,6 +9,188 @@ module rad_nd_module
...
+    do k = lo(3), hi(3)
+       do j = lo(2), hi(2)
+          do i = lo(1), hi(1)
+
+             ! Extrapolate the data in the ghost zones of the box using a one-sided difference:
+             ! f(i-1) = 2 * f(i) - f(i+1)
+             !
+             ! For corners in 2D, or edges in 3D, this is applied in both directions, yielding
+             ! f_{i-1,j-1} = 4 * f(i,j) - 2 * (f(i+1,j) + f(i,j+1)) + f(i+1,j+1)
+             !
```
zingale, line 35: I don't think that this is conservative.  The easy way to see this is imagine that f is constant, then this fills the corner with a value 3 f, instead of f.

Usually for a conservative interpolant, the weights sum to 1.

PR #780, opened 2020-02-19
remove the reconstruction of gamma_e on interfaces

zingale:
## PR summary

This drops the alternative to (rho e) on interfaces -- reconstructing gamma_e.  They both are about the same, so there's no need to support both

## PR checklist


zingale:
Note: as a consequence of this, we will be removing the ppm_temp_fix = 1 option from CTU (since apparently it was never coded into the default rhoe reconstruction for some reason).  See issue #781

zingale:
This depends on #768

zingale:
all tests, aside from the 2 that explicitly used gammae tracing, pass.

PR #782, opened 2020-02-20
reorganize the makefiles

zingale:
## PR summary

This splits up the makefiles. One https://github.com/starkiller-astro/Microphysics/pull/261 is merged,
we can get rid of `Make.Microphysics`

## PR checklist

- [ ] test suite needs to be run on this PR
- [ ] this PR will change answers in the test suite
- [ ] all functions have docstrings as per the coding conventions
- [ ] the `CHANGES` file has been updated
- [ ] if appropriate, this change is described in the docs

Issue #258, opened 2017-11-21
create clang test suite


zingale:
We should test with clang + gfortran

compile with `COMP=llvm`

PR #784, opened 2020-02-21
get the aux_names from the C++ header

zingale:
## PR checklist


maxpkatz:
Since this depends on recent changes to Microphysics, please include an update to the Microphysics submodule to at least the commit corresponding to starkiller-astro/Microphysics#258. If we merge #782 first, this will be unnecessary.

zingale:
already merged #782 with a submodule update, so I think we're good.

PR #785, opened 2020-02-21
Rename Fortran files to end with _nd.F90

maxpkatz:
## PR summary

As we convert things to C++, we will have intermediate states where some parts of a file are in Fortran and others in C++. To make this easier, we should maintain the convention that the C++ file has the normal name (e.g. advection_util.cpp) and the Fortran file has a _nd suffix (e.g. advection_util_nd.F90). This PR converts several files to abide by this convention to make the porting easier later.

## PR checklist


zingale:
All tests pass

PR #786, opened 2020-02-22
convert PPM reconstruction to C++

zingale:
## PR checklist


zingale:
here are the CPU tests:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-02-25-002/index.html

Source/hydro/ppm_nd.F90
```diff
@@ -7,7 +7,7 @@ module ppm_module
 
 contains
 
-  subroutine ppm_reconstruct(s, flatn, sm, sp)
+  subroutine ppm_reconstruct_f90(s, flatn, sm, sp)
```
maxpkatz: We don't need to change the names if they're not bind(C). The mangled name will not conflict with the C++ version.

Source/hydro/ppm.cpp
```diff
@@ -0,0 +1,157 @@
...
+    dsvl_r = std::copysign(1.0_rt, dsc) * std::min(std::abs(dsc),std::min(std::abs(dsl),std::abs(dsr)));
+  }
+
+  // Interpolate s to edges
+
+  sm = 0.5_rt * (s[i0] + s[im1]) - (1.0_rt/6.0_rt) * (dsvl_r - dsvl_l);
+
+  // Make sure sedge lies in between adjacent cell-centered values
+
+  sm = std::max(sm, std::min(s[i0], s[im1]));
```
maxpkatz: On the GPU we need to use amrex::min and amrex::max instead of std::min and std::max.

Source/hydro/trace_ppm.cpp
```diff
@@ -0,0 +1,561 @@
...
+    Real Ip_src[NQSRC][3];
+    Real Im_src[NQSRC][3];
+
+    for (int n = 0; n < NQSRC; n++) {
+
+      // do we even need to trace (non-zero source?)
+#ifndef AMREX_USE_CUDA
+      int do_trace = do_source_trace[n];
+#else
+      int do_trace = 0;
```
maxpkatz: This logic should be reversed to trace by default (safety outweighs performance).

Source/hydro/trace_ppm.cpp
```diff
@@ -0,0 +1,561 @@
...
+
+        // transverse velocities
+        qm(i,j+1,k,QUT) = Ip[QUT][1] + hdt*Ip_src[QUT][1];
+        qm(i,j+1,k,QUTT) = Ip[QUTT][1] + hdt*Ip_src[QUTT][1];
+
+      } else if (idir == 2) {
+        qm(i,j,k+1,QRHO) = std::max(lsmall_dens, rho_ref +  alphap + alpham + alpha0r);
+        qm(i,j,k+1,QUN) = un_ref + (alphap - alpham)*cc_ref*rho_ref_inv;
+        qm(i,j,k+1,QREINT) = rhoe_g_ref + (alphap + alpham)*h_g_ref + alpha0e_g;
+        qm(i,j,k+1,QPRES) = max(lsmall_pres, p_ref + (alphap + alpham)*csq_ref);
```
maxpkatz: amrex::max

Source/hydro/ppm.cpp
```diff
@@ -0,0 +1,157 @@
...
+#endif
+
+#include <cmath>
+#include <ppm.H>
+
+using namespace amrex;
+
+AMREX_GPU_HOST_DEVICE
+void
+Castro::ppm_reconstruct(const Real* s,
```
maxpkatz: For optimal GPU performance, we want to be able to inline the PPM reconstruction and integration. So it would be better if these were implemented as `inline` functions in `ppm.H`.

Source/hydro/ppm.cpp
zingale: fixed

Source/hydro/trace_ppm.cpp
zingale: fixed

Source/hydro/ppm_nd.F90
zingale: I thought it was helpful to us to remember.  These will go away shortly.

Source/hydro/ppm.cpp
zingale: done

zingale:
I think this is happy now.  There was a bug in the source tracing logic that is now fixed.  

GPU tests: http://groot.astro.sunysb.edu/Castro/test-suite/gpu/2020-02-27-001/index.html
CPU tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-02-27-001/index.html

Source/hydro/Make.package
```diff
@@ -18,12 +19,13 @@ endif
 
 ca_F90EXE_sources += advection_util_nd.F90
 
-ca_F90EXE_sources += ppm_nd.F90
+F90EXE_sources += ppm_nd.F90
```
maxpkatz: I don't think it matters, but was this intentional that you changed which make variable it gets added to?

zingale: no, that was a mistake (and breaks megafortran).  I'll fix it.

Source/hydro/trace_ppm.cpp
```diff
@@ -0,0 +1,561 @@
...
+        for (int b = k-2; b <= k+2; b++) {
+          if (std::abs(srcQ(i,j,b,n)) > 0.0_rt) {
+            do_trace = 1;
+            break;
+          }
+        }
+      }
+#endif
+
+      //do_trace = 1;
```
maxpkatz: Can you remove this debugging statement please?

zingale: done

Issue #790, opened 2020-02-26
reorganize the Riemann stuff / convert to C++


zingale:
There are currently 3 main entry points to the Riemann solvers, to support the variety of discretization methods:

1. `cmpflx_plus_godunov`:
    * calls `cmpflx` over `lo:hi`
    * call `ca_store_godunov_state` over `lo:hi`

    * on exit, returns `flux(NVAR)`, `qint(NQ)` and `qgdnv(NGDNV)`

2. `cmpflx`:
    * calls `riemann_state` on `lo:hi`
    * calls `compute_flux_q` on `lo:hi`

    * on exit, returns `flux(NVAR)`, `qint(NQ)`  

3. `riemann_state`:
    * calls `riemannus` or `riemanncg` on `lo:hi`
    * on exit, returns `qint(NQ)`

currently the solvers need the following:

* CTU: `cmpflx_plus_godunov`
* MOL 2nd order: `cmpflx_plus_godunov`
* MOL 4th order: `riemann_state`, `compute_flux_q`, and `ca_store_godunov_state`

zingale:
Some notes:

1. nothing uses the `cmpflx` interface directly from outside of Riemannland
2. the current `cmpflx_plus_godunov` wastes memory, since we come in with `qint(NQ)` over the entire nodal box just to store the intermediate state before calling `ca_store_godunov_state`

zingale:
There are a few ways to recover this memory.

1. We convert the core Riemann solvers (`riemannus` and `riemanncg`) to work on a zone by zone basis instead of over an entire box.  Then we can do the conversion to the Godunov state just using the single zone temporary `qint(NQ)`
2. We use templating, and have a function that can output the state either over NQ or NGDNV while still operating over the whole box.

For option 1, we can make those implementations inline and header-only to avoid the function call overhead and allow vectorization.

zingale:
Note the C++-ing will require the C++ EOS support to be merged into Castro.

zingale:
this has largely been done.

PR #793, opened 2020-02-26
split out the Riemann solvers and make them zone-by-zone

zingale:
## PR checklist


zingale:
this is part of #790

zingale:
when we switch to C++, everything in riemann_solvers_nd.F90 and riemann_util_nd.F90 will be inline header-only.

zingale:
although this works, and seem better on the GPU, we are doing the C++ conversion first

PR #795, opened 2020-02-27
Implement flux limiter based on small eint

maxpkatz:
## PR summary

A new option, castro.limit_fluxes_on_small_eint, is added that limits fluxes if they would generate small/negative internal energies.

## PR checklist


maxpkatz:
Closing; will not implement. If we want to avoid small/negative eint we should do it with retries like our new strategy in #796 for density.

PR #796, opened 2020-02-27
Always retry if the density is small or negative

maxpkatz:
## PR summary

The current code only initiates a retry if the density was updated by enforce_min_density from a sufficiently negative number to a sufficiently positive one. This made sense at the time of its implementation but a simpler method is now available: we check whether the minimum value of the density is less than small_dens after the advance, and if so we reject the timestep and retry the step.

The parameter castro.retry_neg_dens_factor is removed.

## PR checklist


maxpkatz:
This PR is part of a series of PRs that I am planning to simplify and clean up the restart logic. My end goal is to have a series of checks on both the old and the new states in the advance (is the density positive, is the internal energy positive, are there NaNs, are the timestep constraints satisfied, etc.) and if any one of them is violated we issue a retry.

Since we have retries and they are now the default, we should be more aggressive about refusing to advance if bad circumstances arise. This may result in some low resolution runs failing that otherwise would have silently continued with density resets, so we should be aware of this consequence.

zingale:
tests were here: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-04-09-004/index.html

PR #799, opened 2020-03-01
Implement Edd_factor in C++

maxpkatz:
## PR summary

Since we don't have an easy way to access `limiter` and `closure` in device code, these are parameters to the function (for now, at any rate). Also, those variables are made static members of the `Radiation` class so that we can make local copies of them for lambda captures.

## PR checklist


zingale:
where are these parameters set in C++?

zingale:
oh I see -- they are still set via the parmparse, but they are not static members.

PR #800, opened 2020-03-02
Convert sums to C++

maxpkatz:
## PR summary

The functions used by sum_integrated_quantities() now write their reductions in C++ instead of Fortran. Some unused functions are removed.

## PR checklist


zingale:
all tests pass.  

@guadabsb15 can you look at this and check to see if the diagnostic files output with this change agree with the current state?

guadabsb15:
So I ran the inputs and probin in ```Exec/science/wdmerger/tests/wdmerger_3D```, modifying the inputs with ```max_step=10``` , ```amr.max_level=0``` and ```castro.sum_interval=1``` . Then I compared the ```*_diag.out``` files from this branch with the ones on development using ```xxdiff```. There were "small" differences in some quantities in ```amr_diag.out``` , ```species_diag.out``` and ```grid_diag.out```. In ```grid_diag.out``` there were also significant differences in XCOM, YCOM and ZCOM with the development branch showing  numbers of the order of 10^-7 and this branch showing numbers of 10^-31

maxpkatz:
The differences in those quantities like the centers of mass don't matter because at t = 0 the COM is zero in all three dimensions, so this is just different handling of roundoff error. So I think the PR is fine.

PR #804, opened 2020-03-07
convert ctoprim to C++

zingale:
## PR checklist


zingale:
the remaining bit to change is `inertial_to_rotational_velocity`

zingale:
These are the current test failures:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-03-17-002/index.html

maxpkatz:
For Detonation-collision-retry and Detonation-collision-retry-SDC, they both change the number of timesteps taken, but the former doesn't appear to fail in the comparison because apparently there was an earlier plotfile in the benchmark to compare against (the new version stops at step 16, the old one stopped at step 18 but apparently the benchmarks had a plotfile at timestep 16).

PR #810, opened 2020-03-09
Transverse optimizations

maxpkatz:
## PR summary

trans_single and trans_final are split into the individual qm and qp loops, and then we drop the local arrays of edge data. This helps for larger networks. We also create inline function versions of upass_map and qpass_map.

## PR checklist


Source/hydro/Castro_hydro.H
```diff
@@ -151,6 +151,22 @@
 #endif
                        amrex::Real hdt, amrex::Real cdtdx);
 
+     void actual_trans_single(const amrex::Box& bx,
+                              int idir_t, int idir_n, int d,
+                              amrex::Array4<amrex::Real const> const q,
+                              amrex::Array4<amrex::Real> const qo,
```
zingale: can we use a different name that `q` to avoid shadowing the `q` MF?

Source/hydro/Castro_hydro.H
```diff
@@ -151,6 +151,22 @@
...
+                              amrex::Array4<amrex::Real const> const qaux,
+                              amrex::Array4<amrex::Real const> const flux_t,
```
zingale: also use different name for `qaux`

PR #814, opened 2020-03-13
convert the PLM routines to C++

zingale:
## PR checklist


Source/hydro/Castro_mol.cpp
```diff
@@ -6,10 +6,260 @@
...
+  Real lconst_grav =  gravity->get_const_grav();
+#else
+  Real lconst_grav = 0.0_rt;
+#endif
+
+  AMREX_PARALLEL_FOR_4D(bx, NQ, i, j, k, n,
+  {
+
+   // piecewise linear slopes
+   uslope(bx, idir,
```
maxpkatz: I think this call to `uslope` should be outside the 4D parallel for, right?

Source/hydro/Castro_mol.cpp
```diff
@@ -6,10 +6,260 @@
...
+          flatn_arr, dq);
+
+
+   // this is a loop over zones.  For each slope in the zone, fill the
+   // two adjacent edge states (e.g., the right state at i-1/2 and the
+   // left state at i+1/2
+
+   if (idir == 0) {
+
+     if (plm_well_balanced == 1 && n == QPRES && idir == AMREX_SPACEDIM-1) {
```
maxpkatz: For CUDA we'll need local copies of `plm_well_balanced`.

Source/hydro/Castro_mol.cpp
```diff
@@ -6,10 +6,260 @@
...
+  bool lo_bc_test = lo_bc[idir] == Symmetry;
+  bool hi_bc_test = hi_bc[idir] == Symmetry;
+
+  // we have to do this after the loops above, since here we will
+  // consider interfaces, not zones
+
+  if (idir == 0) {
+    if (lo_bc_test) {
+
+      AMREX_PARALLEL_FOR_3D(bx, i, j, k,
```
maxpkatz: To match what we currently do, these loops have to be over a larger box that includes the ghost zones (grown by `dg[idir]` in each orthogonal direction).

Source/hydro/trace_plm.cpp
```diff
@@ -0,0 +1,297 @@
...
+  } else {
+    QUN = QW;
+    QUT = QU;
+    QUTT = QV;
+  }
+
+  Real lsmall_dens = small_dens;
+  Real lsmall_pres = small_pres;
+
+  GpuArray<int, npassive> qpass_map_p;
```
maxpkatz: I recommend using the new `upassmap` and `qpassmap` inline functions rather than this array capture.

Source/hydro/trace_plm.cpp
```diff
@@ -0,0 +1,297 @@
...
+      if ((idir == 0 && i >= vlo[0]) ||
+          (idir == 1 && j >= vlo[1]) ||
+          (idir == 2 && k >= vlo[2])) {
+
+        Real spzero = un >= 0.0_rt ? -1.0_rt : un*dtdx;
+        qp(i,j,k,n) = q_arr(i,j,k,n) + 0.5_rt*(-1.0_rt - spzero)*dq(i,j,k,n);
+#if  PRIM_SPECIES_HAVE_SOURCES
+        qp(i,j,k,n) += 0.5_rt*dt*srcQ(i,j,k,n);
+#endif
+
```
maxpkatz: Looks like the closing bracket is misplaced here for the right state validity check.

Source/hydro/Castro_mol.cpp
zingale: yes, you're right!

Source/hydro/Castro_mol.cpp
zingale: yes

Source/hydro/Castro_mol.cpp
zingale: I think that was not necessary actually.  This box is already bigger than needed since it is the valid box + 1 zone in all directions, plus and minus.  I don't know why we need the extra buffer there for 2nd order.

Source/hydro/trace_plm.cpp
zingale: indeed.  Good catch.

Source/hydro/trace_plm.cpp
zingale: done

Source/hydro/Castro_mol.cpp
maxpkatz: OK, I agree. And anyway the Fortran code was not right for CUDA, it looks like we missed that when we did the original port of that code.

PR #788, opened 2020-02-23
Add capability to call the EOS in C++

maxpkatz:
## PR summary

We can now call the EOS from C++. `ca_compute_temp` is converted to C++ (except for the part dealing with `clamp_ambient_temp`, which is deferred for now), as well as `reset_edge_state_thermo`.

## PR checklist

- [x] test suite needs to be run on this PR
- [x] this PR will change answers in the test suite
- [ ] all functions have docstrings as per the coding conventions
- [ ] the `CHANGES` file has been updated
- [ ] if appropriate, this change is described in the docs

Issue #702, opened 2019-11-28
reset_internal_energy is unnecessary


maxpkatz:
Consider the algorithm for allow_small_energy == 0, which is the stricter version we should ideally enforce. First, it computes small_e = EOS(rho, small_temp). Then it checks if eint < small_e. If so it resets eint it with an EOS call; if not, it applies the dual energy criterion, then checks if eint < small_e and if so resets eint with an EOS call.

There is a simpler way to do this. First, we can apply the dual energy criterion, which either updates UEINT or leaves it alone. In either case, we might get a UEINT that is below small_e. Now, the reason the first EOS call is there is to determine what small_e actually is for this zone. But the EOS should indirectly enforce small_e, because it has a small_temp and small_dens and will floor out if these are violated. So we can simply call the EOS after the dual energy criterion has been applied, relying on the EOS to provide the flooring, and we are done.

And if we're doing that, we can simply merge reset_internal_e into compute_temp, by first applying the dual energy criterion and then calling the EOS, giving us both the correct e and T in one shot. The main thing we have to be careful of is that if we apply eos_input_is_constant, that we ignore that parameter if the internal energy floors, and allow it to be updated in that case. In other cases we can leave the internal energy as it is when we're calling with eos_input_re.

PR #817, opened 2020-03-14
Convert reset_internal_energy to C++ and simplify the algorithm

maxpkatz:
## PR summary

We now simply evaluate the eint corresponding to small_temp, and reset eint (and eden) as needed to be at least that large. We then apply the dual energy criterion. This makes the behavior corresponding to `allow_small_energy == 0` the default, and removes that parameter.

## PR checklist


maxpkatz:
This addresses part of #702 but not all of it (reset_internal_energy is still a discrete routine).

zingale:
Note: this seems to dramatically help the Sod stellar problem (Sod_stellar-test1-2d):

development:
Sod_stellar-test1-2d-development

this PR:
Sod_stellar-test1-2d-reset_internal_energy

Source/driver/Castro_setup.cpp
```diff
@@ -193,6 +193,17 @@ Castro::variableSetUp ()
...
+  // with the minimum permitted by the EOS, and vice versa.
+
+  Real new_min_T = std::max(small_temp, EOSData::mintemp);
+  small_temp = new_min_T;
+  EOSData::mintemp = new_min_T;
+
+  Real new_min_rho = std::max(small_dens, EOSData::mindens);
+  small_dens = new_min_rho;
+  EOSData::mindens = new_min_rho;
+
```
zingale: does Fortran also know about this?

zingale:
tests are here: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-03-14-006/index.html

Source/driver/Castro_setup.cpp
maxpkatz: Yes, we have the same logic in Fortran, it's just written in a slightly different way. We should have done this in the original EOS pull request (#788); that was an oversight on my part.

maxpkatz:
I had missed that SDC called `ca_reset_internal_e` directly. So I moved the actual implementation to a separate interface that takes a box and an Array4, and then implemented both the MultiFab interface and that call from SDC with it. Please retest.

zingale:
here are the tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-03-15/index.html

PR #819, opened 2020-03-17
Switch over to the new Fab interface for BC fills

maxpkatz:
## PR summary

This is in preparation to use the new fab_filcc in AMReX.

generic_single_fill and generic_multi_fill are merged into one function, generic_fill, since the Fab interface has no notion of separate single and group interfaces. Similarly, the calls to denfill and hypfill are now both done in one function (though the functions are not merged, to preserve user-facing code).

The source fill routines are also merged into the generic fill routines.

Finally, instead of locally overriding the BCs in the boundary fill routine, we now globally override any inflow boundary conditions (for the non-State_Type data) to become FOEXTRAP. So this is the same result but with code that is a little simpler and will be easier to link up to AMReX later.

## PR checklist


Source/driver/Castro_generic_fill.cpp
```diff
@@ -6,89 +6,37 @@
...
-#endif
-
-    // Note that these are called with dimension agnostic macros like
-    // AMREX_ZFILL and AMREX_INT_ANYD already, so we should expect that
-    // everything has three entries, not AMREX_SPACEDIM entries.
-    // We still choose to use the macros anyway below, for compatibility
-    // with the GPU pragma script.
+    // Make a copy of the raw BCRec data in the format
+    // our BC routines can handle (a contiguous array
+    // of integers).
```
zingale: will this be undone when we move the fills to C++?

zingale:
before we merge this, can you comment on how hard it will be to sync up MHD?

Source/driver/Castro_generic_fill.cpp
maxpkatz: The AMReX calls in C++ can directly take the Vector<BCRec> without any preprocessing on our parts. For the C++ calls that we implement, we won't need to serialize the vector like this.

maxpkatz:
> before we merge this, can you comment on how hard it will be to sync up MHD?

This change doesn't really conflict with MHD. But it wouldn't be that hard to sync up the MHD face fill calls with this strategy.

Issue #821, opened 2020-03-19
Abort if initialization generates small data


maxpkatz:
We should abort if initdata gives us density or temperature that are below the corresponding "small" quantities. We need to do this in a way that respects the user not setting T in initdata.

PR #822, opened 2020-03-20
small_temp and small_dens check

zingale:
## PR summary

This adds a check just after initialization that ensures that the problem setup did not try to set the temperature or density too close to the small cutoff values.

Addresses #821 

## PR checklist


zingale:
this changes answers to EOS tolerance due to the new computeTemp call after initialization.

Source/driver/Castro.cpp
```diff
@@ -954,6 +954,41 @@ Castro::initData ()
...
+
+           AMREX_PARALLEL_FOR_3D(bx, i, j, k,
+           {
+
+             // if the problem tried to initialize a thermodynamic
+             // state that is at or below small_temp, then we abort.
+             // This is dangerous and we should recommend a smaller
+             // small_temp
+             if (S_arr(i,j,k,UTEMP) < lsmall_temp * 1.001) {
+               amrex::Error("Error: initial data has T <~ small_temp");
```
maxpkatz, line 982: To work as desired in the CUDA build, we should do this as a max reduction on an integer flag that starts as 0 and is set to 1 if the floor is violated, then throw the error outside the parallel for.

zingale: can you point me to an example?

maxpkatz: Yes, look at what we do with `burn_failed` in `react_state`.

Source/driver/Castro.cpp
```diff
@@ -954,6 +954,56 @@ Castro::initData ()
...
+
+           AMREX_PARALLEL_FOR_3D(bx, i, j, k,
+           {
+
+             // if the problem tried to initialize a thermodynamic
+             // state that is at or below small_temp, then we abort.
+             // This is dangerous and we should recommend a smaller
+             // small_temp
+             if (S_arr(i,j,k,UTEMP) < lsmall_temp * 1.001) {
+               *init_failed_T_d += 1;
```
maxpkatz, line 988: In order to be thread-safe on the device, these += statements need to be done in a thread-safe way for the CUDA build (this is fine for OpenMP). See how we use `deviceReduceSum` in `sum_utils.cpp`.

Issue #809, opened 2020-03-08
Geometry source terms should be traced


zingale:
In trace_ppm.cpp, the geometric source terms are added to the states with a mysterious `eta` factor that depends on the Courant number.  We should instead include the geometric source terms into the characteristic tracing, as in the original PPM paper.

zingale:
In particular, it looks like we are using the PLM tracing for PPM

zingale:
Here's a derivation of the geometric sources terms for the primitive system we solve
[euler_geometric_source_terms.pdf](https://github.com/AMReX-Astro/Castro/files/4364153/euler_geometric_source_terms.pdf)

zingale:
One idea was to simply add these in src_to_prim, but the problem there is that these sources depend on the direction of the reconstruction.  When we are doing the tracing in x (r), the sources are present, but they are not present in any of the other tracing directions.  This directional dependence doesn't exist for any of the other primitive sources.

zingale:
I think that the cleanest way to do this is to create a function `add_geometry_sources()` that is called in `trace_ppm.cpp` when we are doing the reconstruction of the other source terms.  This will depend on `idir` and we will need to explicitly set `do_trace` for these components to be true.

zingale:
another issue with trying to do this in `src_to_prim` is that we now call `src_to_prim` when doing the shock detection, so we really do want that to be directionally independent.

PR #824, opened 2020-03-21
hook the geometry sources into the main source tracing

zingale:
## PR summary

This includes the geometric source terms into the characteristic projection for PPM.  Addresses #809 

## PR checklist


zingale:
I think that the cleanest way to do this would be to hook the geometry source terms directly into src_to_prim

Issue #828, opened 2020-03-30
eliminate src_q


zingale:
We only need `src_q` for the CTU methods (including simplified-SDC).  We can eliminate this MF as well as the `QDPDR` and `QDPDE` entries in `qaux` by embedding the conversion of the conserved sources to primitive within the hydro routine itself.  In particular, in `Castro_ctu_hydro.cpp` we can do the conversion in the MFIter loop before we construct the interface states.  This means we only need `src_q` as a FAB.

zingale:
one tricky bit is that if we are doing simplfied-SDC, then we still need to add the reactive sources, as we currently do after the call to srctoprim:
```
        if (time_integration_method == SimplifiedSpectralDeferredCorrections) {                                                                                                                                           
                                                                                                                                                                                                                          
            MultiFab& SDC_react_source = get_new_data(Simplified_SDC_React_Type);                                                                                                                                         
                                                                                                                                                                                                                          
            if (do_react)                                                                                                                                                                                                 
                src_q[mfi].plus<RunOn::Device>(SDC_react_source[mfi],qbx,qbx,0,0,NQSRC);                 
```

PR #829, opened 2020-03-30
Remove src q

zingale:
## PR summary

This removes the src_q multifab and instead converts the conserved state sources to primitive state sources FAB by FAB just before they are needed (in the creation of the interface states).  This saves on memory from the MF as well as allows us to eliminate two entries in qaux (QDPDE and QDPDR).  The trade-off is we now do an EOS call.

Answers may change because of the additional EOS call.

## PR checklist

### DIFF
--- a/Source/driver/Castro.H
+++ b/Source/driver/Castro.H
@@ -1077,12 +1077,6 @@ protected:
 
 
 ///
-/// The source terms in primitive form.
-///
-    amrex::MultiFab src_q;
-
-
-///
 /// Source terms to the hydrodynamics solve.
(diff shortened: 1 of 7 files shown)


zingale:
I think I should be able to do the EOS call as `eos_input_rt` instead of `eos_input_re`, since we've already set the temperature by this point.

zingale:
addresses #828

zingale:
Here are the test results:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-03-30-001/index.html

Issue #816, opened 2020-03-14
revisit well-balanced methods


zingale:
We have several different well-balanced approaches in the code, all for piecewise linear.

For CTU, we have `use_pslope`, which is on by default and the newer `plm_well_balanced`.  It seems that are mostly similar, but we should consider getting rid of `use_pslope`.  Also, if we do reflecting BCs correctly, we should be able to eliminate the need for HSE BCs.  This is done for the 4th order method and some attempt of it is done for PLM + MOL.

zingale:
it turns out that `use_pslope` is essentially a generalization of the `plm_well_balanced`
 / Kappeli method.  The notes here demostrate that.

[pslope.pdf](https://github.com/AMReX-Astro/Castro/files/4411189/pslope.pdf)

zingale:
I'm going to do 2 changes.  First: I think for `use_pslope`, we should always use p_i as the reference for the HSE.  Second, I will remove `plm_well_balanced`.  As the notes show, we were both not doing it correctly at the end and we were overwriting the results with `use_pslope` anyway.

zingale:
Note: in order to use the new pslope stuff for 2nd-order SDC, we need to bring src_q into `mol_plm_reconstruct`.  But at the moment, we don't actually create src_q` for SDC.  Merging #829 will make this a lot easier.

PR #831, opened 2020-03-31
remove plm_well_balanced and update use_pslope

zingale:
now use_pslope always uses zone i for the HSE reference point

## PR checklist

### DIFF
--- a/Source/hydro/slope.cpp
+++ b/Source/hydro/slope.cpp
@@ -52,71 +52,11 @@ Castro::uslope(const Box& bx, const int idir,
     // second-order -- piecewise linear slopes
 
     const int lplm_limiter = plm_limiter;
-    const int lplm_well_balanced = plm_well_balanced;
 
     AMREX_PARALLEL_FOR_3D(bx, i, j, k,
     {
 
-      if (lplm_well_balanced == 1 && n == QPRES && idir == AMREX_SPACEDIM-1) {
-        // we'll only do a second-order pressure slope,
-        // but we'll follow the well-balanced scheme of
-        // Kappeli.  Note at the moment we are assuming
-        // constant gravity.
-
-        Real p0;
-        Real pp1;
-        Real pm1;
-
-        if (idir == 0) {
-          p0 = 0.0_rt;
-          pp1 = q_arr(i+1,j,k,QPRES) - (p0 + 0.5_rt*dx[0]*(q_arr(i,j,k,QRHO) + q_arr(i+1,j,k,QRHO))*lconst_grav);
-          pm1 = q_arr(i-1,j,k,QPRES) - (p0 - 0.5_rt*dx[0]*(q_arr(i,j,k,QRHO) + q_arr(i-1,j,k,QRHO))*lconst_grav);
-
-          if (i == domlo[0] && lo_bc_test) {
-            pm1 = 0.0_rt;  // HSE is perfectly satisfied
-          }
-
-          if (i == domhi[0] && hi_bc_test) {
-            pp1 = 0.0_rt;
-          }
-
-        } else if (idir == 1) {
-          p0 = 0.0_rt;
-          pp1 = q_arr(i,j+1,k,QPRES) - (p0 + 0.5_rt*dx[1]*(q_arr(i,j,k,QRHO) + q_arr(i,j+1,k,QRHO))*lconst_grav);
-          pm1 = q_arr(i,j-1,k,QPRES) - (p0 - 0.5_rt*dx[1]*(q_arr(i,j,k,QRHO) + q_arr(i,j-1,k,QRHO))*lconst_grav);
-
-          if (j == domlo[1] && lo_bc_test) {
-            pm1 = 0.0_rt;  // HSE is perfectly satisfied
-          }
-
-          if (j == domhi[1] && hi_bc_test) {
-            pp1 = 0.0_rt;
-          }
-
-        } else {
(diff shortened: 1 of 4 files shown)


zingale:
This addresses #816

zingale:
pslope now works with sdc-2 as well and gives great results on the hse_convergence test problem.  It essentially does what plm_well_balanced did before but now works for non-constant gravity.

zingale:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-04-02-001/index.html

that's the test results.  

I need to think about why the plm tests without a source term changed.

This also highlights that we don't test something where the well-balancing should have an effect.

zingale:
Here's some results from the ``hse_convergence`` test problem:

  * ``plm.converge.out``: default inputs (HSE BCs, ``use_pslope = 1``)

    * ``development``::

         64         487233.46399          1016837.3102
         128        244861.7364            512627.30949
         256        122761.58066           257385.35067
         512         61467.736267          128964.94629

    * ``pslope``::

         64         626452.88316          1182784.3193
         128        313048.41579           594603.36359
         256        156487.63673           298114.02046
         512         78235.586812          149261.24801

    * observations: the changes on the ``pslope`` branch make this
      slightly worse, but none of them converge second order and the
      velocities are too high.

  * ``plm-hsereflect.out``: change the HSE boundary conditions to use
    ``castro.hse_interp_temp=1`` and ``castro.hse_reflect_vels=1``

    * ``development``::

         64             73.506892196         4813.3864853
         128            14.535760251         1214.117429
         256             7.5935709488         304.82857215
         512             2.0570017194          76.375510114

    * ``pslope``::

         64             13.70622633          4783.9096003
         128            43.447644433         1210.3056976
         256             3.696649138          304.35175517
         512             1.3852275813          76.31695564

    * observations: both have low velocities and converge as second
      order.  Not much difference between the two.

  * ``plm-reflect.converge.out``: change the upper and lower
    boundaries to reflect

    * ``development``::

         64             18.874200656          639.42247572
         128             2.7728314111         154.63440792
         256             0.057978782978        37.100898515
         512             0.12647077385          8.8987132485

    * ``pslope``::

         64             17.976434908          631.92731482
         128             2.9819013642         152.4130912
         256             0.41875360596         36.409723879
         512             0.02156768615          8.7387498788

    * observations: both have low velocities and converge as second
      order.

  * ``plm-reflect-nopslope.converge.out``: reflecting boundaries but with
    ``use_pslope=0``

    * ``development``::

         64            177.70705679       1324129.692
         128            12.373764958       663443.58475
         256             1.0455805935      332063.12615
         512             0.98402053661     166116.2066

    * ``pslope``: same as ``development``

  * ``plm-hsereflect-nopslope.converge.out``:

    * ``development``::

         64             72.737753365         4841.0604709
         128            39.608201512         1217.4941755
         256             7.4871428318         305.25678921
         512             2.9431763011          76.428801698


General PLM observations: 

  * If we are not using a well-balanced method (``use_pslope=0``) then
    the only case that shows second order convergence with low velocites
    is when we use HSE BCs with reflecting velocity and temperature interp.

    It would be interesting to look to see if the temperature bit is
    important (maybe we can get away with an extrapolation of the
    temperature)

  * If we use a well-balanced method (``use_pslope=1``), then either
    HSE+reflect vel or a pure symmetry boundary (reflect) give great
    results.

Source/hydro/slope.cpp
```diff
@@ -280,60 +227,120 @@ Castro::pslope(const Box& bx, const int idir,
...
+      // we need dp_{i+1} = p_{i+1} - p_i
+      // and     dp_{i+2} = p_{i+2} - p_{i-1}
+      //
+      // then we compute
+      //   Dp_{i+1} = min{ 2 |dp_{i+1}|, 2 |dp_{i+2}|, 1/2 |dp_{i+1} + dp_{i+2}|}
+      //
+      // we use the value above with HSE removed
+
+      Real dlftp1 = pp1 - p0;
+      Real drgtp1 = pp2 = pp1;
```
maxpkatz, line 342: Typo here, should be `pp2 - pp1`.

Source/hydro/slope.cpp
```diff
@@ -280,60 +227,120 @@ Castro::pslope(const Box& bx, const int idir,
...
 
-        // this is dp{i+2} = p_{i+2} - p_{i+1}, but we subtract off the
-        // HSE state integrated from p_{i+1}
-        drgtp1 = q_arr(i+2,j,k,QPRES) - q_arr(i+1,j,k,QPRES);
-        drgtp1 += -0.25_rt * (q_arr(i+1,j,k,QRHO) + q_arr(i+2,j,k,QRHO)) *
-          (src(i+1,j,k,QU)+src(i+2,j,k,QU))*dx[0];
+        pm1 = q_arr(i-1,j,k,QPRES) - pm1_hse;
+        pm2 = q_arr(i-2,j,k,QPRES) - pm2_hse;
+
+        if (i == domlo[0] && lo_bc_test) {
```
maxpkatz, line 267: If we enforce the symmetry boundary here, I think the limiter can break it. So I suspect we need to do it after the limiter has been applied. I think it's OK to violate the limiter in this case.

maxpkatz: Actually, this is OK. The second and fourth order limiters will always respect a slope of zero and ensure the final slope is zero.

zingale:
the pure hydro problems with plm now give roundoff diffs

PR #839, opened 2020-04-06
Convert normalize_species and compute_cfl to C++

maxpkatz:
## PR summary

Some Castro class functions are moved from protected to public since that's necessary for CUDA.

## PR checklist


Source/hydro/Castro_hydro.cpp
```diff
@@ -245,22 +245,132 @@ Castro::check_for_cfl_violation(const Real dt)
...
+                dtdx = dt / dx[2];
+            }
+
+            Real courx = (qaux_arr(i,j,k,QC) + std::abs(q_arr(i,j,k,QU))) * dtdx;
+            Real coury = (qaux_arr(i,j,k,QC) + std::abs(q_arr(i,j,k,QV))) * dtdy;
+            Real courz = (qaux_arr(i,j,k,QC) + std::abs(q_arr(i,j,k,QW))) * dtdz;
+
+            if (ltime_integration_method == 0) {
+
+                // CTU integration constraint
```
zingale, line 289: actually, this version is valid for time_integration_method = 0 or 3

maxpkatz: I suspect we missed this routine when we did simplified SDC.

PR #844, opened 2020-04-09
Rework check_timestep to be based on change_max

maxpkatz:
## PR summary

check_timestep is currently effectively useless for CFL violations that occur as the result of an advance, since it only refers to old data, and for burning timestep limiting it uses an average of the old and new data. With this change, we unify how they are used so that they both refer to only the new data, and reject any timestep that would cause dt to shrink by more than a factor of change_max. check_timestep is now effectively the inverse operation of how change_max is used in estdt: whereas in estdt, change_max allows the timestep to grow by no more than that factor, in
check_timestep it allows the timestep to shrink by no more than that factor without doing a retry. In effect, this prevents the state from changing too much in either direction in a single timestep.

## PR checklist


zingale:
test suite results: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-04-09-003/index.html

PR #849, opened 2020-04-12
eliminate the intiial model from the BCs

zingale:
## PR summary

This removes the initial model arrays from the boundary condition routines.  The option of an interpolated boundary is now gone, and in its place we use the ambient state BCs.  The motivation for this change is to reduce the memory footprint on the GPU.

## PR checklist


zingale:
The 2-d no-boost 1000Hz XRB runs fine with these changes on Summit, and is about 5% faster than the `development` version.  Still need to do some visualization on it.

zingale:
I think this is ready for review now.  The reacting_bubble test now has different BCs, but the result still looks as expected, even though it doesn't pass the tests.

Source/problems/bc_ext_fill_nd.F90
```diff
@@ -76,20 +76,13 @@ subroutine ext_fill(lo, hi, adv, adv_lo, adv_hi, &
...
-                   x = problo(1) + delta(1)*(dble(domlo(1)) + HALF)
-
-                   call interpolate_sub(dens_above, x, idens_model)
-                   call interpolate_sub(temp_above, x, itemp_model)
-
-                   do m = 1, nspec
-                      call interpolate_sub(X_zone(m), x, ispec_model-1+m)
-                   end do
-
+                   ! sometimes, we might be working in a corner where
```
maxpkatz, line 80: This comment does not still make sense, I think we should remove this conditional logic entirely, our requirement is that we should always be integrating from a valid state, and if there's a combination where that's not true (like a corner with two intersecting inflow boundaries) then we should throw an error at initialization saying we don't support it.

zingale: I think I've addressed this now

Source/problems/Castro_bc_fill_nd.cpp
```diff
@@ -67,6 +67,61 @@ void ca_statefill(Box const& bx, FArrayBox& data,
...
+    // corner where two (or three) faces want to do HSE, we may run
+    // into a situation that the data is not valid in the corner where
+    // we start the integration.  We'll abort, for now, if we run into
+    // this case.
+    //
+    // The future fix is to first call ext_fill on the ghost cells
+    // that are not corners and then call it a second time on just the
+    // corners.
+
+    std::cout << "about to check" << std::endl;
```
maxpkatz, line 81: Can you remove the debugging output please?

PR #858, opened 2020-04-16
Switch to AMReX ReduceOps reduction approach

maxpkatz:
## PR summary

The original approach to reductions in the CUDA Fortran code was to use AMREX_MFITER_REDUCE_SUM, combined with a call to reduce_add, reduce_min, etc. When we switched to C++ with amrex::ParallelFor, we kept the same approach but called the equivalent C++ call from AMReX instead of reduce_add et al. I recently discovered that there were some edge cases where this is broken, and switched entirely to using atomics in commit b881808bbd1f518c60123b8b195c0d18b8bfff36. While the safety issue was fixed, performance of reductions dropped enough that it is now a relevant factor. 

So we are now switching (hopefully for good) to the AMReX reducer implementation that uses ReduceOps, ReduceData, and ReduceTuple. Some examples can be seen in amrex/Tutorials/GPU/ParallelReduce. In this approach we define a tuple of data to reduce over, along with associated reduction operations for each member in the tuple, and then our lambda function returns the data to reduce instead of being a void. The lambda function is evaluated by the reduce operation, and then we get the data at the end with amrex::get(). This operation is thread-safe, so in CUDA we don't need any special handling (that is done by AMReX) and for OpenMP we don't need to do a reduction (it is done atomically behind the scenes in AMReX; this is less efficient but probably doesn't matter for us).

Going forward, all new conversions to C++ that involve reductions should use this approach.

## PR checklist

### DIFF
--- a/Source/driver/Castro.cpp
+++ b/Source/driver/Castro.cpp
@@ -956,14 +956,15 @@ Castro::initData ()
        // within any small limits
        computeTemp(S_new, cur_time, S_new.nGrow());
 
-       int init_failed_rho = 0;
-       int init_failed_T = 0;
+       ReduceOps<ReduceOpSum, ReduceOpSum> reduce_op;
+       ReduceData<int, int> reduce_data(reduce_op);
+       using ReduceTuple = typename decltype(reduce_data)::Type;
 
 #ifdef _OPENMP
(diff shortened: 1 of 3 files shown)


zingale:
this passes tests

PR #861, opened 2020-04-18
move some more derives to C++

zingale:
## PR checklist


Source/driver/Derive.cpp
```diff
@@ -396,418 +380,520 @@ extern "C"
...
-                         const Real* time, const Real* dt, const int* bcrec, 
-                         const int* level, const int* grid_no)
+    void ca_derradialvel(const Box& bx, FArrayBox& derfab, int dcomp, int /*ncomp*/,
+                         const FArrayBox& datfab, const Geometry& geomdata,
+                         Real /*time*/, const int* /*bcrec*/, int /*level*/)
     {
 
-        IntVect ilo(D_DECL(lo[0], lo[1], lo[2]));
-        IntVect ihi(D_DECL(hi[0], hi[1], hi[2]));
+      auto const dat = datfab.array();
```
maxpkatz, line 444: For the derived fields that depend on location (radial velocity and angular momentum) we should use the `position` function rather than computing the position with `problo + dx * (idx + 0.5))`. The `position` function wraps around appropriately for periodic boundaries and this is relevant for correct evaluation of angular momentum for periodic problems.

maxpkatz: Actually, since this port is faithful to the Fortran, we can do that in a separate PR. I'm not sure if it even matters for the derive functions.

zingale:
test results: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-04-18-004/index.html

PR #863, opened 2020-04-18
Move burn weightings into Reactions_Type

maxpkatz:
## PR summary

We currently have a scheme for storing reactions weightings, which are a measure of the number of RHS evaluations during the burn and therefore a proxy for the difficulty of the burn. These weights were added as separate StateData depending on the runtime option use_custom_knapsack_weights.

This change instead places the weights directly in the Reactions_Type StateData as a new component. Now that we always store the weights, load balancing and other future changes that depend on the weights are easier to manage.

The only thing we have to be careful of now is that code that previously zeroed out all of the reactions components cannot be applied until after the weights have been used for their intended purpose. (For load balancing purposes, we want the minimum weight to be one, not zero.) For safety, then, we now don't zero out the reactions components in the strang_react functions, but instead handle the setting of those components directly in the react_state loop.

The parameter use_custom_knapsack_weights is now renamed use_burn_load_balancing so that it is clearer what is does.

In anticipation of future PRs, and because we burn on some ghost zones (the physical boundaries and coarse-fine boundaries), the number of ghost zones in Reactions_Type is increased to 4.

The checkpoint version has now been incremented; this version of the code will not be able to restart from a checkpoint generated by earlier versions of the code.

## PR checklist


maxpkatz:
Tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-04-19-011/index.html

The tests exposed an indexing bug in Fortran, which has now been fixed.

PR #870, opened 2020-04-20
Throw error if CFL is too large for density flux limiter

maxpkatz:
## PR summary

The paper we derive the flux limiter from observes that the first order Lax-Friedrichs flux only guarantees positivity if CFL <= 0.5, so this should be a requirement.

## PR checklist

### DIFF
--- a/Source/driver/Castro.cpp
+++ b/Source/driver/Castro.cpp
@@ -372,6 +372,10 @@ Castro::read_params ()
       }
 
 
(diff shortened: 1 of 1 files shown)


maxpkatz:
Closing. Even though this is a requirement for L-F positivity, we have a stronger requirement of rho >= small_dens, which means that we often override the L-F flux by directly linearly scaling the final flux, and this will catch any situation where the L-F flux didn't resolve in a positive density due to CFL being too large.

PR #871, opened 2020-04-20
move the check on corner external BCs to the extern routine itself

zingale:
this way it is specific to HSE

## PR checklist


maxpkatz:
Can we query on the HSE boundary conditions from C++? Maybe another way to do this is to only throw this error if it's inflow + HSE. I'm just worried about putting error checks in the Fortran since those won't be active for the CUDA build.

zingale:
yeah, I can try that

zingale:
I don't think this is relevant anymore

Issue #813, opened 2020-03-13
move runtime parameters into their own namespace


zingale:
We want to do is move all of the runtime parameters to be global variables in a (say) `Parameters` namespace, and then we can just put using namespace `Parameters` everywhere.

This will allow us to lambda-capture the parameters on the GPU.
They should be `extern AMREX_GPU_MANAGED` like we do in Microphysics now

zingale:
The C++ way to initialize this with namespaces is to do:

`test.H`:
```
#ifndef _TEST_H_
#define _TEST_H_

namespace test {
    extern double x;
};

#endif
```

`test.cpp`:
```
#include "test.H"

double test::x = 1.0;
```

with `main.cpp`:
```
#include <iostream>
#include "test.H"

int main() {

  std::cout << test::x << std::endl;

  return 0;
}
```

PR #873, opened 2020-04-21
move runtime parameters into a namespace as globals 

zingale:
## PR checklist

### DIFF
--- a/Source/diffusion/diffusion_params.cpp
+++ b/Source/diffusion/diffusion_params.cpp
@@ -1,0 +1,5 @@
+#include <AMReX_BLFort.H>
+
+#include <castro_limits.H>
+
+#include "diffusion_defaults.H"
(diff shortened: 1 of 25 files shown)


zingale:
This will address #813

zingale:
This seems to work.  All runtime parameters are now globals, but in a namespace, `castro::`, `::gravity::`, or `diffusion::`.  To make life easier, the castro namespace is used in `Castro.H`, so we can drop the namespace qualifier for them.  But for the others, we cannot (we can't load all into the default namespace because of name collisions).

zingale:
I think this works now on CPU and GPU.

Source/diffusion/diffusion_params.cpp
```diff
@@ -0,0 +1,7 @@
+#include <AMReX_BLFort.H>
```
maxpkatz, line 1: Perhaps we can auto-generate these files too?

PR #877, opened 2020-04-24
Remove burn load balancing functionality

maxpkatz:
## PR summary

This functionality was not used much and makes maintenance of the reactions harder; plus, it does not make as much sense on accelerator-based architectures as it did on CPU-only architectures.

## PR checklist


zingale:
do you expect the big diffs in wdmerger_collision?
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-04-24-003/index.html

maxpkatz:
We are undoing the big diffs that were done in #863 when the load balancing was turned on for wdmerger_collision (for test coverage purposes -- this is no longer needed, as we're removing the feature in question).

PR #878, opened 2020-04-25
Add a max level for explicitly doing burns

maxpkatz:
## PR summary

Above this max level, we interpolate from reactions data on the coarse level.

## PR checklist


maxpkatz:
I've updated the `inputs-collision` and `inputs-collision.testsuite` test cases to exercise this functionality.

PR #879, opened 2020-04-25
Port check_timestep to C++

maxpkatz:
## PR checklist


Source/driver/timestep.cpp
```diff
@@ -223,3 +223,132 @@ Castro::estdt_temp_diffusion(void)
...
+
+                eos(eos_input_re, eos_state);
+
+                Real u = std::abs(U(i,j,k,UMX)) * rhoInv;
+                Real v = std::abs(U(i,j,k,UMY)) * rhoInv;
+                Real w = std::abs(U(i,j,k,UMZ)) * rhoInv;
+
+                Real c = eos_state.cs;
+
+                Real tau_CFL = amrex::min(AMREX_D_DECL(dx[0] / (c + u), dx[1] / (c + v), dx[2] / (c + w)));
```
zingale, line 298: we should make this general for the true SDC stuff as well, with the different CFL definitions for different `time_integration_method`s

Source/driver/timestep.cpp
```diff
@@ -223,3 +223,132 @@ Castro::estdt_temp_diffusion(void)
...
+            if (castro::change_max * dt_tmp < dt) {
+                zone_failed = 1.0_rt;
+            }
+
+            return zone_failed;
+        });
+
+    }
+
+    ReduceTuple hv = reduce_data.value();
```
zingale, line 348: maybe we add the diffusion check too?

zingale:
this fails to compile with reactions on

Source/driver/timestep.cpp
maxpkatz: I have an idea for how to do this. Ideally we'd call `estdt` on the new-time data and then just check if that timestep is within `change_max` of the old-time `dt` we started with. Then we don't need to have a separate routine, which is likely to diverge in ways like this. This would also solve the diffusion problem.

PR #880, opened 2020-04-25
Replace check_timestep with call to estDt

maxpkatz:
## PR summary

As noted in #879, `check_timestep` had a few issues, including that it didn't cover diffusion or simplified SDC. We replace that custom code with a call to `estDt`, which evaluates the new data. Then we can automatically get the new-time timestep limiter, and compare to see if it's within `change_max` of the old timestep. If not, we consider the step to have failed.

## PR checklist


Source/driver/Castro.cpp
```diff
@@ -1473,7 +1473,7 @@ Castro::estTimeStep (Real dt_old)
 #endif
 
     if (verbose) {
-        amrex::Print() << "Castro::estTimeStep (" << limiter << "-limited) at level " << level << ":  estdt = " << estdt << '\n';
+        amrex::Print() << "Castro::estTimeStep (" << limiter << "-limited) at level " << level << ":  estdt = " << estdt << '\n' << std::endl;
```
zingale, line 1476: do you need both `\n` and `std::endl`?

maxpkatz: My general approach has been that most things should add two newlines so the stdout is easier to read.

PR #882, opened 2020-04-26
if we are running simplified SDC, we need numerical Jacobian

zingale:
Jacobian for now, since the analytic Jacobian is not yet in terms
of the proper state

## PR summary

The analytic Jacobian for simplified-SDC is not yet in terms of the right variables.

## PR checklist


zingale:
This is being addressed in https://github.com/starkiller-astro/Microphysics/pull/228

Issue #887, opened 2020-04-29
check_for_post_regrid is expensive in the CUDA build


maxpkatz:
This is because the logic for both TagBoxArray::collate() and the Castro code in check_for_post_regrid() is serial CPU code, which will become a scaling issue when the number of tagged zones is large.

PR #898, opened 2020-05-01
use_post_step_regrid unconditionally triggers a post-step regrid

maxpkatz:
## PR summary

use_post_step_regrid previously would check if the set of tagged zones at the old time did not cover all zones tagged at the new time, and conditionally asked for a post-step regrid if so. However this was expensive on GPUs due to the serial nature of the algorithm (#887). Rather than sort this out, we just unconditionally regrid at the end of every step when the user uses this option. While regridding is expensive, the expectation is that this option is only used when the user expects their tagging criteria to be unconditionally satisfied at all times, and this comes with a performance tradeoff.

## PR checklist


CHANGES.md
```diff
@@ -1,21 +1,13 @@
...
-     change. First, do `git checkout master` if you're not already on the
-     old master branch. Then do `git pull`. This will gather the updates
-     to the repo, but will fail with the message `Your configuration specifies
-     to merge with the ref 'refs/heads/master' from the remote, but no such ref
-     was fetched.` Then you can simply do `git checkout main` and your local
-     repo should automatically switch to that branch and track updates from
-     the upstream repo on GitHub. If you like, you can then delete the old
-     master branch with `git branch -D master`.
-
    * The CUDA build no longer has a requirement that amr.blocking_factor
```
zingale, line 11: why was this removed?

maxpkatz: That was an accident from the merge, I restored it.

zingale:
this causes big differences to the detonation sdc problem
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-07-03-002/index.html

maxpkatz:
> this causes big differences to the detonation sdc problem
> http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-07-03-002/index.html

What this changes is that this unconditionally regrids every step now rather than checking to see if a regrid is warranted. Since the existing method is a hack which does not exactly reproduce the AMReX tagging algorithm, it is to be expected that the regridding will be different now.

maxpkatz:
To be more specific, the old method would only force a regrid if it detected that there were zones tagged which did not already have a fine level covering them. So there are some coarse timesteps in the current code that didn't have end-of-step regrids. Now we regrid at the end of every coarse timestep. While this might cause relatively small diffs in non-reacting codes, it causes a fairly large difference when reactions on, since even small differences like that can be the difference between whether there are burn failures and a retry or not. The old code has nine retries for this problem while the new one has only one.

PR #909, opened 2020-05-03
split out the MHD Initialization so we can run MHD on any problem

zingale:
## PR summary

This gives us a separate Prob_mhd_nd.F90 that initiializes the magnetic field to 0.  Problems can implement their own version of this.  By providing this default we can run the MHD solver on any problem setup.

## PR checklist


zingale:
test result: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-05-03-003/index.html

Sod fails because of the change in the number of timesteps from a previous PR

zingale:
ah, thanks.  I'll update those problems and remove the prints

zingale:
I think I've updated these -- not sure if we still want B as a probin parameter in some of the hydro problems.

guadabsb15:
Yes, we don't really need those

PR #922, opened 2020-05-06
Strang burn cleanup

maxpkatz:
## PR summary

The Strang interface to react_state is cleaned up: rather than having an entry point in strang_react_first_half and strang_react_second_half, react_state is called directly.

The mask that was used to avoid burns on ghost zones is also removed.

## PR motivation

A note on the mask: while this was originally added under the logic that because burns are expensive, we prefer doing a ghost zone exchange rather than a burn, I don't recall that we had any hard numerical evidence on this. We should pay attention to the walltime of the test suite cases with reactions to see if there's a noticeable difference.

## PR checklist


zingale:
errors are roundoff:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-05-06-001/index.html

Issue #913, opened 2020-05-04
add axisymmetric geometry source terms


zingale:
In Bernard-Champmartin, they give a form of the geometric source terms that arise when writing the axisymmetric equations in conservative form.  These should be added through our regular old/new sources mechanism.  Also... we need to ensure that when run with axisymmetry that the Coriolis force is in a proper righthanded coordinate system (since we store the velocity as r, z, theta in the code, but these new terms are computed with the standard r, theta, z ordering).

zingale:
Here's a derivation:
[axisymmetry_rotation.pdf](https://github.com/AMReX-Astro/Castro/files/4582488/axisymmetry_rotation.pdf)

PR #923, opened 2020-05-06
add axisymmetric geometric sources

zingale:
## PR summary

In axisymmetric geometry, there are additional forces that arise due to the changing direction of the unit vectors in the div{rho U U} term.  The paper by Bernand-Champmartin discusses this.  See issue #913.  Note that the way we handle the Coriolis force for 2-d axisymmetry is already consistent with a righthanded coordinate system even though we order the state as (r, z, theta).

## PR checklist


Source/sources/Castro_geom.cpp
```diff
@@ -0,0 +1,147 @@
...
+      Real r = prob_lo[0] + (static_cast<Real>(i) + 0.5_rt)*dx[0];
+
+      Real rho_half = 0.5_rt*(old_state(i,j,k,URHO) + new_state(i,j,k,URHO));
+
+      // radial momentum: F = rho v_phi**2 / r
+      src(i,j,k,UMX) = 0.25_rt * (old_state(i,j,k,UMZ) + new_state(i,j,k,UMZ)) *
+                                 (old_state(i,j,k,UMZ) + new_state(i,j,k,UMZ)) / (rho_half * r);
+
+      // azimuthal momentum: F = - rho v_r v_phi / r
+      src(i,j,k,UMX) = -0.25_rt * (old_state(i,j,k,UMX) + new_state(i,j,k,UMX)) *
```
harpolea, line 142: I think this should be `src(i,j,k,UMY)`

maxpkatz (approved):

maxpkatz:
Let's go ahead and merge this. I'm still not sure what we should do to make this safe for the case where we simultaneously do rotation with the state data in the inertial frame as Byerly et al. suggest, but that's not the default mode and until now has still mostly been a research idea in Castro.

PR #927, opened 2020-05-07
Rework the Reactions_Type StateData

maxpkatz:
## PR summary

The first NumSpec components of Reactions_Type are rho * omegadot rather than omegadot; then, the NumAux auxiliary components are stored, if the network has any auxiliary variables; then, rho * enuc is stored (enuc itself is removed), and finally the burn weights are stored. The checkpoint version has been incremented.

Some `reactions_max_solve_level` code which was accidentally removed in #922 is restored.

The derives in the flame and toy_flame problems are updated -- it looks like they were not correct before, because we intended to integrate (rho omegadot) before but were only integrating omegadot.

## PR motivation

This cleanup, where the calculation of the burn is separate from the application to the state, will help make it easier to do certain experimentation we'd like to do with the burning, such as doing a higher-order burn, and doing the burn in parallel on the CPU and GPU.

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-05-21-002/index.html
there are some crashes here

maxpkatz:
> there are some crashes here
OK. Can you look at the SDC change at the bottom and tell me what I did wrong? I didn't really get what's going in store_reaction_state.

maxpkatz:
I've updated the simplified SDC burn to be consistent with this PR. Also I fixed the crashes in flame and toy_flame. Ready for retest.

maxpkatz:
Latest tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-05-23/index.html

I am satisfied with these. The derived fields `enuc` and `t_sound_t_enuc` have relatively large diffs in some cases but it was not expected that these should have stayed exactly the same since we are now deriving this based on the change to UEINT.

We'll need a full test suite reset after this since most tests now have a different set of variables in their plotfile.

Source/sdc/sdc_util.cpp
```diff
@@ -415,14 +415,20 @@ void Castro::ca_store_reaction_state(const Box& bx,
...
+#endif
+
     AMREX_PARALLEL_FOR_3D(bx, i, j, k,
     {
-        R_store(i,j,k,NumSpec) = R_old(i,j,k,UEDEN)/state(i,j,k,URHO);
-        R_store(i,j,k,NumSpec+1) = R_old(i,j,k,UEDEN);
-        R_store(i,j,k,NumSpec+2) = 0.0; // we're not storing the weights
+        R_store(i,j,k,NumSpec+NumAux  ) = R_old(i,j,k,UEDEN);
+        R_store(i,j,k,NumSpec+NumAux+1) = 0.0; // we're not storing the weights
     });
```
zingale, line 432: are the weights no longer stored?

maxpkatz: That's not a change, you were already not storing the weights there (#1014).

PR #928, opened 2020-05-07
remove ghost cells from B and extra from state

guadabsb15:
## PR checklist

### DIFF
--- a/Source/driver/Castro_setup.cpp
+++ b/Source/driver/Castro_setup.cpp
@@ -386,7 +386,7 @@ Castro::variableSetUp ()
   bool state_data_extrap = false;
   bool store_in_checkpoint;
 
-#if defined(RADIATION) || defined(MHD)
+#if defined(RADIATION) 
   // Radiation should always have at least one ghost zone.
   int ngrow_state = std::max(1, state_nghost);
 #else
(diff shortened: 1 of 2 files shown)


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-05-07-009/index.html

zingale:
Not directly related to this PR, but I think we are not doing the new thermo source correct.

In `construct_new_thermo_source`, we call `fill_thermo_source` twice, first with the old state and we add -0.5 of that to `thermo_src` and then we call it a second time, and add 0.5 of that to `thermo_src`.  So the idea is that the new source should be of the form 

1/2 (S_new - S_old).

But in the second call, we are doing the time-centering ourselves, by averaging `state_old` and `state_new`.  I think we should only use `state_new`.  We can make this clearer by having `fill_thermo_source()` take only a single state.

guadabsb15:
Thank you for pointing that out. I think it is fixed. I would appreciate if we do the change in fill_thermo_source in another PR, but I'm open to what you prefer. Taking care of the ghost cells, helps with further debugging.

zingale:
okay, I'll open a different PR with that fix

PR #930, opened 2020-05-07
Save integrator stats in the Reactions_Type weights for Simplified SDC

dwillcox:
This PR saves `n_rhs + 2*n_jac` from the integrator into the `Reactions_Type` variable `weights` for simplified SDC as is currently done for Strang.

To see this in the plotfiles requires first merging #929


zingale:
can you add to the CHANGES.md?

dwillcox:
I've updated the changelog for the current PR as well as #929, which I forgot about.

PR #931, opened 2020-05-07
use the hydro ctoprim

zingale:
## PR checklist


zingale:
all tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-05-07-008/index.html

Source/hydro/Make.package
```diff
@@ -1,6 +1,9 @@
 # sources used with hydro
 
-CEXE_sources += Castro_hydro.cpp
+ifneq ($(USE_MHD),TRUE)
```
maxpkatz, line 3: Is there a specific conflict here?

zingale: yes.  When we compile with USE_MHD=TRUE, the ctoprim call expects the B fields, but the hydro uses of it don't have B

PR #942, opened 2020-05-09
another pass at linting

zingale:
## PR checklist


Source/driver/Castro.cpp
```diff
@@ -146,32 +146,33 @@ void
 Castro::variableCleanUp ()
 {
 #ifdef GRAVITY
-  if (gravity != 0) {
-    if (verbose > 1 && ParallelDescriptor::IOProcessor()) {
+  if (gravity != nullptr) {
+    if (verbose >= 1 && ParallelDescriptor::IOProcessor()) {
```
maxpkatz, line 150: I think `verbose > 1` was correct here, this is too much information for the typical user.

Source/driver/Castro.cpp
```diff
@@ -146,32 +146,33 @@ void
...
     delete diffusion;
-    diffusion = 0;
+    diffusion = nullptr;
   }
 #endif
 
 #ifdef RADIATION
-  if (radiation != 0) { int report = (verbose || radiation->verbose);
+  if (radiation != nullptr) {
+    int report = (verbose >= 1 || radiation->verbose);
```
maxpkatz, line 170: Let's also make these both `verbose > 1`

Source/driver/Castro.cpp
```diff
@@ -1353,7 +1371,7 @@ Castro::init ()
     // If we just triggered a regrid, we need to account for the fact that
     // the data on the coarse level has already been advanced.
 
-    if (getLevel(level-1).post_step_regrid)
+    if (getLevel(level-1).post_step_regrid = 1)
```
maxpkatz, line 1374: ==

maxpkatz:
For all the places where this PR does e.g. `if (do_grav)` -> `if (do_grav == 1)`, then to keep the meaning the same I think we should either test on `if (do_grav != 0)` or add a check at initialization that aborts if the value is anything other than 0 or 1.

zingale:
this PR is out of date

PR #947, opened 2020-05-10
an attempt at making the left and right states clearer

zingale:
## PR summary

There are a few issues in how the left and right states are named.  In mhd_plm, `Ip` is used for the left state and `Im` for the right.  These are then stored in `qp` and `qm` in the driver.  But this is opposite the convention that we use in the rest of the code.  To make things clearer, `qp` is now `qleft` and `qm` is now `qright`.

I propagated this naming convention into ct_upwind.  There however, there is more confusion.  We call `corner_couple` as:
```
       call corner_couple(work_lo, work_hi, &                                                                                                  
                           cons_temp_M, q_lo, q_hi, &                                                                                           
                           cons_temp_P, q_lo, q_hi, &                                                                                           
                           um, q_lo, q_hi, &                                                                                                    
                           up, q_lo, q_hi, &                                                                                                    
                           flxy1D, flxy_lo, flxy_hi, &                                                                                          
                           !d1 = x, d2 = y, dir2 = 1 last component in the cons_temp arrays                                                     
                           1 , 2, 1, &                                                                                                          
                           dx(1), dt) !qmpxy                                                                                                    
```                                                                  
Here, `cons_temp_M` and `cons_temp_P` are the right and left new states and `um` and `up` are the right and left input states.  But internal to the `corner_couple` routine, we refer to them as:
```
subroutine corner_couple(w_lo, w_hi, &                                                                                                          
                         uL, ul_lo, ul_hi, &                                                                                                    
                         uR, ur_lo, ur_hi, &                                                                                                    
                         um, um_lo, um_hi, &                                                                                                    
                         up, up_lo, up_hi, &                                                                                                    
                         flxd2, flxd2_lo, flxd2_hi, &                                                                                           
                         d1, d2, dir2, &                                                                                                        
                         dx, dt)                        
```
so what we are calling `uL` is actually the updated *right* state and what we call `uR` is actually the updated *left* state.  This same confusion propagates through 3 other routines.
## PR checklist


guadabsb15:
> There are a few issues in how the left and right states are named. In mhd_plm, `Ip` is used for the left state and `Im` for the right. These are then stored in `qp` and `qm` in the driver. But this is opposite the convention that we use in the rest of the code. To make things clearer, `qp` is now `qleft` and `qm` is now `qright`.
> 

Indeed, I think it would make things clearer changing names as you are suggesting. I believe this names come from the eq.35 in the paper,  which in their notation call the riemann solver as R(U+,U-).

guadabsb15:
> so what we are calling `uL` is actually the updated _right_ state and what we call `uR` is actually the updated _left_ state. This same confusion propagates through 3 other routines.

Yes, you are right, this names uL and uR internal to the subroutines are totally swapped

PR #968, opened 2020-05-16
some more MHD cleaning

zingale:
## PR summary

This simplifies the logic in the corner_couple routine.  I also finished a read through the electric_edge_z and updated some comments.  Finally, we don't need to initialize the interface states to 0 in mhd_plm.

This also removes some old Fortran files that are no longer used.

## PR checklist


guadabsb15:
I think it is also possible to remove ctoprim_mhd from Castro_mhd_F.H

Issue #892, opened 2020-04-30
MHD solver uses different convention for interface state indexing


zingale:
The qL and qR that make up an interface passed to the Riemann solver live at different i,j,k.  We should switch to the convention used in the rest of Castro where, for interface states, i,j,k is the index of the interface.

PR #973, opened 2020-05-18
shift the MHD interface states so left and right are indexed the same

zingale:
## PR summary

Previously for the MHD solver, we were storing q_{i+1/2,L} as qleft(i,j,k) and q_{i-1/2,R} as qright(i,j,k).  This shifts the indexing so (i,j,k) for an interface state always means {i-1/2,j,k}.  This makes the convention consistent with the other hydro solvers.

## PR checklist


zingale:
All tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-05-19-005/index.html

zingale:
this addresses #892

PR #980, opened 2020-05-21
tighten up the boxes for MHD half step

zingale:
Lupe pointed out that for the half step we can use the same box size as
the final flux computation

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-05-21-003/index.html

PR #982, opened 2020-05-21
    we don't need NHYP in Fortran, so just define NUM_GROW in C++

zingale:
## PR checklist

This also increases NUM_GROW to 6 for the MHD algorithm (this is needed to support flattening)


zingale:
MHD passes with this change: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-05-21-004/index.html

guadabsb15:
there is a `BL_ASSERT(NUM_GROW == 4);` in Castro_mhd.cpp so It will give a backtrace when the code is ran in debug mode

zingale:
I just fixed that

Issue #951, opened 2020-05-12
create ptot for MHD, analogous to radiation


zingale:
In radiation, in the primitive variable state we have ptot, which is the sum of the gas and radiation pressures.  I propose creating a ptot for MHD which is the sum of the gas and magnetic pressures.  This will make the flattening simpler, since currently we create a fake "q2" state that is expanded by 1 to include ptot.  With this change, we can easily leverage the flattening routines from hydro.

Issue #891, opened 2020-04-30
MHD flattening


zingale:
The MHD flattening routine should be synced up with what the main hydro does.

zingale:
I think we can just replace the flattening with the hydro version -- it seems to be the same algorithm.

Also, the MHD flattening as written doesn't fill the flattening coefficient in the one ghost cell needed for the slopes, so that needs to be fixed.

zingale:
This paper seems to only use the total pressure (p = p_g + B^2/2) in the flattening procedure:
https://academic.oup.com/mnras/article/368/3/1040/1022149

zingale:
notice that the flattening is not being applied to all the zones that are reconstructed -- we only compute the flattening coefficient in one ghost cell.  I think we need 2, and that means that we would need 5 ghost cells total, instead of the 4 we use for hydro.

Also note that the current behavior means that we will get different answers with different domain decompositions.

zingale:
this is related to #951

PR #984, opened 2020-05-22
Port MHD flattening to C++ and address box sizes

zingale:
## PR summary

Now the flattening is done over the same size box as we do reconstruction on.  Previously we were not computing the flattening coefficient in all the zones where we applied it.  This also reuses the C++ flattening routine.

Note: this adds QPTOT to the primitive variable state so we can flattening on that.

## PR checklist


zingale:
This closes #951 and #891

PR #996, opened 2020-05-25
port the MOL consup and the 2nd order diffusion to C++

zingale:
## PR checklist


zingale:
all tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-05-25-007/index.html

maxpkatz:
> this removes the explicit addition of the hybrid source to the sources in consup
since I believe it was already accounted for in do_old_sources, and therefore we
were double counting.

`add_hybrid_advection_source` is not the same thing as the hybrid source term that gets handled through the standard old sources. It is a pressure gradient term that we compute in the hydro because that is when we have an edge-based pressure to work with. It is equivalent to this code in Castro_ctu_hydro.cpp:

https://github.com/AMReX-Astro/Castro/blob/master/Source/hydro/Castro_ctu_hydro.cpp#L1225-L1244

zingale:
I've fixed the hybrid source update to match what we do in CTU

PR #1010, opened 2020-05-30
update the docs on boundary conditions

### LINKED ISSUE #850: need to document BCs
In addition to the standard BCs, we have ambient and HSE.  These need some documentaiton.

zingale:
## PR summary

closes #850

## PR checklist


maxpkatz:
Perhaps we should be even more specific about what `hse_interp_temp` and `hse_reflect_vels` actually do.

Since we say they're essential for plane-parallel, is there a reason you wouldn't use them when the HSE BC is on? Should we just make them default and/or remove the alternatives?

zingale:
actually, making them the default is probably a good idea.

I meant it to say that HSE BCs in general are important for plane-parallel

zingale:
I believe I've addressed these comments

PR #1018, opened 2020-06-06
Add fluxes consistently between simplified SDC and CTU

maxpkatz:
## PR summary

At the end of the CTU hydro loop, we store the local Fab data from the fluxes and pressure into fluxes, mass_fluxes, and P_radial. This is done as an add because in general we subcycle the CTU advance, and we want the fluxes at the end of the timestep to be the sum of the fluxes from all subcycles. Then this data is reused later in the reflux.

Previously, for simplified SDC, we were copying the data instead of adding it, because we didn't want to add for each simplified SDC iteration, and when retry subcycling was added, it didn't work for simplified SDC. However we can now retry with simplified SDC, so we make it consistent with the CTU advance by adding fluxes as well, but only on the last SDC iteration.

In order to do this we modify Castro_advance_ctu.cpp slightly so that `sdc_iters` is directly modified on a retry (so that we can read its current value in the hydro advance).

We also correct the fact that P_radial was being stored inconsistently; we stored it always for non-Cartesian geometries, which was different from the logic we used for calculating the local pradial_fab update.

## PR checklist

### DIFF
--- a/Source/hydro/Castro_ctu_hydro.cpp
+++ b/Source/hydro/Castro_ctu_hydro.cpp
@@ -1287,10 +1287,6 @@ Castro::construct_ctu_hydro_source(Real time, Real dt)
       nstep_fsp = std::max(nstep_fsp, priv_nstep_fsp);
 #endif
 
-#if AMREX_SPACEDIM <= 2
-      Array4<Real> pradial_fab = pradial.array();
-#endif
-
 
       for (int idir = 0; idir < AMREX_SPACEDIM; ++idir) {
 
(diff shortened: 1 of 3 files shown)


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-06-07/index.html

zingale:
can you update CHANGES.md?

Source/hydro/Castro_ctu_hydro.cpp
```diff
@@ -1311,110 +1307,95 @@ Castro::construct_ctu_hydro_source(Real time, Real dt)
...
-        // we want to copy the fluxes since we expect that there will not be
-        // subcycling and we only want the last iteration's fluxes.
+        bool add_fluxes = true;
 
-        Array4<Real> const flux_fab = (flux[idir]).array();
-        Array4<Real> fluxes_fab = (*fluxes[idir]).array(mfi);
-        const int numcomp = NUM_STATE;
+        if (time_integration_method == SimplifiedSpectralDeferredCorrections &&
+            sdc_iteration == sdc_iters - 1) {
+            add_fluxes = false;
```
zingale, line 1339: should this be `sdc_iteration != sdc_iters - 1`?

maxpkatz: Yes, fixed.

maxpkatz:
> can you update CHANGES.md?

Done

PR #1023, opened 2020-06-07
Move simplified SDC burn into CTU advance

maxpkatz:
## PR summary

Since the simplified SDC burn was happening right after do_advance_ctu, this change simplifies the logic so it happens just inside instead.

## PR checklist


zingale:
test results: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-06-08-002/index.html

PR #1028, opened 2020-06-09
convert the slope routine to a header implementation

zingale:
## PR summary

Now uslope and pslope are header implementations that work zone by zone.  This makes it easier to share them (e.g., with MHD) and saves a lot of memory, since we don't need to create a FAB for the slopes.

To properly test this, I need to add some more PLM problems to the suite, including some that exercise the well-balanced reconstruction.

## PR checklist


zingale:
all PLM tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-06-10-001/index.html

maxpkatz:
Do you think test coverage is sufficient here? I'm particularly concerned about the boundary conditions -- do we have one exercising the symmetric boundary functionality?

zingale:
I explicitly added the hse_convergence tests to test the reflecting BCs -- this is with SDC + PLM.

PR #1033, opened 2020-06-11
Do ctoprim on a fab-by-fab basis

### LINKED ISSUE #908: compute q and q_aux as FABs in the main MFIter loop
Instead of having these as MFs, we should compute them on a box-by-box basis as needed in the MFIter loop.  Note: this is essentially what MHD does now.

maxpkatz:
## PR summary

Closes #908.

## PR checklist

### DIFF
--- a/Source/driver/Castro.H
+++ b/Source/driver/Castro.H
@@ -1261,7 +1261,9 @@ protected:
 ///
 /// The primitive variable state array.
 ///
+#ifdef TRUE_SDC
     amrex::MultiFab q;
+#endif
 
 
 ///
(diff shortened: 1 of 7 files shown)


zingale:
I don't see a performance issue on the CPU.  Here are the tests (note we need to fix MHD compilation):

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-06-11-001/index.html

zingale:
Here are the timings on GPU: http://groot.astro.sunysb.edu/Castro/test-suite/gpu/2020-06-11-001/index.html

maxpkatz:
The GPU timings on groot are slightly slower, but it's not a representative test of GPU performance in a production run. I checked on a V100 on a problem with larger boxes and there is no slowdown from this change.

PR #1038, opened 2020-06-13
Write a wrapper for print_update_diagnostics

maxpkatz:
## PR summary

Code redundancy is reduced because a few places that were repeating logic for how to sum up and print the contributions to the state are combined into a single wrapper function.

The diagnostics for the CTU hydro source are moved into the CTU advance rather than being called directly in the source constructor.

## PR checklist


maxpkatz:
All CPU tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-06-13/index.html

PR #1047, opened 2020-06-17
allow convergence testing in 3d + plm/ppm

zingale:
## PR checklist

### DIFF
--- a/Exec/hydro_tests/acoustic_pulse/convergence_plm.sh
+++ b/Exec/hydro_tests/acoustic_pulse/convergence_plm.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# echo the commands
+set -x
+
+DIM=3
+EXEC=./Castro${DIM}d.gnu.MPI.ex
+
+RUNPARAMS="castro.ppm_type=0 castro.time_integration_method=0"
+
+mpiexec -n 8 ${EXEC} inputs.3d.64 ${RUNPARAMS} amr.plot_file=acoustic_pulse_64_plm_plt &> 64.out
+mpiexec -n 8 ${EXEC} inputs.3d.128 ${RUNPARAMS} amr.plot_file=acoustic_pulse_128_plm_plt &> 128.out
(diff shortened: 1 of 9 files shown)


Exec/hydro_tests/acoustic_pulse/convergence_ppm.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# echo the commands
+set -x
+
+DIM=3
+EXEC=./Castro${DIM}d.gnu.MPI.ex
+
+RUNPARAMS="castro.ppm_type=1 castro.time_integration_method=0"
+
+mpiexec -n 8 ${EXEC} inputs.3d.64 ${RUNPARAMS} amr.plot_file=acoustic_pulse_64_ppm_plt &> 64.out
```
guadabsb15, line 11: should it be inputs.64 instead of inputs.3d.64 , the same for the other input names on this file

Exec/hydro_tests/acoustic_pulse/convergence_plm.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# echo the commands
+set -x
+
+DIM=3
+EXEC=./Castro${DIM}d.gnu.MPI.ex
+
+RUNPARAMS="castro.ppm_type=0 castro.time_integration_method=0"
+
+mpiexec -n 8 ${EXEC} inputs.3d.64 ${RUNPARAMS} amr.plot_file=acoustic_pulse_64_plm_plt &> 64.out
```
guadabsb15, line 11: similar to the convergence_ppm, maybe here should be should be inputs.64 instead of inputs.3d.64 ...etc.

Exec/hydro_tests/acoustic_pulse/convergence_ppm.sh
zingale: yes, you are right.  I renamed them after I wrote the script.  I'll fix it now.

PR #1051, opened 2020-06-21
Dimension agnostic ca_compute_avgpres

maxpkatz:
## PR checklist


maxpkatz:
Note: we don't have any tests of GR gravity in the suite, so I can't easily verify correctness. But we should at least run the gravity tests in the suite to make sure they still compile.

PR #1070, opened 2020-06-27
checking if divergence of B is 0 at initialization

guadabsb15:
## PR summary
Adding a check during the initialization, for the divergence of magnetic field to be 0 

## PR checklist

### DIFF
--- a/Source/driver/Castro.cpp
+++ b/Source/driver/Castro.cpp
@@ -995,7 +995,7 @@ Castro::initData ()
       add_magnetic_e(Bx_new, By_new, Bz_new, S_new);
       
       //check divB
-      check_div_B(Bx_new, By_new, Bz_new, cur_time, S_new);    
+      check_div_B(Bx_new, By_new, Bz_new, S_new);    
 
 #endif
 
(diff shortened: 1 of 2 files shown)


zingale:
I don't think you need ghost cells, I think you should create a cell-centered div B just like we do for the derived variables, e.g.,
```
divB(i,,j,k) = (Bx(i+1,j,k) - Bx(i,j,k))/dx + (By(i,j+1,k) - By(i,j,k))/dy + (Bz(i,j,k+1) - Bz(i,j,k))/dz
```

also, I think you need to normalize the check by the magnitude of B, since roundoff will be relatively to the size of the field.

guadabsb15:
> I don't think you need ghost cells, I think you should create a cell-centered div B just like we do for the derived variables, e.g.,
> 
> ```
> divB(i,,j,k) = (Bx(i+1,j,k) - Bx(i,j,k))/dx + (By(i,j+1,k) - By(i,j,k))/dy + (Bz(i,j,k+1) - Bz(i,j,k))/dz
> ```
> 
> also, I think you need to normalize the check by the magnitude of B, since roundoff will be relatively to the size of the field.

I think I am doing that now

Source/driver/Castro.cpp
```diff
@@ -3485,6 +3489,67 @@ Castro::add_magnetic_e( MultiFab& Bx,
...
+      {
+          
+          Real divB = (Bx_arr(i+1,j,k) - Bx_arr(i,j,k))/dx[0] +
+                      (By_arr(i,j+1,k) - By_arr(i,j,k))/dx[1] + 
+                      (Bz_arr(i,j,k+1) - Bz_arr(i,j,k))/dx[2];
+
+          Real magB = std::sqrt(Bx_arr(i,j,k) * Bx_arr(i,j,k) + 
+                                By_arr(i,j,k) * By_arr(i,j,k) +
+                                Bz_arr(i,j,k) * Bz_arr(i,j,k));
+
```
zingale, line 3531: I think we want magB to be cell-centered, so you should average the face centered B to centers and then compute mag B of that

Source/driver/Castro.cpp
```diff
@@ -3485,6 +3489,67 @@ Castro::add_magnetic_e( MultiFab& Bx,
...
+          if (std::abs(divB) > 1.0e-10*magB){
+             fail_divB = 1; 
+          }
+          
+
+          return {fail_divB}; 
+      });
+
+      ReduceTuple hv = reduce_data.value();
```
maxpkatz, line 3543: This should go outside the MFIter loop

Source/driver/Castro.cpp
guadabsb15: true, I just changed it

Source/driver/Castro.cpp
guadabsb15: thanks for pointing that out

PR #1072, opened 2020-06-27
change the preprocessor check from naux to NAUX_NET

zingale:
it is not clear that the naux check every worked (it didn't in
Microphysics), but now the network explicitly sets NAUX_NET

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-06-27-003/index.html

PR #1078, opened 2020-06-29
Always use previous phi as guess for solve

maxpkatz:
## PR summary

The Poisson solve has an option to not use the previous gravitational potential as a guess for the new solve (instead, a monopole solve is done and used to fill the domain). This is only used after initialization or restart. This option is now turned off, so that we always use the previous phi. At initialization we will pay a (one-time) larger cost to do the solve, but this should have no other effects.

## PR motivation

I would like to remove make_radial_phi; for the most part we don't need it and it makes monopole gravity harder to maintain. If this bit is removed then the only remaining use of make_radial_phi is in the BCs for the Poisson solve, and there it can be replaced entirely with the multipole BCs.

## PR checklist

### DIFF
--- a/Source/driver/Castro.cpp
+++ b/Source/driver/Castro.cpp
@@ -2013,13 +2013,11 @@ Castro::post_restart ()
             {
                 if (gravity->NoComposite() != 1)
                 {
-                   int use_previous_phi = 1;
-
                    // Update the maximum density, used in setting the solver tolerance.
 
                    gravity->update_max_rhs();
 
-                   gravity->multilevel_solve_for_new_phi(0,parent->finestLevel(),use_previous_phi);
+                   gravity->multilevel_solve_for_new_phi(0, parent->finestLevel());
                    if (gravity->test_results_of_solves() == 1)
                        gravity->test_composite_phi(level);
                 }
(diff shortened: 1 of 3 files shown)


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-06-28-003/index.html

maxpkatz:
> 
> 
> tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-06-28-003/index.html

Some of those failures are from the branch not being even with development, that's now resolved

maxpkatz:
Updated tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-06-29-002/index.html

This is consistent with my expectation. The maximum diffs are consistent with the tolerance in the Poisson gravity solve, and the time taken is about the same. This is ready for review.

zingale:
latest tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-06-29-002/index.html

zingale:
The Poisson solve has an option to not use the previous gravitational potential as a guess for the new solve (instead, a monopole solve is done and used to fill the domain). This is only used after initialization or restart. This option is now turned off, so that we always use the previous phi. At initialization we will pay a (one-time) larger cost to do the solve, but this should have no other effects.

I would like to remove make_radial_phi; for the most part we don't need it and it makes monopole gravity harder to maintain. If this bit is removed then the only remaining use of make_radial_phi is in the BCs for the Poisson solve, and there it can be replaced entirely with the multipole BCs.

PR #1083, opened 2020-06-29
Switch some more references from master to main

maxpkatz:
## PR checklist


harpolea:
Since the switch to the github action, `deploy_docs.sh` isn't used anymore, so we should probably just delete it?

maxpkatz:
> 
> 
> Since the switch to the github action, `deploy_docs.sh` isn't used anymore, so we should probably just delete it?

Done

PR #1085, opened 2020-06-30
Convert interpolate_monopole_grav to C++

maxpkatz:
## PR checklist


Source/gravity/Gravity.cpp
```diff
@@ -1377,21 +1377,109 @@ Gravity::interpolate_monopole_grav(int level, RealVector& radial_grav, MultiFab&
...
-                          n1d, level);
+        const Box& bx = mfi.growntilebox();
+
+        auto grav = grav_vector.array(mfi);
+        Real* const radial_grav_ptr = radial_grav.dataPtr();
+
+        // Note that we are interpolating onto the entire range of grav,
+        // including the ghost cells. Taking the absolute value of r ensures
+        // that we will get the correct behavior even for the ghost zones with
+        // negative indices, which have a reflecting boundary condition.
```
zingale, line 1401: I don't see where the abs() is taken

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-06-30-001/index.html

Source/gravity/Gravity.cpp
maxpkatz: This was a slightly out of date comment that comes from when I converted the 1D/2D/3D implementation to the dimension-agnostic form earlier this month. We do indirectly ensure that `r` is positive by computing sqrt(loc(:)**2), but in 1D this was just an `abs()`. I've removed the sentence just to avoid the confusion.

PR #1086, opened 2020-06-30
Convert put_direct_sum_bc to C++

maxpkatz:
## PR summary

This also fixes what looks like a bug in the indexing, which wouldn't have been detectable on the uniform_cube_sphere problem due to the symmetry of the cube.

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-06-30-002/index.html

PR #1093, opened 2020-07-02
make problem parameters available to C++

### LINKED ISSUE #1065: Auto prob params should be available in C++
The _prob_params variables that get generated in Fortran should have corresponding C++ variables generated as well. There should also be autogenerated get/set routines so that we can access them from C++. For now the ground truth copy of the variables would continue to live in Fortran, but this will ease the porting process to C++ for the problem setups.

zingale:
## PR summary

For problems that are using `USE_PROB_PARAMS = TRUE`, this now makes all of the problem runtime parameters
available in C++ in addition to Fortran.  This applies only to the problem runtime parameters that are controlled by the namelist, and not any other variables that we're put in _prob_params just to be included in the Fortran module.

The way this works is similar to how we do the extern runtime parameters -- the reading is still controlled by Fortran via the namelist, but now the `write_probdata.py` script also writes a "get" routine for each of the Fortran parameters that is accessible from C++.  These are then all called by `init_prob_parameters()` just before we do the initialization of the state.

This closes #1065.  This depends on https://github.com/starkiller-astro/Microphysics/pull/336

## PR checklist


Source/driver/Castro.cpp
```diff
@@ -848,6 +851,11 @@ Castro::initData ()
 {
     BL_PROFILE("Castro::initData()");
 
+    // initialize the C++ values of the runtime parameters
+#ifdef PROB_PARAMS
+    init_prob_parameters();
```
maxpkatz, line 856: `initData()` is called on each level, and is not called after a restart -- can this be put in `variableSetup()` instead?

zingale: yes, thanks.  I wasn't sure the best place to put it.

zingale: okay, I moved it to the Castro constructor  and verified that the data is properly initialized

PR #1098, opened 2020-07-03
B field derives need a grown box

maxpkatz:
## PR summary

The derived variables depending on gradients of B need to have sources defined on a box that is grown by one ghost zone so that the gradients can be correctly calculated at box boundaries.

## PR checklist

### DIFF
--- a/Source/driver/Castro_setup.cpp
+++ b/Source/driver/Castro_setup.cpp
@@ -1012,13 +1012,13 @@ Castro::variableSetUp ()
 //
 //Magentic Field Cell Centered
 //x component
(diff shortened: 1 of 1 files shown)


maxpkatz:
Turns out this doesn't really work. There's a fundamental issue with the E/B derives which is that they mix cell centered and nodal data, or nodal data in one direction with nodal data in another direction, and the derive infrastructure is not set up to handle this. We need to solve this more generally either with a new derive interface in AMReX, or by overriding the AmrLevel derive interface specifically for the E and B field derives.

PR #1103, opened 2020-07-04
Remove make_radial_phi

maxpkatz:
## PR summary

This no longer has any effect.

Also, some obsolete interfaces to Fortran are removed.

## PR checklist


maxpkatz:
This removes the last #pragma gpu instance in gravity.

Source/gravity/Gravity.cpp
```diff
@@ -574,27 +574,12 @@ Gravity::gravity_sync (int crse_level, int fine_level, const Vector<MultiFab*>&
...
+#elif (BL_SPACEDIM == 2)
+      fill_multipole_BCs(crse_level,fine_level,amrex::GetVecOfPtrs(rhs),*delta_phi[crse_level]);
 #else
-      if (gravity::lnum >= 0) {
-          fill_multipole_BCs(crse_level,fine_level,amrex::GetVecOfPtrs(rhs),*delta_phi[crse_level]);
-      } else {
-          int fill_interior = 0;
-          make_radial_phi(crse_level,*rhs[0],*delta_phi[crse_level],fill_interior);
-      }
+      fill_multipole_BCs(crse_level,fine_level,amrex::GetVecOfPtrs(rhs),*delta_phi[crse_level]);
```
zingale, line 582: what's the difference in this if?

Source/gravity/Gravity.cpp
```diff
@@ -3541,27 +3407,12 @@ Gravity::solve_phi_with_mlmg (int crse_level, int fine_level,
...
+#elif (BL_SPACEDIM == 2)
+        fill_multipole_BCs(crse_level, fine_level, rhs, *phi[0]);
 #else
-        if (gravity::lnum >= 0) {
-            fill_multipole_BCs(crse_level, fine_level, rhs, *phi[0]);
-        } else {
-            int fill_interior = 0;
-            make_radial_phi(crse_level, *rhs[0], *phi[0], fill_interior);
-        }
+        fill_multipole_BCs(crse_level, fine_level, rhs, *phi[0]);
```
zingale, line 3415: this if is unnecessary

Source/gravity/Gravity.cpp
maxpkatz: Nothing. It was already there, I just didn't remove it. I was going to clean that up later -- my next plan is to make the direct sum BCs work in <3D, and then we don't need any of this logic.

PR #1112, opened 2020-07-06
Ensure mass fluxes are correct in reflux after a retry

### LINKED ISSUE #1020: mass_fluxes is incorrect after a retry when using AMR
The flux-based gravitational source term uses the `mass_fluxes` array. After a reflux, we add the reflux contribution to `mass_fluxes` and then re-evaluate all source terms, including gravity. But in the CTU advance `mass_fluxes` is always a copy of the latest fluxes, not the sum over all subcycles, because it has to be that way if the evaluation of the gravity source term within that particular subcycle is to be correct. So the update_source_after_reflux calculation will only be using the last subcycle's worth of mass fluxes, not the full timestep.

maxpkatz:
## PR summary

As described in #1020, if a retry occurred during a timestep then the mass_fluxes data would not have been correct during the reflux operation. This is now fixed so that the mass fluxes just copy the URHO component of the fluxes.

Fixes #1020 

## PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # 20.08
 
+   * When using AMR and gravity or rotation, the source terms applied after
+     a reflux would have been incorrect if the previous timestep had a retry
+     (#1020). This has now been fixed. (#1112)
+
    * We now have the ability to access the problem-specific runtime
      parameters in C++ (#1093)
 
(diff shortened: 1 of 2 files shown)


zingale:
this makes a big diff for the 2 wdmerger retry problems (as expected):
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-07-06-004/index.html

PR #1122, opened 2020-07-09
Copy additional data from old Castro level on regrid

maxpkatz:
## PR summary

On a regrid where we're copying data from an existing level, we need to make sure we copy the class data relevant to post_timestep if a retry has occurred. This change copies some of these scalar parameters, and also fillpatches the old data in addition to the new data (if the old data is available in the state we're copying from).

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-07-09-002/index.html

zingale:
note: this PR was not up to date with development, hence a few failures.

Issue #1127, opened 2020-07-11
PhiRot_Type and Rotation_Type should be removed


maxpkatz:
The rotational phi and acceleration state types were originally added for two reasons: (1) it was useful to have a stateful version of the acceleration in the scenario where the rotation period is changing with time; (2) the energy update for rot_source_type == 4 was written based on the rotational "potential" to be symmetric with gravity. (1) is no longer an issue because we removed the ability for omega to change in time, and we don't plan to bring it back. (2) is not required; we can recast the conservative energy update entirely in terms of the acceleration, as we have done for gravity. So we should remove these StateData types to save on memory.

zingale:
I think we are clear to do this now

maxpkatz:
Rotation_Type has already been removed. PhiRot_Type (which is just a scalar) still exists, and will have to be removed from a couple places for this to occur. The first is in the SCF code, the second is in wdmerger's derived quantities.

zingale:
to remove it in the derived quantities, should we just recompute it locally as needed?

maxpkatz:
Yes, that could be done.

zingale:
this is implemented in #1974

PR #1128, opened 2020-07-11
Calculate rotational acceleration inline and remove Rotation_Type

maxpkatz:
## PR summary

Rather than calculate rotational acceleration first and then use it in the rotation sources, we calculate it on demand in the source evaluation. Rotation_Type is removed from StateData.

## PR motivation

This gets us part of the way to completing #1127.

## PR checklist

### DIFF
--- a/Exec/science/wdmerger/Problem_Derives.H
+++ b/Exec/science/wdmerger/Problem_Derives.H
@@ -84,7 +84,7 @@
     //
     derive_lst.add("rho_phiRot",IndexType::TheCellType(),1,ca_derrhophiRot,the_same_box);
     derive_lst.addComponent("rho_phiRot",desc_lst,State_Type,URHO,1);
-    derive_lst.addComponent("rho_phiRot",desc_lst,Rotation_Type,0,1);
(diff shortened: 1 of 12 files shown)


Exec/science/wdmerger/Problem_Derives.H
```diff
@@ -84,7 +84,7 @@
     //
     derive_lst.add("rho_phiRot",IndexType::TheCellType(),1,ca_derrhophiRot,the_same_box);
     derive_lst.addComponent("rho_phiRot",desc_lst,State_Type,URHO,1);
-    derive_lst.addComponent("rho_phiRot",desc_lst,Rotation_Type,0,1);
+    derive_lst.addComponent("rho_phiRot",desc_lst,PhiRot_Type,0,1);
```
maxpkatz, line 87: Note: this line looks it was a bug in the old code, so I fixed it. If PhiRot_Type is removed, we'll have to deal with its use here in the wdmerger derive routines.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/

look good

PR #1130, opened 2020-07-11
Implement conservative rotations on edges like gravity

maxpkatz:
## PR summary

Rather than referring to the rotational potential we can consider only the rotational acceleration (-grad(phi), the equivalent of the gravitational acceleration) on edges to implement the conservative rotation source type.

## PR motivation

This eliminates one source of requiring the rotational potential as StateData (#1127).

## PR checklist


maxpkatz:
The reason this changes the answer to more than roundoff level is that the rotational potential is nonlinear in *x*. So it's not true that if we take a centered difference between two adjacent cells that we'll get the same answer that we would by directly evaluating the centrifugal force at that point.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-07-11-005/index.html

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-09-02-001/index.html

Source/rotation/rotation_sources.cpp
```diff
@@ -447,41 +457,51 @@ Castro::corrrsrc(const Box& bx,
...
+              int ke = dir == 2 ? k + edge : k;
+
+              position(ie, je, ke, geomdata, loc, ccx, ccy, ccz);
+
+              for (int n = 0; n < AMREX_SPACEDIM; ++n) {
+                  loc[n] -= problem::center[n];
+              }
+
+              GpuArray<Real, 3> temp_vel{};
+              Real temp_Sr[3];
```
zingale, line 489: should we initialize this to 0 just to be safe?

zingale: oh, nevermind, that does by default

zingale (approved):

PR #1134, opened 2020-07-17
Move hydro device functions to headers

maxpkatz:
## PR summary

Due to AMReX-Codes/amrex#1148, accesses to class members and functions in device code will fail to compile with CUDA 11. So we need to put these in headers.

## PR checklist


zingale:
all tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-07-17-002/index.html

PR #1139, opened 2020-07-17
if we are doing NSE, then abar was getting added to the derives twice

zingale:
## PR checklist


maxpkatz:
Where is the other abar derive coming from?

zingale:
when we do NSE (using aprox19_nse, e.g.), then there are 3 aux variables: ye, abar, and bea.  So the conserved state carries rho*abar, and we derive abar from that.

maxpkatz (approved):

PR #1145, opened 2020-07-24
Compiler warnings 2

zingale:
## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-07-26-002/index.html

Source/driver/Castro.cpp
```diff
@@ -2345,8 +2345,8 @@ Castro::post_init (Real stop_time)
 
         if (sum_per > 0.0) {
 
-          const int num_per_old = floor((cumtime - dtlev) / sum_per);
-          const int num_per_new = floor((cumtime        ) / sum_per);
+          const int num_per_old = static_cast<int>(std::floor((cumtime - dtlev) / sum_per));
+          const int num_per_new = static_cast<int>(floor((cumtime        ) / sum_per));
```
maxpkatz, line 2349: These should either both be std::floor or neither, right?

Source/driver/Castro.cpp
```diff
@@ -1820,9 +1820,9 @@ Castro::computeNewDt (int                   finest_level,
 
 void
 Castro::computeInitialDt (int                   finest_level,
-                          int                   sub_cycle,
+                          int                   /*subcycle*/,
                           Vector<int>&           n_cycle,
-                          const Vector<IntVect>& ref_ratio,
+                          const amrex::Vector<amrex::IntVect>& /*ref_ratio*/,
```
maxpkatz, line 1825: The compiler complained about Vector in this line but not in the previous line?

zingale: Oh, I probably misunderstood the issue the first time around

zingale: I removed the amrex::

Source/driver/Castro.cpp
zingale: I added the std::

Source/driver/Castro.H
```diff
@@ -628,33 +595,33 @@ public:
...
 /// @param dt_min
 /// @param dt_level
 /// @param stop_time        End time of simulation
 /// @param post_regrid_flag Have we done regridding yet?
 ///
     void computeNewDt (int                   finest_level,
-                       int                   sub_cycle,
+                       int                   /*subcycle*/,
                        amrex::Vector<int>&           n_cycle,
                        const amrex::Vector<amrex::IntVect>& ref_ratio,
```
harpolea, line 626:
```suggestion
                       const amrex::Vector<amrex::IntVect>& /*ref_ratio*/,
```
The docstring says that this is unused

zingale: done

PR #1150, opened 2020-07-28
move inactive developers into "former developers"

zingale:
## PR summary

We have not updated the developers page in some time.  Our policy in the `README.md` states:

> If a core developer is inactive for 3 years, we may reassess their status as a core developer.

None of these developers were active since 2014 (some don't appear in the git history at all).  

This additionally adds Jean Sexton to the list of developers as a new contributor.

## PR checklist


cmsquared:
This seems reasonable to me.

Issue #1161, opened 2020-08-13
Cannot compile following instructions in the tutorial


emprice:
This looks like a great piece of code, and I'd love to give it a try! Unfortunately, my attempts to compile the Sedov hydro test following instructions at [this link](https://amrex-astro.github.io/Castro/docs/getting_started.html) have failed. I get the following output when I run `make` after editing the `GNUMakefile`.

```
Loading ../../../external/amrex/Tools/GNUMake/comps/gnu.mak...
Loading ../../../external/amrex/Tools/GNUMake/sites/Make.unknown...
../../../external/amrex/Tools/GNUMake/Make.rules:474: tmp_build_dir/d/2d.gnu.MPI.EXE/f90.depends: No such file or directory
make: *** No rule to make target `actual_network.F90', needed by `tmp_build_dir/d/2d.gnu.MPI.EXE/f90.depends'.  Stop.
```

I'm not sure what I'm doing wrong here. I am running on a CentOS Unix system and have all the prerequisites I think I need to run this test (and anyway, the error looks like some file is missing or not being written, not a missing dependency). I would greatly appreciate any help!

zingale:
This source file is automatically created at build time, so I suspect the issue is that the python script that builds it failed.

What version of python are you using?

emprice:
I'm running Python 3.7.6 in a virtualenv.

emprice:
Okay, I was able to solve this in a weird way. At first, after you asked about Python, I thought maybe I was missing a Python dependency. But I was able to `import dep` (from the build docs, that's the script that generates the missing file) without error. I was stumped, but on a whim, I tried updating GNU make. Now, I had version 3.81, which is supposed to work, but I thought it couldn't hurt to try something newer, if only to get better debug info. I grabbed the latest release (4.3), built it, and it ran completely without error and built the test! I'm not sure *why* that should have helped, but maybe it will help someone else in the future. Version strings:

```
$ make -v
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for x86_64-redhat-linux-gnu
```

```
$ /path/redacted/make-4.3-gcc/bin/make -v
GNU Make 4.3
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2020 Free Software Foundation, Inc.
[...]

PR #1162, opened 2020-08-13
bump up required make version

zingale:
we don't compile with 3.81 (but that version is from 2006)

## PR checklist


zingale:
this was uncovered in #1161

Issue #1170, opened 2020-08-20
MHD RT has erroneus arithmetic error


zingale:
Building the MHD RT problem and running with `amrex.fpe_trap_invalid=1` gives:

```
 0: ./Castro3d.gnu.DEBUG.ex() [0x79a708]
    amrex::BLBackTrace::print_backtrace_info(_IO_FILE*)
/home/zingale/development/amrex//Src/Base/AMReX_BLBackTrace.cpp:170:29

 1: ./Castro3d.gnu.DEBUG.ex() [0x79a2e0]
    amrex::BLBackTrace::handler(int)
/home/zingale/development/amrex//Src/Base/AMReX_BLBackTrace.cpp:80:8

 2: /lib64/libc.so.6(+0x3c6b0) [0x7f8b20e346b0]
    __restore_rt
??:0

 3: ./Castro3d.gnu.DEBUG.ex() [0x4ce375]
    Castro::ctoprim(amrex::Box const&, double, amrex::Array4<double const> const&, amrex::Array4<double const> const&, amrex::Array4<double const> const&, amrex::Array4<double const> const&,
 amrex::Array4<double> const&, amrex::Array4<double> const&)::{lambda(int, int, int)#1}::operator()(int, int, int) const
../../../Source/hydro/advection_util.cpp:66:25

 4: ./Castro3d.gnu.DEBUG.ex() [0x4e1da9]
    ParallelFor<Castro::ctoprim(const amrex::Box&, amrex::Real, const amrex::Array4<double const>&, const amrex::Array4<double const>&, const amrex::Array4<double const>&, const amrex::Array
4<double const>&, const amrex::Array4<double>&, const amrex::Array4<double>&)::<lambda(int, int, int)> >
/home/zingale/development/amrex//Src/Base/AMReX_GpuLaunchFunctsC.H:49:5

 5: ./Castro3d.gnu.DEBUG.ex() [0x4d2fb1]
    Castro::ctoprim(amrex::Box const&, double, amrex::Array4<double const> const&, amrex::Array4<double const> const&, amrex::Array4<double const> const&, amrex::Array4<double const> const&,
 amrex::Array4<double> const&, amrex::Array4<double> const&)
../../../Source/hydro/advection_util.cpp:211:1

 6: ./Castro3d.gnu.DEBUG.ex() [0x547936]
    Castro::just_the_mhd(double, double)
../../../Source/mhd/Castro_mhd.cpp:171:18

[...]

PR #1171, opened 2020-08-20
fix the size of the domain

zingale:
we can have the domain width be < number of ghost cells

## PR checklist


zingale:
this fixes #1170

PR #1178, opened 2020-08-26
Mhd detonation

Ishita-S:
## PR summary

This is the MHD Detonation problem. 

## PR checklist


zingale:
likewise, you can remove all of the SDC inputs/probins

Exec/hydro_tests/Sedov/GNUmakefile
```diff
@@ -3,11 +3,11 @@ PROFILE    = FALSE
 
 DEBUG      = FALSE
 
-DIM        = 3
+DIM        = 2
```
zingale, line 7: can you revert these changes?

Exec/hydro_tests/Sedov/GNUmakefile
```diff
@@ -3,11 +3,11 @@ PROFILE    = FALSE
...
 
 COMP	   = gnu
 
-USE_MPI    = TRUE
+USE_MPI    = FALSE
 USE_OMP    = FALSE
```
zingale, line 11: this one too

Exec/mhd_tests/Detonation/Problem.H
```diff
@@ -0,0 +1,32 @@
+
```
zingale, line 1: I don't think we need this file -- this was for the collisions stuff

Exec/mhd_tests/Detonation/Problem.f90
```diff
@@ -0,0 +1,91 @@
+! problem-specific Fortran stuff goes here
```
zingale, line 1: I don't think we need this file either

Exec/mhd_tests/Detonation/Problem_F.H
```diff
@@ -0,0 +1,22 @@
+#include <AMReX_BLFort.H>
+
```
zingale, line 2: this also can go

Exec/mhd_tests/Detonation/bc_fill_nd.F90
```diff
@@ -0,0 +1,125 @@
+module bc_fill_module
```
zingale, line 1: I think we shouldn't do special boundary conditions, because I don't think these handle the magnetic fields.  So let's just use the normal BCs to start.

Exec/mhd_tests/Detonation/ext_src_nd.F90
```diff
@@ -0,0 +1,50 @@
+
```
zingale, line 1: this is just a stub and can be removed

zingale:
I am not sure why the submodules are out of sync, maybe a git pull is needed there (for the `external/*`)?

harpolea:
I think this may be because there was some conflict and we had to explicitly checkout the development branches of the submodules for things to work?

zingale:
okay, once we tag Castro 20.09, I'll be updating the submodules, so then maybe this will just sort itself out.  Just waiting on AMReX.

zingale:
we're switching to a new branch

PR #1186, opened 2020-09-03
move problem derives over to C++

zingale:
## PR checklist


zingale:
update: this compiles now, but it NaNs.  Not sure if I am doing something wrong with the `ManagedVector`

zingale:
These are the remaining problems to switch:
```
./reacting_tests/reacting_bubble/
./radiation_tests/RadSphere/
./radiation_tests/RadThermalWave/
./gravity_tests/code_comp/```

**Comment by zingale** [comment]
for `reacting_bubble`, we need to wait until the rest of the problem is moved to C++, including the model parser, since it uses the model parser.

**Comment by zingale** [comment]
we should wait until the dt issue is fixed for the 2 radiation tests before converting those, so we can test this.  But those look straightforward to convert otherwise.

**Comment by zingale** [comment]
depends on #1201

**Comment by zingale** [comment]
some errors are large:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-09-07-001/index.html

**Comment by zingale** [comment]
I think radsphere is okay, even though some errors seem large.  The analytic solution is a really crazy function, so roundoff will happen.  But the plot in the test suite shows that it looks good:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-09-07-001/radsphere.png

**Comment by zingale** [comment]
likewise, I think that the RadThermalWave diffs are roundoff in a highly nonlinear function

File: Exec/gravity_tests/code_comp/Problem_Derive.cpp [hunk 1]
```diff
@@ -4,88 +4,130 @@
...
-
-#pragma gpu
-        deradinvariant(der, AMREX_INT_ANYD(der_lo), AMREX_INT_ANYD(der_hi), *nvar,
-                      data, AMREX_INT_ANYD(data_lo), AMREX_INT_ANYD(data_hi), *ncomp,
-                      AMREX_INT_ANYD(lo), AMREX_INT_ANYD(hi),
-                      AMREX_INT_ANYD(domain_lo), AMREX_INT_ANYD(domain_hi),
-                      AMREX_REAL_ANYD(dx), *time);
-    }
+  GpuArray<Real, 3> center;
+  ca_get_center(center.begin());
```
maxpkatz, line 20: center is unused

Exec/hydro_tests/gamma_law_bubble/Problem_Derive.cpp
```diff
@@ -1,99 +1,390 @@
...
+          (density[j] + density[j-1]) * std::abs(gravity::const_grav);
+     }
+
+     eos_t eos_state;
 
+     eos_state.rho = density[j];
+     eos_state.T = temp[j];
+     for (int n = 0; n < NumSpec; n++) {
+       eos_state.xn[n] = xn[n];
+     }
```
maxpkatz, line 82: I guess for this problem it's fine to not do naux, but this PR makes me think we really should have a convenience function that fills an eos_state from a state array so that we don't forget it in general.

Exec/hydro_tests/gamma_law_bubble/Problem_Derive.cpp
```diff
@@ -1,99 +1,390 @@
...
+
+  // only initialize the first species
+  Real xn[NumSpec];
+  xn[0] = 1.0_rt;
+
+  // compute the pressure scale height (for an isothermal, ideal-gas
+  // atmosphere)
+  Real H = pres_base / dens_base / std::abs(gravity::const_grav);
+
+  for (int j = 0; j < npts_1d; j++) {
```
maxpkatz, line 344: This looks fine but we repeat this multiple times so I wonder if we could later create a function to do this.

zingale: I put this code in a inline function in a new `prob_util.H`

Exec/hydro_tests/gamma_law_bubble/Problem_Derive.cpp
zingale: yeah, I think we can do that for filling from a conserved or primitive state, but this is the initial conditions, so it really only makes sense when the runtime parameters are there.

Exec/gravity_tests/code_comp/Problem_Derive.cpp
zingale: fixed

zingale:
going to split this up

Issue #1187, opened 2020-09-03
convert ext_src to C++


zingale:
Currently 8 problems have their own version of `ext_src_nd.F90`:
```
./science/wdmerger/ext_src_nd.F90
./science/planet/ext_src_nd.F90
./science/Detonation/ext_src_nd.F90
./hydro_tests/heating_verification/ext_src_nd.F90
./hydro_tests/test_convect/ext_src_nd.F90
./hydro_tests/toy_convect/ext_src_nd.F90
./gravity_tests/hydrostatic_adjust/ext_src_nd.F90
./gravity_tests/code_comp/ext_src_nd.F90
```
by converting these to C++, we can remove the following code:
```
Source/hydro/hybrid_advection_nd.F90
Source/rotation/Rotation_nd.F90 
```
and maybe others

PR #1191, opened 2020-09-04
Convert ext_src to C++

harpolea:
## PR summary

This address #1187. `fill_ext_source` is moved from `Castro_external.cpp` into its own source file so that this can be overwritten for individual problems.

## PR checklist


maxpkatz:
Rather than override the full `fill_ext_source` routine, what I'd like is for that routine to call an inline function on a per-zone basis which is a header file, and the problem setups should override that header file with the implementation of the function which just fills a given zone.

Also only problems marked `GPU_COMPATIBLE_PROBLEM` in the Makefile should have this change. If that's not the case, then we should retain the old implementation.

harpolea:
How would that work for problems like `wdmerger` with parameters that are calculated for the whole domain as a function of `ca_ext_src`'s input arguments? (see https://github.com/AMReX-Astro/Castro/blob/main/Exec/science/wdmerger/ext_src_nd.F90). Would those parameters be calculated everywhere in the parameter space (so they'd be moved to within the inline function)? 

Also for some problems (e.g. `wdmerger`) there are multiple problem types. Would the if statement for those go within the inline function?

maxpkatz:
Yes, in this case we would just have redundant calculations for each zone, so all of that calculation and the conditionals would move inside the function. This increases the cost a little bit but we never expect ext_src to be a significant part of the computation time so that's OK.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-09-10-002/index.html
hydrostatic_adjust is expected because we no longer use both the old and new state at the same time
wdmerger is roundoff

Exec/gravity_tests/code_comp/Castro_ext_src.H
```diff
@@ -0,0 +1,33 @@
...
+#include <prob_parameters.H>
+
+using namespace amrex;
+
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE
+void do_ext_src(int i, int j, int k,
+                GeometryData const& geomdata, 
+                Array4<const Real> const& state,
+                Array4<Real> const& src,
+                GpuArray<Real, 3> center,
```
zingale, line 13: should center be a reference? or does this not matter?

Exec/gravity_tests/hydrostatic_adjust/Castro_ext_src.H
```diff
@@ -0,0 +1,58 @@
...
+
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE
+void do_ext_src(int i, int j, int k,
+                GeometryData const& geomdata, 
+                Array4<const Real> const& state,
+                Array4<Real> const& src,
+                GpuArray<Real, 3> center,
+                const Real dt)
+{
+    const Real* prob_lo = geomdata.ProbLo();
```
zingale, line 21: this is missing the check on `time < t_step`.  For this problem, we only want to heat for some of the runtime.

zingale: I think this means that we should always have `do_ext_src` pass in the time

Exec/gravity_tests/hydrostatic_adjust/Castro_ext_src.H
```diff
@@ -0,0 +1,58 @@
...
+
+        auto x = prob_lo[0] + (Real(i) + 0.5_rt)*dx[0] - center[0];
+        auto y = prob_lo[1] + (Real(j) + 0.5_rt)*dx[1] - center[1];
+        auto z = prob_lo[2] + (Real(k) + 0.5_rt)*dx[2] - center[2];
+
+        auto dist = std::sqrt(x*x + y*y + z*z);
+
+        auto Hext = H_0 * std::exp(-((dist - r_0)*(dist - r_0))/(W_0*W_0)) * 
+            state(i,j,k,UFS-1+ihe4) / state(i,j,k,URHO);
+
```
zingale, line 52: I think that `networ_spec_index` returns a 0-based index in C++, which means we should index He just as `UFS+ihe4` and not `UFS-1+ihe4`

Exec/gravity_tests/code_comp/Castro_ext_src.H
harpolea: I had intended this to be a reference, but checking AMReX there are a few examples of GpuArrays not being passed by reference to functions, so maybe it's not necessary? I'll switch it to be passed by reference anyway.

Exec/gravity_tests/hydrostatic_adjust/Castro_ext_src.H
harpolea: Yeah, I skipped this because I didn't want to have to pass another argument (especially as it's only required for this one problem), but will switch them all so time is passed in

Exec/gravity_tests/code_comp/Castro_ext_src.H
zingale: I don't know which is preferred here.  I recall once deciding that we should pass these all by reference, but since this will be inlined, maybe it doesn't matter anyway.

Exec/gravity_tests/hydrostatic_adjust/Castro_ext_src.H
zingale: I don't see the shut off the tstop, only the print
so this is still heating for t > t_stop

harpolea: There's `if (time > t_stop) return;` on line 21

maxpkatz (approved):

zingale (approved):

Issue #391, opened 2018-08-30
have VODE start with the last timestep


zingale:
VODE currently uses its own internal algorithm to decide on the starting timestep.  If you look at the path that VODE takes in phase space, it has lots of small steps at the start of integration.  We should be able to start with a larger step, especially for SDC in later iterations.

VODE stores this information in `RWORK(5)`, with a value of `0` indicating that it should determine the initial step size.

dwillcox:
I'm taking a look at this, I should have a PR for you to look at sometime soon.

dwillcox:
Yes, the alternative using VODE is to save the state data and restore it, but that's a lot of memory. There doesn't seem to be a good way to avoid saving its internal state and still start at high order since it needs to bootstrap itself up to high order when it starts from scratch.

zingale:
okay.  Closing this then.  We can explore other integrators separately.

PR #1195, opened 2020-09-05
switch to the init provided by Microphysics

### LINKED ISSUE #1194: have castro use microphysics_init
Since we are using the interfaces from `Microphysics/interfaces/`, we should use the `microphysics_init()` routine there to initialize the EOS, network, and other stuff.  This way we don't need to maintain our own versions in `Castro_nd.F90`.  The one trick is that currently it takes optional arguments, so we'll need to provide a C-safe wrapper.

zingale:
closes #1194

## PR checklist


zingale:
requires https://github.com/starkiller-astro/Microphysics/pull/391

maxpkatz:
In Fortran, the reason we pass small_temp and small_dens to EOS initialization is that small_temp and small_dens may be altered (if they're smaller than the EOS mintemp and mindens). In C++ that is done separately here: https://github.com/AMReX-Astro/Castro/blob/8b1da109c99412a6c1af9d7dd2d6b381a1c7939b/Source/driver/Castro_setup.cpp#L240-L249

I'm not sure which approach I like more, but if we're going to keep this bit of code, then we don't need to pass small_temp and small_dens to the EOS. But I don't think it's a good idea if we have the same code in both Castro and Microphysics, so perhaps if we want to keep the C++ version we should do the same thing in Fortran.

zingale:
I'm not sure I fully understand.  Ultimately, I want small_dens and small_temp to be set in C++ and propagated elsewhere as needed.

I don't see why we don't need that extra code, but perhaps I should move it to before the initialization?

maxpkatz:
What I'm saying is that the bit in those ten lines of code is also done in the Fortran eos_init. So now it's being done twice. One possible conclusion is that we just don't need to pass small_temp and small_dens to Fortran.

zingale:
okay, this now calls microphysics_init from Fortran

maxpkatz:
I am good with this now; let's see what the test suite says.

zingale:
tests look good: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-09-06/index.html

maxpkatz (approved):

PR #1204, opened 2020-09-07
ca_initdata does not abort if unused

maxpkatz:
## PR summary

The state data is initialized to zero before the call to ca_initdata and there is a check to ensure rho and T are > 0, so we will still get an abort if for some reason ca_initdata was not overridden for a given problem.

## PR motivation

This will allow us to implement a C++ version of ca_initdata without requiring a special ifdef.

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-09-07-004/index.html

zingale (approved):

PR #1210, opened 2020-09-14
Add a _default_prob_params file

maxpkatz:
## PR summary

This opens up the option to have parameters that are defined for every problem (and thus effectively for the code as a whole). We may want some global parameters to live in the problem parameters namespace because they're updated in the F90 amrex_probinit and so it's easiest to have them automatically updated afterward, using already existing functionality. In particular, the goal is to do `center` this way (which will be added in a later PR).

## PR motivation

This will make it easier to convert Fortran problem setups to C++ (the goal is for every variable that is modified/set in probinit to be part of _prob_params so that it's automatically updated for us in C++).

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-09-14-002/index.html

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-09-20-002/index.html

zingale (approved):

PR #1222, opened 2020-09-22
Convert center to auto prob params

maxpkatz:
## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-09-22-001/index.html

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-09-23-001/index.html

maxpkatz:
The only failure was RadSuOlson which was due to a flaw in my logic in the probdata script (for the case with a non-zero number of parameters but none of which use the fortin namelist). This has been fixed so we can retest.

zingale:
retest: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-09-23-002/index.html

maxpkatz:
That crashes because the problem doesn't call `probdata_init()`. This is a problem because we cannot assume external setups will know to call it, so fixing it in all of the problems in Exec/ is not a real solution. I'll need to remove the dependence on the problem calling it in a separate PR before we can merge this one.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-09-23-006/index.html
now Sedov fails to compile

maxpkatz:
Sedov should be fixed

zingale (approved):

PR #1224, opened 2020-09-22
Problem parameters not in namelist also have a C++ version

maxpkatz:
## PR summary

Previously you could add parameters to _prob_params that were not in the Fortran namelist. No C++ version was made of these parameters. However, there is a need for even the ones not in the namelist to have a C++ version, since we may set those variables in probinit and want them later in C++. This implements that change, with the exception of parameters whose size depends on a Fortran module parameter (since we have no direct analogue of that in C++). As a demonstration, many of the wdmerger C++ problem parameters (which were manually managed as duplicates of the corresponding Fortran parameters) are converted to this new format.

## PR motivation

This will also be helpful for finishing up #1222 since we will need `center` in C++.

## PR checklist

### DIFF
--- a/Exec/hydro_tests/toy_convect/Prob_nd.F90
+++ b/Exec/hydro_tests/toy_convect/Prob_nd.F90
@@ -18,6 +18,10 @@ subroutine amrex_probinit(init, name, namlen, problo, probhi) bind(c)
   ! Namelist defaults
   call probdata_init(name, namlen)
 
+  if (num_vortices > max_num_vortices) then
+     call castro_error("num_vortices too large, please increase max_num_vortices and the size of xloc_vortices")
+  end if
+
   ! Read initial model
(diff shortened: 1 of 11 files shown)


zingale (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-09-22-006/index.html

maxpkatz:
I've addressed the compile fails for xrb_mixed and toy_convect by hardcoding the maximum number of vortices (since we declare the global `__managed__` data in C++ as a fixed-size array).

I notice that xrb_mixed does not currently run, because interp_BC in the probin is not a recognized parameter in the _prob_params setup.

zingale:
those tests pass now: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-09-22-008/index.html

PR #1241, opened 2020-09-28
set the composition if we are in NSE appropriately

zingale:
## PR checklist

### DIFF
--- a/Exec/science/rotating_star/Prob_nd.F90
+++ b/Exec/science/rotating_star/Prob_nd.F90
@@ -72,7 +72,7 @@ subroutine ca_initdata(level, time, lo, hi, nscal, &
 #ifdef NSE_THERMO
   use burn_type_module
   use nse_check_module
-  use nse_module
+  use nse_module, only: nse_interp
 #endif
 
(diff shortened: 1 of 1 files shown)


harpolea (approved):

Exec/science/rotating_star/Prob_nd.F90
```diff
@@ -69,6 +69,11 @@ subroutine ca_initdata(level, time, lo, hi, nscal, &
   use eos_type_module
   use amrex_constants_module, only : ZERO, HALF, ONE, TWO
   use amrex_fort_module, only : rt => amrex_real
+#ifdef NSE_THERMO
+  use burn_type_module
+  use nse_check_module
+  use nse_module
```
harpolea, line 75:
```suggestion
  use nse_module, only: nse_interp
```

Need to only include `nse_interp` or else there is a conflict with `nspec`

harpolea (changes requested):

zingale:
I've added the only

harpolea (approved):

Issue #154, opened 2017-08-08
ca_denfill is used for things other than density


zingale:
In particular, it seems that `ca_denfill` is used for `Source_Type`, `SDC_Source_Type`, `SDC_React_Type`

This can be a problem, since some code with inflow / external boundaries takes specific actions for the density.  For example, the HSE BCs

maxpkatz:
It is also the case that denfill is used when filling boundaries for derived data that need a ghost zone (e.g. divu) so it's still not safe for denfill to actually assume density. We should stop calling it denfill and make it generic for any state component.

zingale:
I don't think this is true anymore.  Can we close this?

maxpkatz:
This is still a problem with `ext_denfill`. Despite the name, this routine is not called on only the density; it's called whenever we need a single component of the state. In practice in Castro, this will happen when we use `AmrLevel::derive()` to make a copy of a particular state variable, and if that is not the density, then `ext_denfill` is not right. Fortunately this is not a very serious problem because in the `derive()` cases I can think of (tagging and diagnostics) we don't end up using the ghost zones.

maxpkatz:
We're also making this mistake with the option `fill_ambient_bc`.

zingale:
this is now fixed in #1356.

PR #1247, opened 2020-09-29
start porting the model parser to C++

zingale:
## PR checklist


Util/model_parser_cxx/model_parser.H
```diff
@@ -0,0 +1,308 @@
...
+/// we read in the number of variables and their order and use this to
+/// map them into the model_state array.  We ignore anything other than
+/// density, temperature, pressure and composition.
+///
+/// composition is assumed to be in terms of mass fractions
+
+// remove whitespace -- from stackoverflow
+
+
+inline std::string& ltrim(std::string& s)
```
maxpkatz, line 39: Can we please put these functions in a class or namespace?

zingale: done.

Util/model_parser_cxx/model_parser_data.H
```diff
@@ -0,0 +1,21 @@
...
+    constexpr int nvars = 3 + NumSpec;
+    constexpr int idens = 0;
+    constexpr int itemp = 1;
+    constexpr int ipres = 2;
+    constexpr int ispec = 3;
+
+    extern AMREX_GPU_MANAGED int npts;
+    extern AMREX_GPU_MANAGED bool initialized;
+
+    extern amrex::Gpu::ManagedVector<amrex::Real> state;
```
maxpkatz, line 18: If the model data is going to be a std::vector, we cannot directly access it inside a ParallelFor if we're running on the GPU. We'd have to get its `dataPtr()` prior to the ParallelFor and then pass that pointer to the `interpolate` function.

maxpkatz: I suggest having the data be separate managed Array1Ds (similar to what we do in other places like Helmholtz) for density, pressure, etc., with a fixed size at compile time that is probably bigger than we realistically need. (Even if it's a large number like 10k, it won't be that much memory consumed.) If the user has a larger model than our default, they can change it themselves.

zingale: but this would mean an Array1D for each species?  So I would store those as a Vector of size NumSpec of Array1D?

maxpkatz: Since NumSpec is known at compile time, that one can be an Array2D.

zingale: Ah, indeed.  okay.  I'll do this.  I'll make the number of points a preprocessor variable.

zingale: okay, I made this all an Array2D, and there is a make variable MAX_NPTS_MODEL that you can set to change the size.  I still have the BC issue with the divu / magvort though.

Util/model_parser_cxx/model_parser_data.cpp
```diff
@@ -0,0 +1,14 @@
...
+#include <AMReX_GpuContainers.H>
+#include <model_parser.H>
+
+namespace model
+{
+
+    AMREX_GPU_MANAGED int npts;
+    AMREX_GPU_MANAGED bool initialized;
+
+    amrex::Array2D<Real, 0, NPTS_MODEL-1, 0, nvars-1> state;
```
maxpkatz, line 11: Note: these need to be AMREX_GPU_MANAGED also

Source/problems/Castro_bc_fill_nd.cpp
```diff
@@ -109,17 +110,21 @@ void ca_statefill(Box const& bx, FArrayBox& data,
...
     if (numcomp == 1) {
 
+#ifndef CXX_MODEL_PARSER
 #pragma gpu box(bx)
         ext_denfill(AMREX_INT_ANYD(bx.loVect()), AMREX_INT_ANYD(bx.hiVect()),
                     BL_TO_FORTRAN_N_ANYD(data, dcomp),
                     AMREX_INT_ANYD(geom.Domain().loVect()), AMREX_INT_ANYD(geom.Domain().hiVect()),
                     AMREX_REAL_ANYD(geom.CellSize()), AMREX_REAL_ANYD(geom.ProbLo()), time, bc_f);
+#else
+        ext_denfill_c(bx, data.array(), dcomp, geom, bcr[dcomp], time);
```
maxpkatz, line 126: The `array()` method has an overload that accepts the component, so you could do `data.array(dcomp)` here if you wanted the denfill routine to look more like it does in Fortran now.

Util/model_parser_cxx/model_parser.H
```diff
@@ -0,0 +1,307 @@
...
+
+    // allocate storage for the model data
+
+    amrex::Print() << "reading initial model" << std::endl;
+    amrex::Print() << model::npts << " points found in the initial model" << std::endl;
+    amrex::Print() << nvars_model_file << " variables found in the initial model file" << std::endl;
+
+    // start reading in the data
+
+    Real vars_stored[nvars_model_file];
```
maxpkatz, line 216: Since variable-length arrays are not legal C++ this should probably be a vector.

Source/problems/Castro_bc_fill_nd.cpp
maxpkatz, line 126: The indexing into `bcr` is not consistent with the Fortran here -- `dcomp` is the destination component to be filled in the Fab, but `bcr` is a vector of length `numcomp` which are the BC values corresponding to components `dcomp` to `dcomp + numcomp - 1`.  So here you want to be doing `bcr[0]` here for consistency.

Exec/hydro_tests/toy_convect/problem_setup.H
```diff
@@ -0,0 +1,131 @@
...
+
+        for (int vortex = 0; vortex < problem::num_vortices; vortex++) {
+
+            Real xdist = x - problem::xloc_vortices[vortex];
+
+            Real r = std::sqrt(xdist * xdist + ydist * ydist);
+
+            upert[0] += -(ydist/problem::velpert_scale) * problem::velpert_amplitude *
+                std::exp(-r * r / (2.0_rt * problem::velpert_scale * problem::velpert_scale)) *
+                std::pow(-1.0_rt, vortex);
```
maxpkatz, line 113: As written this makes the velocity perturbations exactly opposite in sign to how they were in Fortran -- we need to make the exponent `vortex + 1` to match.

maxpkatz:
This PR reminds me that we still need to fix #154; this is an example of a case where, whenever we ask for a single component for a boundary fill (like in the derives) we're just filling the ghsot zones with the density at the base of the atmosphere regardless of what the component is.

Util/model_parser_cxx/model_parser_data.cpp
zingale: fixed

Source/problems/Castro_bc_fill_nd.cpp
zingale: fixed

zingale:
I agree re: 154 -- we should just do FOEXTRAP here I think.

Exec/hydro_tests/toy_convect/problem_setup.H
zingale: fixed

zingale:
test: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-05-002/toy_convect.html

Util/model_parser_cxx/model_parser.H
zingale: fixed

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-05-003/toy_convect.html

maxpkatz (approved):

PR #1248, opened 2020-09-30
on GPUs we were disabling the ext denfill BCs for the x-dir

zingale:
## PR checklist

### DIFF
--- a/Source/problems/bc_ext_fill_nd.F90
+++ b/Source/problems/bc_ext_fill_nd.F90
@@ -941,7 +941,6 @@ contains
     ! that the same function is called here and in hypfill where all the
     ! states are filled.
 
-#ifndef AMREX_USE_CUDA
     ! XLO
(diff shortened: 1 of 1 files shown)


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gpu/2020-09-30-001/index.html

maxpkatz:
Do we know why this was disabled originally?

zingale:
I suspect it was just a mistake

maxpkatz:
OK, I agree it was an oversight. Commit 54cbab886834682513ee53c0f541211fcc39a63c added ext_denfill but not in the x direction, with an error case if calling in that mode that was ifdef'ed out for CUDA, presumably because we didn't have amrex_error for CUDA. Commit 18c497f39714806722c6c213e757b60b3643959f turned on CUDA support but x-direction support still wasn't there at that time; it was later added in #654, and the errors were replaced with the BC fill code but we forgot to remove the ifdef guarding out those errors.

maxpkatz (approved):

PR #1257, opened 2020-10-05
move all of the runtime parameter initialization to a central place

zingale:
## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-10-05-002/index.html

I'm not sure why the radiation stuff is failing

zingale:
actually... I suspect it is because we do some magic after the radiation params are set and before the radsolve params are set
so perhaps I need to leave the radsolve params where they were for now.

maxpkatz:
I'm OK with leaving RadSolve out for now. The parameters need to be dealt with more comprehensively there anyway.

zingale:
updated to put the radsolve stuff back into the radsolve constructor so the radiation tests pass.  New tests:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-10-06-001/index.html

maxpkatz (approved):

PR #1259, opened 2020-10-07
remove an extra interface for the ext fill stuff

zingale:
## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-10-07-001/index.html

maxpkatz:
We should do this after #1281.

maxpkatz:
Elaborating on the above: the reason for the convoluted structure where the standard and external BC interfaces are in different files is that we used to have CUDA register capping (which is applied on a per-source-file basis) for the standard BCs, but it wasn't necessary for the external BCs. Now that we no longer need the special CUDA make rules, it should be fine (and definitely cleaner) to merge everything back into one file.

zingale:
I'll merge `development` and retest

PR #1260, opened 2020-10-13
Permit level dependent thresholds in new tagging scheme

maxpkatz:
## PR summary

We take advantage of AMReX-Codes/amrex#1441 to allow the user to specify a tagging threshold that can vary by level. If a value for every level is not provided, we assume the last value holds for all remaining levels, similar to how we treat quantities like amr.n_error_buf.

## PR checklist

### DIFF
--- a/Source/driver/Castro.cpp
+++ b/Source/driver/Castro.cpp
@@ -500,23 +500,23 @@ Castro::read_params ()
             info.SetMaxLevel(max_level);
         }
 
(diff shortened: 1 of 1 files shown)


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-10-13-001/index.html

zingale (approved):

PR #1272, opened 2020-10-27
Add diagnostic information if we retry for small/neg density

maxpkatz:
## PR summary

On a retry we print the new density, the old density, and the zone index, to help the user understand the context for where the update went wrong. Although this adds some parallel reductions, they only occur on a retry, which should be uncommon.

## PR checklist

### DIFF
--- a/Source/driver/Castro_advance_ctu.cpp
+++ b/Source/driver/Castro_advance_ctu.cpp
@@ -194,14 +194,40 @@ Castro::do_advance_ctu(Real time,
 
       Real minimum_density = S_new.min(URHO);
 
-      if (minimum_density < 0.0_rt) {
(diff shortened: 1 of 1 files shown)


maxpkatz:
Sample output (using `inputs_2d_collision` in `wdmerger_collision` with additional options `amr.n_cell=32 64 amr.max_level=1`):
```
Advance was unsuccessful with reason: small density (density = 9.998080e-06 at index (33,124); started at 1.000000e-05); proceeding to a retry.
```

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-10-27-002/index.html

zingale (approved):

PR #1273, opened 2020-10-27
Allow retry to be skipped if starting density is too small 

maxpkatz:
## PR summary

A new option, `castro.retry_small_density_cutoff`, has been added. In some cases a small or negative density retry may be triggered on an update that moves a zone already close to small_dens just below it. This is not uncommon for "ambient"/"fluff" material outside a star. Since these zones are not dynamically important anyway, triggering a retry is unnecessary (and possibly counterproductive, since it may require a very small timestep to avoid). By setting this cutoff value appropriately, the retry will be skipped if the density of the zone prior to the update was below the cutoff. The zone will be reset using the standard approach in enforce_min_dens.

## PR motivation

This is done in conjunction with #1272, which provides us the necessary data for making this determination. I have observed that the WD collision problem can get into situations where densities just above small_dens get pushed just below it. It's very hard to fix this problem with a smaller dt, since the difference between the starting and ending densities may be very small already in absolute terms. So the problem setup can crash after hitting too many retry subcycles. Since we don't care about zones near small_dens anyway, it is more efficient to just skip the retry in that case and let clean_state handle it.

## PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,14 @@
+# 20.11
+
+   * A new option, `castro.retry_small_density_cutoff`, has been added. In some
+     cases a small or negative density retry may be triggered on an update that
+     moves a zone already close to small_dens just below it. This is not uncommon
+     for "ambient"/"fluff" material outside a star. Since these zones are not
+     dynamically important anyway, triggering a retry is unnecessary (and possibly
+     counterproductive, since it may require a very small timestep to avoid). By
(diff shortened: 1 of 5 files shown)


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-10-27-003/index.html

zingale (approved):

PR #1281, opened 2020-10-30
Remove old CUDA make rules

maxpkatz:
## PR summary

We had rules in the build system that capped the number of registers per thread used in the CUDA build for the boundary condition routines. This allowed us to be consistent with similar logic I had inserted in AMReX in the Fortran filcc. However, we no longer use that boundary condition interface, so the special handling we have in the build system is no longer necessary (#819, #1048, #1049). These CUDA build rules are a holdover from that time and have been obsolete for a while.

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gpu/2020-10-30-001/index.html

performance does not look to be impacted.

zingale (approved):

PR #1219, opened 2020-09-18
Convert estdt_burning to C++

harpolea:
## PR summary

This converts the `estdt_burning` subroutine to C++. It also converts `okay_to_burn` in `react_util.F90` to C++.

## PR checklist

- [ ] test suite needs to be run on this PR
- [ ] this PR will change answers in the test suite to more than roundoff level
- [ ] all newly-added functions have docstrings as per the coding conventions
- [ ] the `CHANGES` file has been updated, if appropriate
- [ ] if appropriate, this change is described in the docs

PR #1282, opened 2020-10-30
Allow problems to turn off use of GPU pragma

### LINKED ISSUE #1203: eliminate #pragma gpu for Sedov
As a first pass on portability, we should eliminate the CUDA Fortran in Sedov.  A build of Sedov points out where all of the `#pragma gpu` are in the compiler warnings.  The following need to be updated:

- [x] `ca_initdata`
- [ ] `set_problem_tags`
- [ ] `denfill`
- [ ] `hypfill`
- [ ] `ext_fill`
- [ ] `ext_denfill`
- [x] `ca_ext_src`

maxpkatz:
## PR summary

If a problem is known to not require the GPU pragma script anymore, this can now be disabled by setting `USE_GPU_PRAGMA = FALSE` in its `GNUmakefile`. We implement this for Sedov and thus this closes #1203. Technically we have not eliminated all of the places in the Sedov build where pragma gpu is still in a compile path -- as mentioned there, the relevant places are `set_problem_tags` and then `hypfill`/`denfill` and `ext_fill` and `ext_denfill`. However, `set_problem_tags` does no work for Sedov, so there's no need for a GPU compile (it will just be a no-op on the CPU). A similar comment is true for the BC fill routines -- they only still do work at this point if we're using ambient BCs (`hypfill`/`denfill`) or HSE BCs (`ext_fill`/`ext_denfill`). So we should be able to eliminate the GPU build for those routines and trust that there is no performance impact because the loops that do any work never get evaluated.

## PR checklist


maxpkatz:
The only subtlely I am currently aware of is that we do rely on some functionality in AMReX which is enabled by `AMREX_USE_PRAGMA`: the reducer helper macro, `AMREX_MFITER_REDUCE_{MIN,MAX,SUM}`. However it looks like the only place this is still used outside of radiation is in the Fortran reactions, so we should be OK to use this in problems where `USE_CXX_REACTIONS = TRUE` once #1219 is complete.

maxpkatz:
I've tried Sedov on this branch and there are no obvious gaps on the GPU timeline reported by Nsight Systems (which would be there if it was relying on a GPU loop which had now been accidentally turned off), so I think this is OK.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-10-30-002/index.html

zingale (approved):

PR #1284, opened 2020-10-31
Bump up minimum C++ standard to C++14

maxpkatz:
## PR summary

C++14 is now required as the minimum C++ standard. This should cause no changes except for RHEL 7/CentOS 7 systems where users will be required to use a more recent gcc than the system default.

## PR motivation

This will support some C++14 constructs we want to add to Microphysics.

## PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # 20.11
 
+   * The minimum C++ standard supported by Castro is now C++14. Most modern compilers
+     support C++14; the notable exception is RHEL 7 and its derivatives like CentOS 7,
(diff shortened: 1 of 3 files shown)


zingale:
tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-10-31/index.html

zingale (approved):

maxpkatz:
Since this would break the case where we use COMP=PGI on Summit (which blocks us from loading a new enough gcc module), I added a workaround to hardcode the gcc used by nvcc in Make.Castro. This workaround can be removed when we have switched over all of the problem setups to not require PGI anymore.

PR #1286, opened 2020-11-01
Implement estdt_burning in C++

maxpkatz:
## PR checklist


maxpkatz:
This replaces #1219 as we still need to have the Fortran implementation for now.

Source/driver/reduction_nd.F90
```diff
@@ -63,6 +63,9 @@ subroutine reduce_min(x, y)
     real(rt), intent(in   ) :: y
     real(rt), intent(inout) :: x
 
+#ifdef _OPENMP
+    !$omp atomic update
```
maxpkatz, line 67: Normally we don't need this; this is a workaround for the fact that I removed the parallel reduction on the outer region. This routine is only still used by estdt_burning anyway, so when we switch to only the C++ interfaces, it will go away.

maxpkatz:
Note: this is gated on having a test in the suite which exercises this functionality.

maxpkatz:
rotating_star is now in the test suite (AMReX-Astro/actual_test_files#3) and it exercises this functionality (its timestep is set by the burning timestep limiter).

maxpkatz:
Current tests here: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/

Looks like some tests aren't compiling.

zingale (approved):

PR #1294, opened 2020-11-04
only support USE_SIMPLIFIED_SDC with USE_CXX_REACTIONS=TRUE

zingale:
## PR summary

Now that we have wrappers that provide C++ interfaces for any pure-Fortran network, we can enforce that
simplified-SDC only worked with a C++ network.  This will make it easier for use to develop new features by not requiring us to port those features to both the C++ and Fortran code paths.

## PR checklist


maxpkatz (approved):

zingale:
The detonation comparisons are way off because the C++ version does a retry very early on (due to VODE having h < hmin)

CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
    * An interface for doing problem tagging in C++ has been added. (#1289)
 
+   * Simplified SDC now only supports C++ reaction networks (#1294)
```
maxpkatz, line 5: Can  you clarify this comment? What we only support now is C++ integrators. You can still use a Fortran network.

zingale: done

maxpkatz (approved):

PR #1299, opened 2020-11-06
convert StarGrav to C++

zingale:
## PR checklist


zingale:
B field stuff is not tested in the suite.  Here's the dipole field:

Screenshot from 2020-11-06 13-00-56
Screenshot from 2020-11-06 13-01-03
Screenshot from 2020-11-06 13-01-10

zingale:
depends on #1298

zingale:
errors seem big: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-06-006/StarGrav-2d.html

zingale:
the difference seems to be most pronounced at a single zone right at the center.

curiously, the differences at step 0 are all O(1.e-12) or smaller

maxpkatz:
We have tested this carefully and it is correct. The pronounced differences at timestep 25 turned out to be unrelated (#1373); the unrelated issue was unintentionally exposed due to small roundoff differences propagating a certain way in development and a different way in the new setup.

maxpkatz (approved):

PR #1302, opened 2020-11-07
Apply Poisson gravity tagging restriction to all setups

maxpkatz:
## PR summary

Poisson gravity doesn't have support for fine levels that touch the physical domain boundary (that is, the boundary is not either periodic or symmetric). This is enforced as an abort in the Poisson solver, but we can also do better by just preventing this from happening directly. This was already the case in wdmerger; this PR applies this across all problems that use Poisson gravity.

## PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # 20.12
 
+   * For setups using Poisson gravity, tagging is now turned off in locations where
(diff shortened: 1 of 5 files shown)


zingale (approved):

zingale:
all tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-07-002/index.html

PR #1303, opened 2020-11-07
Remove wdmerger tagging capability based on distance from center

maxpkatz:
## PR summary

This capability is not currently used in any science setup so it is no longer needed.

## PR checklist


maxpkatz:
The wdmerger_retry test was using this tagging scheme so that will need a reset. However the test still is subject to retries even without AMR so we should keep it in the suite.

zingale (approved):

zingale:
several tests crash: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-07-004/index.html

maxpkatz:
Fixed

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-07-005/index.html

those failures are expected

PR #1305, opened 2020-11-08
Convert Detonation early stopping to C++

maxpkatz:
## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-08-004/index.html

it doesn't look like this behaves the same way

maxpkatz:
This should be fixed. I made a subtle mistake -- I declared the type of these cutoff parameters as "Real" rather than "real" in the _prob_params file. As a result, our parser treated that as a custom type and just inserted it as "Real" into the prob_params_auto.F90 file, rather than `real (kind=rt)` as we intended. Since Fortran is case insensitive that compiled and it got treated as a single precision type instead of double precision, and as a result the value got truncated and ended up in C++ as about 1.3e9, hence the slightly earlier stopping.

zingale:
tests pass now: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-08-007/index.html

zingale (approved):

PR #1311, opened 2020-11-08
Convert problem_checkpoint and problem_restart to C++

maxpkatz:
## PR summary

`problem_checkpoint` and `problem_restart` are moved to C++ from Fortran. At this point, only wdmerger still implemented them, so that is switched over, and stubs are placed in `Source/problems`. Since those were the only two routines in `Problem_F.H` and `Problem.f90`, those files are deleted.

## PR checklist


maxpkatz:
We don't really have tests of this functionality -- I think it would require a wdmerger restart test -- but I have manually verified that the checkpoint/restart functionality works as I expect in wdmerger, and actually fixes a bug in that problem.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-08-011/index.html

zingale (approved):

Issue #422, opened 2018-10-11
add the tile size to the job_info file


zingale:
It is useful for scaling studies to have the tile size in the `job_info` file along with the number of MPI tasks and OMP threads.

PR #1313, opened 2020-11-09
Switch from Fortran to C++ reactions

maxpkatz:
## PR summary

The Fortran reactions interfaces (except for true SDC) are removed.

## PR checklist


maxpkatz:
Initial test run: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-08-014/index.html

Most tests either pass or the diffs are acceptably small. But some tests (e.g. convective_flame) fail right out of the gate with burns that do not complete.

maxpkatz:
The failing burns are due to the custom logic we have which prevents abundances from changing too much in a single integration step:

https://github.com/starkiller-astro/Microphysics/blob/ccd82d1e962b79e15d19fd77575b69e536550553/integration/VODE/vode_dvstep.H#L224-L228

The requirement for this logic is X > atol_spec, but atol_spec is 1.e-12 by default and that also happens to be our default small_x for many problems. So the trace species are trying to move away from small_x (to 1.e-10, say) and are not able to because those changes are much larger than the allowable VODE change factors.

zingale (approved):

zingale:
tests with the looser tolerances (https://github.com/starkiller-astro/Microphysics/pull/422)
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-10-005/index.html

The one crash is because the gridding is different now.

zingale:
with the caveat that Detonation just is a bad test problem, I think the errors look fine.

maxpkatz:
wdmerger_collision gets a lot slower, but we concluded that this is because the run is in debug mode -- there is no slowdown for optimized code. So this change looks good to me.

PR #1326, opened 2020-11-12
the "interp" external BC option was removed long ago

zingale:
Running with these inputs would immediately hit an abort complaining about an unsupported BC type.
Many of these instances were not actually used, but a few were changed to
outflow.  

## PR checklist

### DIFF
--- a/Exec/gravity_tests/hse_convergence_general/inputs_1d.128.hse_test
+++ b/Exec/gravity_tests/hse_convergence_general/inputs_1d.128.hse_test
@@ -21,7 +21,7 @@ castro.xl_ext_bc_type = "hse"
 castro.hse_interp_temp = 1
 castro.hse_reflect_vels = 1
 
-castro.xr_ext_bc_type = "interp"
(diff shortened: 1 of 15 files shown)


maxpkatz:
What did interp used to be when it was available?

zingale:
interp would use the model held by the model parser to fill the ghost cells.  It looks like most of the uses were switched over to the ambient fill.

maxpkatz:
Oh OK. I was thinking that denfill basically does that now, except it's not really correct because denfill never actually fills density. So we already have the code to do it, if we want to.

zingale:
correct.  We can add this back in easily if needed.  But right now, if we set a BC to "interp" the code aborts.

maxpkatz:
OK. I'm planning to submit a PR shortly to just remove denfill and override any inflow BCs with outflow, so we'll lose that code, but it'll be available in the history if we do want to resuscitate it.

zingale:
yes, that's the right thing to do.

maxpkatz (approved):

PR #1335, opened 2020-11-14
Convert get_v_dcf to C++

maxpkatz:
## PR checklist


Source/radiation/Radiation.cpp
```diff
@@ -1983,17 +1987,39 @@ void Radiation::get_rosseland_v_dcf(MultiFab& kappa_r, MultiFab& v, MultiFab& dc
...
+            auto kp_arr = kp.array();
+            auto kp2_arr = kp2.array();
+            auto c_v_arr = c_v.array();
+            auto dcf_arr = dcf[mfi].array();
+
+            amrex::ParallelFor(reg,
+            [=] AMREX_GPU_HOST_DEVICE (int i, int j, int k)
+            {
+                const Real fac0 = 4.e0_rt * sigma_loc * delta_t / dT_loc;
+                const Real fac2 = c_loc * delta_t / dT_loc;
```
zingale, line 2004: can't these be done outside of the ParallelFor?

zingale (approved):

Source/radiation/Radiation.cpp
maxpkatz: Yes, they could be, I just wasn't being too careful to avoid redundancy here because we don't spend much time in this part of the code compared to e.g. Hypre.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-14-002/index.html

PR #1337, opened 2020-11-14
clean up the cpp parameters python script

zingale:
we now ignore case in the _cpp_parameters file for the data type.
we also remove the option for the Fortran version of a parameter to
use a different name or data type.  One Fortran runtime parameter
was changed as a result (rot_periond -> rotational_period).

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-14-005/index.html

maxpkatz (approved):

PR #1338, opened 2020-11-14
convert gamma_law_bubble to C++

zingale:
## PR summary

The switches the initial model generation to be stored in the model_parser data structures.
It also removes the custom boundary conditions that were done previously to instead use
our general HSE BCs.  This change will cause some differences in the output, but it should
be fine.  The old BCs were essentially integrating the initial model into the BCs, so it was fixed
in time.

## PR checklist


maxpkatz:
There is also `inputs_2d_MAESTRO.big` which should have the inflow BC bit added?

maxpkatz (approved):

PR #1342, opened 2020-11-15
Convert Rad2Tshock BCs to C++

maxpkatz:
## PR summary

Rad2Tshock's BCs are converted to C++. Since it was previously using EXT_DIR and then overriding with ext_fill, simply moving the BCs to problem_bc_fill.H is not sufficient, because that will still have the main Castro EXT_DIR routines be called, which is not what we want. So the inputs files are updated to use outflow BCs along the problem direction -- the outflow fill will get overwritten anyway.

This does cause a breaking change in the regression suite. Previously this problem was also using ext_denfill and filling the density on either end. We do not attempt to replace it, since ext_denfill is usually incorrect to use in the way it is used here. In this case, because this was being used for derives, it caused spurious tagging at either end of the domain, where the temperature saw a "gradient" due to a density-scaled value being located adjacent to the temperature-scaled value in the interior. In this new configuration, the spurious tagging at the domain edges is avoided.

## PR checklist


maxpkatz:
Tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-14-011/index.html

comoving aborts during the run. The rest complete but do not compare cleanly due the reason I mentioned (grids on level > 0 will be different).

maxpkatz:
> comoving aborts during the run.

There are two zones with density exactly equal to zero that are causing problems during the clean_state call. The zones are ghost zones, but clean_state checks those (both in post_regrid and in initialize_advance). I'm concerned that the zones are not being correctly filled during the regrid operation, and it may be unrelated to this PR (but accidentally exposed by it).

Exec/radiation_tests/Rad2Tshock/problem_bc_fill.H
```diff
@@ -0,0 +1,119 @@
...
+    if (problem::idir == 1) {
+
+        // XLO
+        if ((bcs(URHO).lo(0) == EXT_DIR || bcs(URHO).lo(0) == FOEXTRAP) && i < domlo[0]) {
+            state(i,j,k,URHO) = problem::rho0;
+            state(i,j,k,UMX) = problem::rho0 * problem::v0;
+            state(i,j,k,UMY) = 0.e0_rt;
+            state(i,j,k,UMZ) = 0.e0_rt;
+            state(i,j,k,UFS) = state(i,j,k,URHO);
+            if (NAUX_NET > 0) {
```
zingale, line 27: I usually do this as an #if, but I imagine this is okay.

zingale (approved):

Issue #561, opened 2019-03-21
eliminate the ghost cell from State_Type when running with radiation


zingale:
The hyperbolic system does not need the ghost cell in the State_Type data, even with radiation.  The only reason it seems to be there is for constructs like this in `MGFLDRadSolver.cpp`:
```
MultiFab& S_lag = castro->get_old_data(State_Type);
for (FillPatchIterator fpi(*castro,S_lag,ngrow,oldtime,State_Type,
                                      0,S_lag.nComp()); fpi.isValid(); ++fpi) {
    S_lag[fpi].copy(fpi());
}
```
but this can be changed with 
```
FillPatchIterator fpi(....);
MultiFab& S_lag = fpi.get_mf();
```
or doing a FillPatch on a temporary multifab.

Note: that this applies to State_Type, but Rad_Type still needs one ghost cell because of how the gradient is computed for the flux limiter.  At a cource-fine boundary, the values in the ghost cells are set to -1.0 which is a flag for the coarse-fine boundary, in which case we do a one-sided derivative.

zingale:
This probably addresses #423

maxpkatz:
Assigning this to myself as part of the radiation cleanup I'm doing.

PR #1344, opened 2020-11-15
Include ghost zones when filling data on new level

maxpkatz:
## PR summary

This addresses the issue that was unintentionally exposed by #1342. When we're creating a new fine level and filling data from the coarse level below it, we need to include the ghost zones in the fillpatch operation, or else they will be uninitialized and cause problems during the subsequent advance.

## PR checklist

### DIFF
--- a/Source/driver/Castro.cpp
+++ b/Source/driver/Castro.cpp
@@ -1547,7 +1547,7 @@ Castro::init ()
 
     for (int s = 0; s < num_state_type; ++s) {
         MultiFab& state_MF = get_new_data(s);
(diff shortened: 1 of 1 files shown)


zingale:
this likely only affects radiation because of #561

maxpkatz:
Tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-15/index.html

The only non-radiation test affected is Detonation-collision-retry-SDC. It's not surprising that at least some tests might be affected since there are non-State_Type variables that have ghost zones such as the Simplified SDC reactions type.

zingale (approved):

PR #1346, opened 2020-11-15
Convert double_mach_reflection BCs to C++

maxpkatz:
## PR summary

In order to avoid conflicting with the HSE BCs, the inputs files are set to outflow, but the outflow BCs are overridden by the problem-specific BC.

## PR checklist


zingale:
The old BC integrated the initial model analytic profile into the ghost cells.  For an isentropic atmosphere, this used the proper isentropic profile.  Now we interpolate T into the ghost cells and integrate p via HSE.  So it is not exactly the same as the analytic profile.  I think it is good enough for a general HSE BC.

In the future, we could have the general HSE BC allow for an isentropic profile in the ghost cells too.  That actually would not be too hard to add.

zingale:
the fcompare fails because the grids are ordered differently, but if we manually compare while letting the grids be different, we get:
```
 ../../../amrex/Tools/Plotfile/fcompare.gnu.ex -a double_mach_reflection_plt00100 ../../../Castro-SBU-benchmarks/double_mach_reflection_plt00100/

            variable name            absolute error            relative error
                                        (||A - B||)         (||A - B||/||A||)
 ----------------------------------------------------------------------------
 level = 0
 density                                          0                         0
 xmom                                             0                         0
 ymom                                             0                         0
 zmom                                             0                         0
 rho_E                                            0                         0
 rho_e                                            0                         0
 Temp                                             0                         0
 rho_X                                            0                         0
 pressure                                         0                         0
```

zingale (approved):

PR #1352, opened 2020-11-16
Convert get_omega to C++

maxpkatz:
## PR summary

get_omega is ported to C++ as an inline function. All current uses of get_omega in C++ (which were calling into Fortran) are now  replaced with the new implementation.

Since no part of the rotation simulation capability remains in Fortran, the interface to Fortran is removed. The Fortran version of get_omega is left for now because two problem setups in Fortran still depend on it.

(This change does cause a slight issue for wdmerger, which modifies rotational_period in probinit. However the impact is minor and should have no outcome on the simulation result, and at any rate will be fixed when its initialization is converted to C++.)

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-16-002/index.html

maxpkatz:
I've fixed the wdmerger_2D issue (that problem updates rot_axis, so we need to sync that to C++ too).

zingale:
all tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-16-003/index.html

zingale (approved):

PR #1356, opened 2020-11-17
Remove ext_denfill

maxpkatz:
## PR summary

As discussed in #154 and several other issues, we have been using ext_denfill incorrectly for a long time. When we come into the BC routine with a single component, it is usually not the density, usually it is a derive and it may be any quantity, so filling with the density is simply wrong. So the easy solution here is just to remove the application of ext_denfill. This should not have very large effects, as we were only getting it wrong during derives, whose only effect on the simulation is through tagging (and this error would only have occurred at physical domain boundaries). We already fill the data with FOEXTRAP and this should be good enough for most cases. (This does not preclude future efforts to actually do the BCs on those drives correctly, taking into account HSE and problem-specific BCs.)

I've also used this opportunity to do a small bit of cleanup to GpuBndryFuncFab, and to remove the redundant call to CpuBndryFuncFab (the former should do the right thing even if not in a launch region, and in any case we do not ever turn off the launch region).

## PR checklist


zingale (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-17-006/index.html

anything running the HSE BC crashes.

maxpkatz:
This is because I confused myself and thought we were already dealing with the inflow BCs prior to the call. This is now fixed, I believe.

zingale:
tests look good: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-17-007/index.html
only magvort and divu really change here for problems that did inflow or HSE.

PR #1364, opened 2020-11-18
Convert kepler_third_law and freefall_velocity to C

maxpkatz:
## PR checklist

### DIFF
--- a/Exec/science/wdmerger/wdmerger_util.cpp
+++ b/Exec/science/wdmerger/wdmerger_util.cpp
@@ -1,0 +1,71 @@
+#include <wdmerger_util.H>
+
+extern "C" {
+
+// Given total mass of a binary system and the initial separation of
+// two point particles, obtain the velocity at this separation
+// assuming the point masses fell in from infinity. This will
+// be the velocity in the frame where the center of mass is stationary.
+
+void freefall_velocity (Real mass, Real distance, Real& vel)
+{
+    vel = std::sqrt(2.0_rt * C::Gconst * mass / distance);
+}
+
+// Accepts the masses of two stars (in solar masses)
+// and the orbital period of a system,
+// and returns the semimajor axis of the orbit (in cm),
+// as well as the distances a_1 and a_2 from the center of mass.
+
+void kepler_third_law (Real radius_1, Real mass_1, Real radius_2, Real mass_2,
+                       Real& period, Real eccentricity, Real phi, Real& a,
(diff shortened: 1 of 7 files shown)


Exec/science/wdmerger/wdmerger_util.cpp
```diff
@@ -0,0 +1,71 @@
...
+    // First, solve for the orbit in the reduced one-body problem, where
+    // an object of mass mu orbits an object with mass M located at r = 0.
+    // For this we follow Carroll and Ostlie, Chapter 2, but many texts discuss this.
+    // Note that we use the convention that phi measures angle from aphelion,
+    // which is opposite to the convention they use.
+
+    if (period > 0.0_rt && a < 0.0_rt)
+    {
+        a = std::pow(C::Gconst * M * period * period / (4.0_rt * M_PI * M_PI), 3); // C + O, Equation 2.37
+    }
```
zingale, line 38: shouldn't this be to the 1/3 power? not 3?

Exec/science/wdmerger/wdmerger_util.cpp
```diff
@@ -0,0 +1,71 @@
...
+    r_2  =  (mu / mass_2) * r;
+
+    v_1r = -(mu / mass_1) * v_r;
+    v_2r =  (mu / mass_2) * v_r;
+
+    v_1p = -(mu / mass_1) * v_phi;
+    v_2p =  (mu / mass_2) * v_phi;
+}
+
+}
```
zingale, line 71: there were some additional bits in the Fortran version that checked to make sure that the domain was big enough to hold the stars.  Did you intend for that to be cut out?

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-18-002/index.html

Exec/science/wdmerger/wdmerger_util.cpp
maxpkatz: Good catch, fixed

Exec/science/wdmerger/wdmerger_util.cpp
maxpkatz: I do like that check, however the problem is that it depended on access to `physbc_lo` which is available as a 'global' in Fortran but not currently in C++. So I punted on that for now, and moved that check to the calling routine of kepler_third_law in Fortran.

zingale (approved):

PR #1367, opened 2020-11-18
Convert compute_temp_from_rhoe to C++

maxpkatz:
## PR summary

Note: in some of these cases we come in with an array called "temp" which actually contains (rho*e), and then overwrite it with the actual temperature. So the code can look a little confusing. In some cases I was able to merge this with a previous loop so the result is clearer. I chose to write each loop out explicitly rather than keep this in a separate function because I suspect that we will be able to fuse even more loops as time goes on, and in some cases we should eventually be able to use the main compute_temp function.

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-18-005/index.html

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-18-005/index.html

zingale (approved):

PR #1374, opened 2020-11-20
Convert Rosseland opacity to C++

maxpkatz:
## PR checklist

### DIFF
--- a/Source/radiation/MGFLD.cpp
+++ b/Source/radiation/MGFLD.cpp
@@ -296,8 +296,10 @@ void Radiation::eos_opacity_emissivity(const MultiFab& S_new,
 
       bool use_dkdT_loc = use_dkdT;
 
-      GpuArray<Real, NGROUPS> nugroup = {0.0};
-      ca_get_nugroup(nugroup.begin());
+      GpuArray<Real, NGROUPS> nugroup_loc;
+      for (int g = 0; g < NGROUPS; ++g) {
+          nugroup_loc[g] = nugroup[g];
+      }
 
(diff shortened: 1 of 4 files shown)


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-20-005/index.html

Source/radiation/MGFLD.cpp
```diff
@@ -925,30 +927,78 @@ void Radiation::MGFLD_compute_rosseland(FArrayBox& kappa_r, const FArrayBox& sta
...
+      Real kp, kr;
+      bool comp_kp = false;
+      bool comp_kr = true;
+
+      for (int g = 0; g < NGROUPS; ++g) {
+          opacity(kp, kr, rho, temp, Ye, nugroup_loc[g], comp_kp, comp_kr);
+          kpr(i,j,k,g) = kr;
+      }
+  });
+  Gpu::synchronize();
```
zingale, line 959: why does this need a Gpu::synchronize?

zingale: oh I see -- it was there previously

zingale (approved):

Source/radiation/MGFLD.cpp
maxpkatz: It was there previously because this is the Fab interface, not the MultiFab interface, and when you only have the Fab at function scope you don't know if it's being used in a fashion safe for asynchrony. This is a pain point I want to clean up after I get everything converted to C++.

PR #1379, opened 2020-11-20
Abort if mass fraction is unphysical

maxpkatz:
## PR checklist


zingale:
tests:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-20-008/index.html
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-20-009/index.html

zingale (approved):

maxpkatz:
Detonation-collision fails because it interpolates the burning on the fine level and that does not natively enforce 0 < X < 1.

zingale:
I'll retest just the Detonations.

zingale:
actually, I think we need `development` merged in first.

maxpkatz:
Marking as draft until I figure out what to do about the reactions.max_solve_level interpolation issue.

zingale:
do we still want to do this?

maxpkatz:
Definitely. I just didn't want to break my science simulations. But since I'm not currently running those right now, maybe we should just merge this and solve that later.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-03-03-005/index.html

The SDC detonations fail -- I suspect that we are missing a normalization in that code.

zingale:
in particular, these failures are all using VODE for the SDC solver, and we don't seem to have any normalization in the RHS there.  For the SDC Newton solve, we do have the normalization.

PR #1382, opened 2020-11-21
Error if density is exactly zero

### LINKED ISSUE #1345: Restore check on density exactly equal to zero
If the density is exactly equal to zero then that usually indicates a code bug like data is not being properly filled when constructing new grids. We used to have an explicit check on this somewhere, but we don't now, so the issue in #1342 and #1344 was only revealed through a divide by zero in enforce_minimum_density caught by amrex.fpe_trap_invalid. We should probably restore an explicit check on that.

maxpkatz:
## PR summary

Closes #1345.

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-21-001/index.html

zingale (approved):

PR #1388, opened 2020-11-21
Hardcode the maximum number of points in the wdmerger initial models

maxpkatz:
## PR checklist

### DIFF
--- a/Exec/science/wdmerger/initial_model.F90
+++ b/Exec/science/wdmerger/initial_model.F90
@@ -12,6 +12,8 @@ module initial_model_module
   use fundamental_constants_module, only: Gconst, M_solar
   use meth_params_module, only: small_temp
 
+  integer, parameter :: initial_model_max_npts = 10000
+
   type :: initial_model
 
      ! Physical characteristics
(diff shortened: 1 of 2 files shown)


Exec/science/wdmerger/initial_model.F90
```diff
@@ -12,6 +12,8 @@ module initial_model_module
   use fundamental_constants_module, only: Gconst, M_solar
   use meth_params_module, only: small_temp
 
+  integer, parameter :: initial_model_max_npts = 10000
+
```
zingale, line 16: why not use `MAX_NPTS_MODEL` that is defined by the preprocessor?

maxpkatz: For now this maximum is arbitrary and isn't necessarily the same as the one in the model parser module (although I happened to use the same number). Longer term I do want to get this switched over to the model parser module, but we'd have to address the fact that right now it only has a single model (while I need to build two).

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-21-012/index.html

zingale (approved):

PR #1403, opened 2020-11-25
Separate out initialization routines into separate, consistent headers

maxpkatz:
## PR summary

We've been moving in the direction where each problem-related header has exactly one routine and the name of that routine is the same as the header file. This PR makes that consistent for the problem setup routines as well.

## PR motivation

We've been burned in the past by wanting to make changes to problem-facing source files that contain multiple functions and thus causing larger breakage than we intended. With one function per header, we limit the scope of the problem if we do have to change a function interface later, and we also make the source more predictable to understand.

## PR checklist


zingale:
all tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-25-001/index.html

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-25-005/index.html

zingale (approved):

PR #1407, opened 2020-11-26
Convert wdmerger to C++

maxpkatz:
## PR checklist


maxpkatz:
Note: for wdmerger_2D, at t = 0 the profile looks correct, but the rotation field differs. The 2D setup alters castro::rot_axis, but on development this is done incorrectly because it's only altered in C++ after the rotation field in StateData is initialized. This is an artifact of when we moved rotation from Fortran to C++. So this causes a small diff because the wdmerger update does depend on the rotation field, but it is more correct now.

zingale:
tests look reasonable: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-27-002/index.html

zingale (approved):

PR #1410, opened 2020-11-28
convert xrb_mixed to C++

zingale:
## PR checklist


Exec/science/xrb_mixed/problem_initialize.H
```diff
@@ -0,0 +1,43 @@
...
+
+    if (ParallelDescriptor::IOProcessor()) {
+        for (int i = 0; i < model::npts; i++) {
+            std::cout << i << " " << model::profile(0).r(i) << " " << model::profile(0).state(i,model::idens) << std::endl;
+        }
+    }
+
+    // velocity perturbation stuff
+
+    Real offset = (probhi[0] - problo[0]) / problem::num_vortices;
```
maxpkatz, line 36: Is the change here with respect to the Fortran intentional?

zingale: no, it wasn't.  I just copied this vortex initialization from another problem.  And looking at it now, I'm not sure which we'd prefer.  I think this new method makes more sense with periodic BCs, because then they are evenly spaced.

maxpkatz (approved):

PR #1420, opened 2020-11-30
Convert hbmat to C++

maxpkatz:
## PR checklist


Source/radiation/HypreABec.cpp
```diff
@@ -421,6 +421,162 @@ void HypreABec::hbcoef (const Box& bx,
...
+        else {
+            amrex::Error("Unknown cdir");
+        }
+
+    }
+    else if (AMREX_SPACEDIM == 2) {
+
+        if (cdir == 0) {
+            xlo = true;
+            h = dx[0];
```
zingale, line 460: this is atleast the 3rd time we've had a PR with this logic in it.  Maybe we can extract this bit to an inline function?

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-11-30-002/index.html

zingale (approved):

PR #1436, opened 2020-12-03
Move wdmerger subsampled interpolation scheme into model_parser header

maxpkatz:
## PR checklist


zingale:
TDE fails: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-12-03-004/index.html

maxpkatz:
The TDE setup is challenging because the coarse grid dx is larger than the radius of the WD. I had a hack in there previously that detected this and just inserted the star center density on the grid. The goal was that we'd be using AMR and we just needed to have some mass on the grid to trigger refinement, which would then overwrite this crude guess. But we can also achieve this effect by just using a high subsampling factor, which also has the benefit that the mass on the coarse grid should be approximately correct even without refinement. So let's reset the benchmark after this merge.

zingale (approved):

PR #1443, opened 2020-12-05
convert radbreakout to C++

zingale:
this also extends the C++ model parser to handle aux and vel data

## PR checklist


zingale:
depends on #1449

zingale:
results look good: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-12-07-006/rad-breakout.html

zingale:
note: this removes some unused initial models.  Since we don't know how these were generated, I didn't think we should keep them around.

zingale:
previous tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-12-07-006/index.html
RadSuOlson compilation is now fixed:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-12-07-007/index.html

Source/driver/Castro.cpp
```diff
@@ -1407,12 +1407,19 @@ Castro::initData ()
           auto r = Rad_new[mfi].array();
           auto geomdata = geom.data();
 
+          GpuArray<Real, NGROUPS+1> xnu = {0.0};
+#if NGROUPS > 1
+          ca_get_xnu(xnu.begin());
```
maxpkatz, line 1412: Let's avoid the proliferation of this Fortran API please. `xnu` is already available as an array in C++, it's just a class member so we can't pass it directly to the ParallelFor but instead need to make a local copy of it. https://github.com/AMReX-Astro/Castro/blob/main/Source/radiation/MGFLD.cpp#L326-L329

zingale: okay, I switched this, by making the radiation group info public, and then making a copy before passing into the rad init

maxpkatz (approved):

Issue #113, opened 2017-02-28
revisit problem_post hooks and document


zingale:
Currently there are 4 ifdefs for the problem post hooks:

* `DO_PROBLEM_POST_TIMESTEP`
* `DO_PROBLEM_POST_RESTART`
* `DO_PROBLEM_POST_INIT`
* `DO_PROBLEM_POST_SIMULATION`

Only the last of these is controlled by a make option (`USE_POST_SIM`).

Usually the problem-specific implementation of these is in `Prob.cpp`.

We should revisit whether we really want these ifdef-ed out or if we just have simple stubs always defined.  For the latter, we might need to update some problem setups.

zingale:
We also need to document this functionality

PR #1452, opened 2020-12-08
Rework diagnostic sums to be more customizable

maxpkatz:
## PR summary

The current diagnostic sums functionality (`castro.sum_interval = 1`) has two limitations. First, customizing it for a problem setup currently means that the whole sum_integrated_quantities.cpp file has to be copied, which means that if the functionality from the main sum_integrated_quantities routine is desired, the problem has to duplicate that code. Second, even if you did customize it, the Amr class ownership of the data logs means that customizing the number and name of the logs is brittle. One unfortunate way that brittleness plays out is that you must specify `amr.data_log` in the inputs file with the exact ordering of the diagnostic names.

The intent of this change is that problems should have not have to override sum_integrated_quantities.cpp to get custom diagnostics. A new hook is provided, `problem_diagnostics.H`, which is called at the end of `sum_integrated_quantities()` for problem-specific diagnostics. To support this change, we stop depending on the Amr ownership of the data logs. Two new Castro class members are added, `data_logs` and `problem_data_logs`. The former is managed by Castro and can be changed over time if we want to add more diagnostic routines, with initialization done in `Castro::read_params()`. The latter is managed by the problem itself, with initialization done in `problem_initialize()`. The log names are coded directly into the source, which means that you don't need to remember to put those names in your inputs file.

test_convect and toy_convect are converted to the new scheme.

## PR checklist

### DIFF
--- a/Exec/hydro_tests/test_convect/inputs_2d
+++ b/Exec/hydro_tests/test_convect/inputs_2d
@@ -68,6 +68,3 @@ amr.derive_plot_vars = ALL
 
 #PROBIN FILENAME
 amr.probin_file = probin
-
-# DATA LOG
-amr.data_log = log
--- a/Exec/hydro_tests/test_convect/problem_diagnostics.H
+++ b/Exec/hydro_tests/test_convect/problem_diagnostics.H
@@ -1,14 +1,10 @@
-#include <iomanip>
+#ifndef problem_diagnostics_H
+#define problem_diagnostics_H
 
-#include <Castro.H>
-#include <Castro_F.H>
-
-using namespace amrex;
-
+AMREX_INLINE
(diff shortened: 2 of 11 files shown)


maxpkatz:
I am leaving this undocumented for now. I do want to document all these problem hooks (#113) but I'm about to rework the old ones to be consistent with the new convention we've been using, and between that and the de-Fortran-ification we're currently doing, I'd rather wait until the dust settles and we're happy with the layout before doing the final documentation.

zingale:
tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-12-08-002/index.html

zingale (approved):

PR #1454, opened 2020-12-08
Move wdmerger initial model constructor to model_parser

maxpkatz:
## PR summary

The isothermal initial model generator is moved from wdmerger into the C++ model parser. This will allow it to be reused multiple times in the coming split of wdmerger into separate problem setups for mergers, collisions, and TDEs. At some point we can also merge this with the initial model generator for the setups like flame_wave.

## PR checklist


Util/model_parser_cxx/model_parser.H
```diff
@@ -202,6 +205,270 @@ Real interpolate_3d (const Real* loc, const Real* dx, int var_index, int nsub =
...
+
+    if (!mass_converged && max_mass_iter > 1) {
+        amrex::Error("ERROR: WD mass did not converge.");
+    }
+
+    central_density_want = model.state(0, model::idens);
+    mass_want = mass;
+
+    model::initialized = true;
+    model::npts = NPTS_MODEL;
```
maxpkatz, line 469: Note: I felt that if we were already specifying NPTS_MODEL at compile time that including an explicit npts argument to this routine is over-specifying the problem when we already have dx. The model points exist anyway, so we might as well just fill them all with fluff instead of arbitrarily cutting the model off after some number of points.

Util/model_parser_cxx/model_parser_data.H
```diff
@@ -26,6 +34,17 @@ namespace model
         amrex::Array1D<amrex::Real, 0, NPTS_MODEL-1> r;
     };
 
+    // Tolerance used for getting the total star mass equal to the desired mass.
+    // It can be reasonably small, since there will always be a central density
+    // value that can give the desired mass on the grid.
+
+    const amrex::Real mass_tol = 1.0e-6_rt;
```
maxpkatz, line 41: I never changed these in practice, so I just hardcoded them here. But we could make these runtime parameters later if we need to.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-12-08-004/index.html
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-12-08-003/index.html

Util/model_parser_cxx/model_parser.H
zingale: I used to rely on knowing the actual number of points, to play some tricks in the BCs (I'd like to know where the actual edge of the model is).

zingale (approved):

PR #1470, opened 2020-12-13
convert evrard_collapse to C++

zingale:
## PR checklist


maxpkatz (approved):

zingale:
the errors are not small: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-12-13/evrard_collapse.html

zingale:
I wonder if the issue here is that we are not setting small_pres and small_ener in C++, whereas before that info was never propagated to C++.

zingale:
actually the issue is that this problem is just really sensitive.  Just tweaked some floating point operations.

zingale:
errors are much smaller now: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-12-13-005/evrard_collapse.html

PR #1497, opened 2020-12-18
Move set_wd_composition into model_parser

maxpkatz:
## PR summary

Copy the Dan et al. 2012 prescription for WD composition profiles to the main model_parser code.

## PR motivation

This is part of the goal to move as much of wdmerger as possible into the mainline code so that the problem can more easily be split up.

## PR checklist


Source/problems/_default_prob_params
```diff
@@ -1 +1,25 @@
 center                          real                  0.0e0_rt                 n                      3
+
```
zingale, line 2: why do we want these defined for all problems?  isn't there a better way to separate it out?

maxpkatz: Yes, I suppose the more natural thing would be that they only get added in if you use the model parser module, but we don't really have the infrastructure for that do we? I thought there could only be one _prob_params file, and this _default_prob_params file is something I hacked in that allows us to do things like this. If _prob_params was cumulative like the extern _params is, this would work.

zingale: ah indeed.  I think we can make it cumulative.  I'd prefer that.

maxpkatz: Due to #1500, these parameters could be moved to the model parser directory, so I have now done so.

PR #1500, opened 2020-12-20
allow for multiple _prob_param files

zingale:
we now search all of the Castro source directories
the problem's own parameters are always parsed last.
We also get rid of _default_prob_params in Source/problems and just treat it like
any other _prob_params file.

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-12-20-002/index.html
and then: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2020-12-20-003/index.html

zingale:
this also starts some cleaning up of the python script to use f-strings

Exec/Make.auto_source
```diff
@@ -47,8 +47,12 @@ CEXE_headers += prob_parameters.H
 $(CASTRO_AUTO_SOURCE_DIR)/prob_parameters.cpp: $(CASTRO_AUTO_SOURCE_DIR)/prob_params_auto.F90
 
 PROB_PARAMS_TEMPLATE := $(TOP)/Util/scripts/prob_params.template
-PROB_PARAMETERS := $(PROBLEM_DIR)/_prob_params
-DEFAULT_PROB_PARAMETERS := $(TOP)/Source/problems/_default_prob_params
+# search Bdirs for _prob_params files
+PROB_SEARCH := $(foreach dir, $(Bdirs), $(TOP)/$(dir))
+PROB_PARAMETERS := $(shell $(TOP)/Util/scripts/findparams.py $(PROB_SEARCH))
```
maxpkatz, line 52: Can't you also do this in GNU make with a wildcard on $(Bdirs)/_prob_params, instead of writing a new script?

zingale: ah indeed, I think maybe that could work.  Let me try.

zingale: okay, that's implemented.  Sedov compiles okay with the changes

maxpkatz (approved):

PR #1531, opened 2021-01-12
move the call to init_prob_parameters() to after the Fortran read

zingale:
## PR checklist


Source/driver/Castro.cpp
```diff
@@ -598,9 +598,9 @@ Castro::Castro (Amr&            papa,
 
     buildMetrics();
 
-    // initialize the C++ values of the runtime parameters
     if (do_init_probparams == 0) {
```
maxpkatz, line 601: Well now this variable name needs to be updated.

zingale: done

zingale:
this doesn't work properly if we modify one of the _prob_params values in the Fortran probinit,

in the current setup, we have:
* fortran initializes
* fortran probinit
* copy from Fortran to C++
* C++ probinit
* copy from C++ to Fortran
* ...
* initialize the state date

but now are proposing to do:
* fortran initializes
* copy from Fortran to C++
* fortran probinit
* C++ probinit
* copy from C++ to Fortran
* ...
* initialize the state date

so if the fortran probinit changes the value of a variable in the `_prob_init`, then it is not changed in C++ and we overwrite it when we copy from C++ to Fortran.

zingale:
now that we don't have any Fortran probinit that can change values, I think that this PR is okay.  This will then support reading the problem parameters in C++ directly (#1532)

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-22-006/index.html

maxpkatz (approved):

PR #1538, opened 2021-01-13
some fixes to get this running again

zingale:
note: it does not seem to get the timescales correct

## PR checklist


zingale:
it would be good to check an older version of Castro to see how the problem looks

maxpkatz:
This is covered by #1534

PR #1541, opened 2021-01-16
AMReX now calls HYPRE_init() so we don't have to

maxpkatz:
## PR summary

We also update to use the AMReX GNU make package for Hypre instead of doing it on our own.

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-01-16-001/index.html

zingale (approved):

Issue #1548, opened 2021-01-22
flame_wave dies on restart


zingale:
This was seen on Summit on the GPU, but this can be reproduced on groot.  Build `flame_wave` as:
```
make -j 20 USE_CUDA=TRUE
```
then run as:
```
./Castro2d.gnu.MPI.CUDA.ex inputs_2d.testsuite max_step=200
```
then restart as:
```
./Castro2d.gnu.MPI.CUDA.ex inputs_2d.testsuite max_step=200 amr.restart=chk00100
```

It will die with `too many subcycles`

zingale:
my theory is that this is somehow related to the ambient BCs at the upper boundary.

zingale:
If we run on CPU, then it crashes immediately upon restart with:
```
amrex::Error::0::Density exactly zero at -4, 64, 0 !!!
```

with a backtrace:
```
 0: ./Castro2d.gnu.MPI.ex() [0x62f7f6]
    amrex::BLBackTrace::print_backtrace_info(_IO_FILE*)
/home/zingale/development/amrex//Src/Base/AMReX_BLBackTrace.cpp:179:39

 1: ./Castro2d.gnu.MPI.ex() [0x63196d]
    amrex::BLBackTrace::handler(int)
/home/zingale/development/amrex//Src/Base/AMReX_BLBackTrace.cpp:85:8

 2: ./Castro2d.gnu.MPI.ex() [0x4c9ff5]
    std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_data() const inlined at /usr/include/c++/10/bits/basic_string.h:
222:23 in Castro::do_enforce_minimum_density(amrex::Box const&, amrex::Array4<double> const&, int)
/usr/include/c++/10/bits/basic_string.h:187:28
std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_is_local() const
/usr/include/c++/10/bits/basic_string.h:222:23
std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_dispose()
/usr/include/c++/10/bits/basic_string.h:231:18
std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::~basic_string()
/usr/include/c++/10/bits/basic_string.h:658:19
operator()
../../../Source/hydro/advection_util.cpp:1105:78
[...]

PR #1549, opened 2021-01-22
Move C++ probinit and prob params to buildMetrics

maxpkatz:
## PR summary

Currently in C++, problem_initialize() and the read of the C++ parameters is not happening on a restart (as well as the setting of the C++ ambient data, the proximal cause for the issue in #1548). Moving to buildMetrics() fixes this, as it is also done on a restart.

## PR checklist

### DIFF
--- a/Source/driver/Castro.cpp
+++ b/Source/driver/Castro.cpp
@@ -598,32 +598,6 @@ Castro::Castro (Amr&            papa,
 
     buildMetrics();
 
-    // initialize the C++ values of the runtime parameters
-    if (do_init_probparams == 0) {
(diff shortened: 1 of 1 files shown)


maxpkatz:
Note: in Fortran when we called probinit, we had a parameter which told us whether we were doing it on a restart (which allowed you to skip certain unnecessary calculations on a restart). We don't currently have this in C++, but could add it if needed.

maxpkatz:
Note 2: this infrastructure is fragile. We should strive in the future to make it less likely to have a feature gap between the initial constructor and the restart code.

zingale (approved):

PR #1556, opened 2021-01-25
fix the error checking on the axisymmetric geom term

zingale:
## PR summary

this also adds an abort if we try to use the new geometry term with state_in_rotating_frame=0

## PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,6 @@
 
+# 21.02
+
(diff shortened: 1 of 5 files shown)


zingale:
this fixes the test suite:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-01-25-001/index.html

the 3 failures are expected, because that's where the term kick in

maxpkatz (approved):

PR #1559, opened 2021-01-30
Use eos_re_t in places where it is sufficient

maxpkatz:
## PR summary

In several places like `computeTemp()`, we only need (rho, T, e) in the EOS state, we do not need pressure (or other fields like entropy and enthalpy). For those cases we switch to the thinner state, which allows equations of state to skip calculating those unused fields.

## PR checklist


zingale:
this fails to build on GPUs:
http://groot.astro.sunysb.edu/Castro/test-suite/gpu/2021-01-30-001/flame_wave.html

not sure why this is not an issue on CPUs...

zingale:
this failure was due to #1560

zingale (approved):

PR #1564, opened 2021-02-02
Perform a more efficient reduction in multipole BCs

maxpkatz:
## PR summary

The multipole reduction works by summing a quantity from every zone to a set of arrays in a weighted linear combination. This is presently implemented with fully atomic reductions on GPUs. The overhead from atomic serialization becomes very large with hundreds of thousands of threads updating the same quantity simultaneously. We can improve on this with reductions that do some of the work within a threadblock using `amrex::Gpu::deviceReduceSum()`. The main changes that are required are:
- Every thread must be doing the same reduction at the same time, so we have to move the reduction outside of the conditional that checks whether we're contributing to the upper or lower multiple arrays
- We must use the four-argument variant of the device lambda which provides a `amrex::Gpu::Handler` to help implement the reduction
- The `ParallelFor` must be launched with `setReduction(true)` which indicates to AMReX the relevant information needed to perform this reduction safely (we will launch more threads than there are valid grid zones and we need to take special care in the boxes where not every active thread is in a zone, to avoid possible deadlocks)

For a 128^3 grid, this improves the performance of this kernel by a factor of 5.

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-02-03-004/index.html

harpolea (approved):

PR #1583, opened 2021-03-05
some normalization efforts for true SDC

zingale:
## PR checklist


zingale:
here are the tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-03-07-003/index.html

I think they look good.  In particular, this fixes some crashes with the detonation that arose when we added the abort for the mass fraction non-conservation.  I think that this is much more robust now.

maxpkatz:
If I understand this correctly, it's basically doing what the code was doing before we merged #1379, except it's being handled closer to where the negative mass fraction is being generated instead of letting the generic clean_state do it for you. The concern I have is that the point of #1379 is that if you got to the point where we would be aborting for negative X, it's an indication that your algorithm actually did something really bad, and papering it over with a floor/normalization is not the right thing to do. If VODE is generating a negative X (for example), the point of #1379 is that we should be fixing VODE not to do that, instead of just shifting where the hiding of the negative X gets inserted.

zingale:
I mostly agree.  In the Strang stuff, we do a normalization at the end of the solve as well, so this is consistent with that.

But I think that what we also need is a TRUE_SDC code path in VODE that does the `valid_update` check on the partial densities.  I think that together with this will solve a lot of the issues.

maxpkatz:
Yes, I would be OK with this if we also did the proposed VODE fix.

zingale:
I think that most of these are now in #2586

PR #1585, opened 2021-03-06
remove an unneeded clean_state in true SDC advance

zingale:
we do a clean_state at the top of the node loop, so we don't
need to do it at the very end too.  Plus, we were not calling
it on the right MF at the end anyway...

## PR checklist

### DIFF
--- a/Source/driver/Castro_advance_sdc.cpp
+++ b/Source/driver/Castro_advance_sdc.cpp
@@ -220,9 +220,6 @@ Castro::do_advance_sdc (Real time,
 
       do_sdc_update(m, m+1, dt); //(dt_sdc[m+1] - dt_sdc[m])*dt);
 
-      // we now have a new value of k_new[m+1], do a clean_state on it
-      clean_state(S_new, cur_time, 0);
-
     }
 
 
--- a/Source/driver/Castro_advance_sdc.cpp
+++ b/Source/driver/Castro_advance_sdc.cpp
@@ -220,6 +220,9 @@ Castro::do_advance_sdc (Real time,
 
       do_sdc_update(m, m+1, dt); //(dt_sdc[m+1] - dt_sdc[m])*dt);
 
+      // we now have a new value of k_new[m+1], do a clean_state on it
+      clean_state(*(k_new[m+1]), cur_time, 0);
+
     }
 
 


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-03-06/index.html

the detonation diffs are big, but that problem is so nonlinear...

I think that we are just seeing some EOS tolerance differences by calling clean_state on the wrong state twice.  In effect, we were calling it on the initial state a second time instead of on the state at the new time node.

maxpkatz:
The philosophy of the code elsewhere has been to call clean_state immediately after an update that changes the state, rather than immediately before one thinks it is needed, because (to me) the latter seems less error-prone. Is it possible to do this here?

maxpkatz:
And if this clean_state were operating on the correct state, what would it have looked like?

zingale:
it would operate on `*(k_new[m+1])`

zingale:
the point is that we do the `clean_state()` on:

```
 77     MultiFab::Copy(S_new, *(k_new[m]), 0, 0, S_new.nComp(), 0);             
 78     clean_state(S_new, cur_time, 0);                                        
 79     expand_state(Sborder, cur_time, NUM_GROW); 
```
so we will always do that, since we need `Sborder` to have valid ghost cells.

maxpkatz:
Understood. My preference would be to do a perhaps unnecessary clean_state on the correct data immediately after the update. When we get the "reference counting" system in we can have a true fix to this.

zingale:
okay, that's fine.  I'll fix the `clean_state` in this PR and retest.

zingale:
switch back to doing the clean_state on the proper MF.  Here's the tests:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-03-06-001/index.html

maxpkatz (approved):

zingale:
this change still gives the expects convergence for `bubble_convergence`:

```
 $\rho$                           3.585788e+15   3.265       3.729242e+14   3.727          2.817e+13
 $\rho u$                         1.118655e+24   3.796       8.055983e+22   3.961       5.174663e+21
 $\rho v$                         1.312309e+24   3.544       1.124968e+23   3.852       7.792148e+21
 $\rho E$                          3.69012e+32   2.949       4.780079e+31   3.669       3.757999e+30
 $\rho e$                         3.689677e+32   2.949       4.779642e+31   3.669       3.757798e+30
 $T$                              1.437985e+18   3.508       1.264001e+17   3.831       8.884047e+15
 $\rho X(\isotm{He}{4})$          3.584552e+15   3.268       3.722151e+14   3.725       2.814821e+13
 $\rho X(\isotm{C}{12})$          1.519908e+13   2.544       2.605549e+12   3.797       1.874237e+11
 $\rho X(\isotm{O}{16})$              35844210   3.264            3731950   3.728             281580
 $\rho X(\isotm{Fe}{56})$             35856870   3.265            3729094   3.727             281700
```

(some diffs in the 2nd decimal place compared to the original paper).

PR #1609, opened 2021-03-26
don't do the clean_state on S_new after applying the advective update

zingale:
that state is incomplete -- it had a reactive predictor, but we didn't apply the reactive
state to it

## PR checklist


zingale:
I don't think this matters anymore.  The issue I was trying to fix was fixed in Microphysics (https://github.com/starkiller-astro/Microphysics/pull/643)

PR #1610, opened 2021-03-30
Fix issue where Castro was using uninitialized tracer particle parameters

dwillcox:
`CastroParticles.cpp` was previously using an unitialized, unnamed namespace in that file to hold tracer particle parameters that were also defined in the `particles` namespace defined with `_cpp_parameters` and included through:

 ```
#include <particles_params.H>
#include <particles_declares.H>
```

As a result, the code inside this source file was using the uninitialized parameters in the unnamed namespace instead of the parameters by the same names in the `particles` namespace.

This removes the duplicate parameters from the unnamed namespace and adds the `particles::` namespace specifier where necessary.

## PR checklist

### DIFF
--- a/Source/particles/CastroParticles.cpp
+++ b/Source/particles/CastroParticles.cpp
@@ -16,11 +16,6 @@ using namespace amrex;
 AmrTracerParticleContainer* Castro::TracerPC =  0;
 
 namespace {
-    std::string       particle_init_file;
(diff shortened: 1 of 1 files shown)


dwillcox:
Will be useful for @nmford20 to develop some particle analysis :)

zingale (approved):

PR #1622, opened 2021-04-14
move the passive upwinding out of the core Riemann solvers

zingale:
## PR summary

All the approximate state solvers do the same thing, so we move this out to the caller.  Soon this will be merged with a call to the solver over just a single interface.

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-04-15-003/index.html

PR #1625, opened 2021-04-14
compute dpde and dpdr_e manually as needed

zingale:
this is in prep for removing them from the eos types

## PR checklist


zingale:
we decided against this for now -- not all EOSes permit the same way of computing these quantities

PR #1658, opened 2021-05-05
convert FastRarefaction to C++

zingale:
## PR checklist


zingale:
this gives identical answers to the Fortran

Exec/mhd_tests/FastRarefaction/problem_initialize_mhd_data.H
```diff
@@ -0,0 +1,99 @@
...
+#if AMREX_SPACEDIM == 3
+    z = problo[2] + dx[2] * (static_cast<Real>(k) + 0.5_rt);
+#endif
+
+
+    if (problem::idir == 1) {
+        // initialize along the x-direction
+
+        if (ibdir == 0) {
+            if (x <= problem::split[0]+1) {
```
maxpkatz, line 31: Can you explain where the +1 comes from here?

zingale: oh, I think that is a copy-paste from a different problem
I'll fix it

zingale: oh wait... indeed that was in the mag intiialization originally, but not in the hydro state.  I am not sure why that is there.

zingale: it looks like it is only in the problem direction that we do that.

maxpkatz: yeah, DaiWoodward did that too, that's weird but whatever.

maxpkatz:
Looks like you also need to rm the prob_mhd_nd.f90

zingale:
done

maxpkatz (approved):

PR #1661, opened 2021-05-06
port species MHD problem to C++

zingale:
## PR checklist


zingale:
this has 0 diffs with the Fortran version

Exec/mhd_tests/species/problem_initialize_state_data.H
```diff
@@ -0,0 +1,49 @@
...
+    state(i,j,k,UMY) = problem::rho * problem::u_y;
+    state(i,j,k,UMZ) = problem::rho * problem::u_z;
+    state(i,j,k,UEDEN) = problem::rhoe +
+        0.5_rt * problem::rho * (problem::u_x * problem::u_x +
+                                 problem::u_y * problem::u_y +
+                                 problem::u_z * problem::u_z);
+    state(i,j,k,UEINT) = problem::rhoe;
+    state(i,j,k,UTEMP) = problem::T;
+
+    Real r2 = (std::pow(x - 0.5_rt, 2) +
```
maxpkatz, line 39: Looks like this offset by 0.5 would be wrong in 1D/2D, we might want to fix this test by setting center to 0.5 in all of the simulated dimensions (only) and then doing this offset with respect to center.

maxpkatz (approved):

Exec/mhd_tests/species/problem_initialize_state_data.H
zingale: fixed

PR #1665, opened 2021-05-06
convert ca_find_center to C++

zingale:
note: we don't really use this, but it runs w/o crashing when
I set castro.moving_center = 1

## PR checklist


Source/driver/Castro.cpp
```diff
@@ -4440,15 +4440,61 @@ Castro::define_new_center(MultiFab& S, Real time)
...
+            new_center[d] = geom.ProbLo(d) +  (static_cast<Real>(mi[d]) + 0.5_rt) * dx[d];
+        }
+
+        // Fit parabola y = a x^2  + b x + c through three points
+        // a = 1/2 ( y_1 + y_-1)
+        // b = 1/2 ( y_1 - y_-1)
+        // x_vertex = -b / 2a
+
+        // ... in x-direction
+        Real a = 0.5_rt * (data(mi[0]+1, mi[1], mi[2]) + data(mi[0]-1, mi[1], mi[2])) - cen;
```
maxpkatz, line 4477: I think the `- cen` here is not right because we've already offset the data by `cen` before this point, so in the Fortran code `data(0,0,0)` should have been exactly zero by this point? And while that's harmless, in this case we have already cached the value of `cen` to be the original density value before the offsetting.

zingale: fixed -- I removed the `- cen`

maxpkatz (approved):

PR #1669, opened 2021-05-07
we don't seem to use the outflow data in Fortran anymore

zingale:
## PR checklist


maxpkatz:
Do you remember what this data was for, because I don't.

zingale:
my guess was that it was for setting monopole BCs for the gravity solve?

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-06-003/index.html

maxpkatz (approved):

maxpkatz (approved):

maxpkatz:
For posterity, we think this is related to the old `spherical_star` functionality where you could represent a star which is bigger than the domain, and use the data from the domain corner to fill boundary conditions on the domain faces (which are geometrically closer to the center of the star).

PR #1673, opened 2021-05-07
Fix tiling bug in radiation estimate_gamrPr

maxpkatz:
## PR summary

The `gPr` fab that is filled in `estimate_gamrPr` is [sized to a tilebox] in the radiation dt estimator in Castro.cpp but the box we pass to the function to operate over is the validbox. So the loop is trying to fill in zones in `gPr` that don't exist, leading to illegal memory accesses.

## PR checklist


maxpkatz:
Closing. I introduced a bug in #1554 but this is the wrong fix to it, I will do the right one in a separate PR.

PR #1674, opened 2021-05-07
Fix loop bounds in estimate_gamrPr

maxpkatz:
## PR summary

PR #1554 introduced a bug for CPU builds (with tiling) by making the bounds of the loop based on the valid box, while the extent of the `gPr` array that is being filled is only the tile box.

## PR checklist

### DIFF
--- a/Source/radiation/MGFLD.cpp
+++ b/Source/radiation/MGFLD.cpp
@@ -1069,7 +1069,7 @@ void Radiation::estimate_gamrPr(const FArrayBox& state, const FArrayBox& Er,
         }
 #endif
 
-        amrex::ParallelFor(box,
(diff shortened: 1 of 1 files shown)


maxpkatz:
Note: tiling in this function was added in commit 06cd0da7be2de5ec13bf8c174980990898ef3d96. It looks like this should be revisited. In the current code we use the bounds of the tile box to determine whether we need to do a one-sided difference or a central difference. This means we're going to do more one-sided differences than necessary, since on the faces of tileboxes that are interior to the valid box, there is valid data in `Er` to difference with respect to. This was a non-issue prior to the introduction of tiling in this routine because the extent of `gPr` and `Er` would have been the same.

zingale (approved):

PR #1682, opened 2021-05-08
we no longer need to know about the pointmass in Fortran

zingale:
this removes the hooks and some unused headers

## PR checklist


Docs/source/gravity.rst
```diff
@@ -447,7 +447,7 @@ option works by adding the gravitational acceleration of the point
 mass onto the acceleration from whatever other gravity type is under
 in the simulation.
 
-.. note:: the point mass have a mass < 0
+.. note:: The point mass may have a mass < 0
```
maxpkatz, line 450: As far as I can tell this is actually true, but it worries me that it's true.

zingale: me too

PR #1695, opened 2021-05-09
remove ghost cells for React_Type

zingale:
these are never used

## PR checklist


zingale:
this has an issue with simplified-SDC

zingale:
curiously, the reacting tests have diffs:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-09-005/index.html

so those ghost cells are used for something?

zingale:
the issue is that we are actually making us of the reactions ghost cells at the moment.

We store the burn data in the Reactions_Type StateData and then after the burn, we apply it to the state.

In the case of the first half of Strang, we are burning on Sborder, which has 4 ghost cells, so we need the ghost cells in Reactions_Type so we can update everything in Sborder

Issue #1693, opened 2021-05-09
does Reactions_Type really need 4 ghost cells?


zingale:
It is not clear why Reactions_Type needs so many ghost cells.

maxpkatz:
This was done in #863. The logic at the time was that we were doing the load balancing based on the previous burn, and since we burn on the ghost zones in addition to the valid zones, we would want to have data on the weights for the load balancing in the ghost zones too. But we are no longer doing that, and in any case even if we did, in retrospect we could have done an estimation using the valid zones only. So I agree we could set it back to 0 now.

zingale:
There are a few issues here:

* We burn on Sborder in the first half of Strang.  We store the burn data in the Reactions_Type StateData and then after the burn, we apply it to the state.
* If we want to do a 4th order burn, we will need the ghost cells

zingale:
I wonder if we can get away with just a single time-level of Reactions_Type -- do we need both the old and new time?

maxpkatz:
We don't need Reactions_Type at all. As we discussed it's only relevant for plotfile data at this time.

zingale:
We also rely on this for `reactions_max_solve_level`

zingale:
this is done in #1770

PR #1705, opened 2021-05-12
change how Reactions_Type works

zingale:
this StateData is only for the plotfile in its current usage
we now only store the omegadot's in the plotfile if store_omegadot=1 is
explicitly set.  We also no longer need ghost cells here.  The reaction
update is now directly applied to the state in the Strang reactions
instead of doing it separately from adding Reactions_Type

## PR checklist


zingale:
I'm going to test this with `castro.store_omegadot=1` enabled for all problems

zingale:
this addresses #1693

zingale:
this probably breaks restart from old checkpoint files, but honestly, I never understood when Reactions_Type was stored in the checkpoint file anyway.

zingale:
there will be roundoff diffs

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-12-005/index.html

I may have messed something up with the omegadot's in the plotfiles?

Source/reactions/Castro_react.cpp
```diff
@@ -139,32 +139,56 @@ Castro::react_state(MultiFab& s, MultiFab& r, Real time, Real dt)
 
                 if (do_burn) {
 
-                    // Add burning rates to reactions MultiFab, but be
-                    // careful because the reactions and state MFs may
-                    // not have the same number of ghost cells.
+                    // update the state
 
-                    if (reactions.contains(i,j,k)) {
+                    if (U.contains(i,j,k)) {
```
maxpkatz, line 144: This contains check would no longer be necessary, since the box is defined with respect to the same array

Source/reactions/Castro_react.cpp
```diff
@@ -139,32 +139,56 @@ Castro::react_state(MultiFab& s, MultiFab& r, Real time, Real dt)
...
                         for (int n = 0; n < NumSpec; ++n) {
-                            reactions(i,j,k,n) = U(i,j,k,URHO) * (burn_state.xn[n] - U(i,j,k,UFS+n) * rhoInv) / dt;
+                            U(i,j,k,UFS+n) += U(i,j,k,URHO) * (burn_state.xn[n] - U(i,j,k,UFS+n) * rhoInv);
```
maxpkatz, line 146: We can simplify this to just be
```
U(i,j,k,UFS+n) = burn_state.xn[n] * U(i,j,k,URHO)
```

Source/reactions/Castro_react.cpp
```diff
@@ -139,32 +139,56 @@ Castro::react_state(MultiFab& s, MultiFab& r, Real time, Real dt)
...
+
+                    if (reactions.contains(i,j,k)) {
+
+                        reactions(i,j,k,0) = U(i,j,k,URHO) * burn_state.e / dt;
+                        reactions(i,j,k,1) = amrex::max(1.0_rt, static_cast<Real>(burn_state.n_rhs + 2 * burn_state.n_jac));
+
+                        if (store_omegadot == 1) {
+                            if (reactions.contains(i,j,k)) {
+                                for (int n = 0; n < NumSpec; ++n) {
+                                    reactions(i,j,k,2+n) = U(i,j,k,URHO) * (burn_state.xn[n] - U(i,j,k,UFS+n) * rhoInv) / dt;
```
maxpkatz, line 169: This needs to be done before the update to U, otherwise since U has already been updated to be the same as burn_state.xn this will just be zero

Source/reactions/Castro_react.cpp
zingale: fixed

Source/reactions/Castro_react.cpp
zingale: changed

Source/reactions/Castro_react.cpp
zingale: fixed

zingale:
latest tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-13-006/index.html

still not sure why the errors in omegadot (especially for the last one) are so large.

maxpkatz:
I recommend that you split off the diff that applies the reactions directly to the state in a separate PR so that we can isolate the numerical effect from that change.

zingale:
okay, splitting this PR up.

First PR is #1723

PR #1719, opened 2021-05-16
Convert filt_prim to C++

maxpkatz:
## PR checklist


maxpkatz:
Note: I deleted the mask because it wasn't used by the one setup actually using this functionality (RadBreakout), but it's easy to put back if we need it later.

Exec/radiation_tests/RadBreakout/filt_prim.H
```diff
@@ -0,0 +1,117 @@
...
+                Array4<Real> const& Stmp,
+                Array4<Real> const& Snew,
+                int filt_T, int S,
+                const GeometryData& geomdata,
+                Real time)
+{
+    using namespace filter;
+
+    bool filtered = false;
+
```
zingale, line 18: I think that we should have an assert here that this filtering is only implemented for 1-d

Exec/radiation_tests/RadBreakout/filt_prim.H
```diff
@@ -0,0 +1,117 @@
...
+        for (int n = 0; n < NumAux; ++n) {
+            eos_state.aux[n] = Stmp(i,j,k,UFX+n) * rhotmpInv;
+        }
+#endif
+
+        eos(eos_input_rt, eos_state);
+
+        Snew(i,j,k,UEINT) = Snew(i,j,k,URHO) * eos_state.e;
+        Snew(i,j,k,UEDEN) = Snew(i,j,k,UEINT) + 0.5_rt *
+                                                (Snew(i,j,k,UMX) * Snew(i,j,k,UMX)) /
```
zingale, line 100: our convention is to always include all 3 components of momentum in the kinetic energy

Exec/radiation_tests/RadBreakout/filt_prim.H
maxpkatz: I'll add an assert, but the change for dealing with multiple momentum components should come in some other PR since that's actually a change to the problem setup (and it wouldn't just be the kinetic energy, it would also be an actual filtering on the other momentum components too).

zingale (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-16-002/index.html

Issue #1645, opened 2021-04-23
rotating_torus has lots of Riemann errors


zingale:
In the first steps of the `rotating_torus` problem we get lots of warnings from the Riemann solver:

```
WARNING: (rho e)_r < 0 or pr < small_pres in Riemann: -1.590857916e-11 1.610360901e-15 1e-200
```

zingale:
This happens right at the first Riemann solve, so this is not due to the transverse corrections.  It seem like the data is so discontinuous in (rho e) that we are generating a negative interface value of (rho e), so maybe the warning is doing the right thing?

Still need to check if it is due to any source terms.

maxpkatz:
This is definitely due to the source terms. I have observed this now in multiple problems with self-gravity (but not any problems without gravity). For example you can see it in 1D DustCollapse. In particular you can see this even in the inner part of the domain where the density is initially large and uniform, with zero velocity. At t = 0, there is no velocity on the grid, and in the PPM the velocity jump is entirely dominated by the gravity source term. But the velocity jump has a contribution to the edge state for density and internal energy in the PPM scheme, and that is creating extreme values. In the 1D DustCollapse setup we have uniform rho = 10^9, but edge states that have densities of 1e12 and -1e12 (which then gets floored to 1e-6).

zingale:
this is likely related to #1599

maxpkatz:
Sketching out when we should expect this to be a problem: when there is uniform pressure and density, the alpha term goes as du * rho_ref / cs_ref, and when we are initially at rest (or at uniform velocity) du is dt * I_src / 2. So if we compare the ratio of this alpha term to rho_ref, we should expect it to cause a problem when it is of the same order of magnitude as rho_ref, which means dt * I_src / cs_ref ~ 1. This means we should expect it to be more likely to be a problem either when g is large or when dt is large. Since dt ~ dx / cs for hydro, this means we would also expect it to occur more preferentially at low resolution. This is consistent with my observation that I can make it go away with a *very* restrictive init_shrink -- for the giant impact setup, merely choosing 0.01 didn't do anything, but using 1e-6 did. This makes sense to me now, because the value I'd have to choose would need to be small enough to make the g term "small," which will be problem dependent and could be arbitrarily small for a given resolution.

zingale:
could we take the approach that if rho < rho_cutoff, then we don't care about the dynamics and not trace the source term in that cell?

maxpkatz:
[...]

PR #1725, opened 2021-05-19
Include the effect of the sources in the reference state

maxpkatz:
## PR summary

As discussed in #1645 it appears to be incorrect that we include the source term in the characteristic tracing but do not include the source term in the reference state. (This is done for PPM only.)

## PR checklist


Source/hydro/trace_ppm.cpp
```diff
@@ -468,11 +490,11 @@ Castro::trace_ppm(const Box& bx,
 
       // Set the reference state
       // This will be the fastest moving state to the right
-      Real rho_ref = Ip_rho[2];
-      Real un_ref = Ip_un_2;
+      Real rho_ref = Ip_rho[2] + hdt * Ip_src_un_2;
```
zingale, line 493: this should be Ip_src_rho_2

zingale:
I like this approach.  

I will want to test a longer run of `flame_wave` just to see if there are any issues

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-19-002/index.html

zingale:
I ran flame_wave for 47000 steps, comparing the current development to this PR

Here's development:
flame_wave_1000Hz_development_smallplt47484_slice

here's this PR:
flame_wave_1000Hz_source_reference_smallplt47140_slice

they look largely the same.  Maybe a little more stuff at the symmetry axis with this PR.  But the flame structure looks the same and they are taking pretty much the same timestep as before (and as we did in paper II).

zingale:
Note: this change eliminates the warnings like:
```
WARNING: (rho e)_r < 0 or pr < small_pres in Riemann: -2.580387688e+17 5.195193164e+10 1e-200
WARNING: (rho e)_r < 0 or pr < small_pres in Riemann: -2.581237387e+17 5.594825754e+10 1e-200
```
in the `DustCollapse` problem during the first timestep when using ``inputs_3d_poisson_regtest``

maxpkatz:
I have also observed this in some cases in the giant impact problem (#1650) and this change resolves the issue there too.

zingale:
can we do the ppm_integrate_profile instead of ppm_integrate_profile_single?  then I think it is ready to go.

maxpkatz:
I think I am doing more work than I need to here -- I added Im_src_rho_2 and Ip_src_rho_0, but these represent the waves moving in the other direction of the interface so they don't participate in the reference state. So maybe I just delete those extraneous ones and use ppm_int_profile_single since there's only 2 out of 3 needed?

zingale:
that's fine

zingale:
oh, but I think ppm_int_profile_single is per wave.  So you can't only get Ip for the 3 wave you also get Im for the 3 wave

maxpkatz:
Yes, that must be why I did that. OK, updated to use ppm_int_profile for rho and rhoe.

zingale (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-06-03-004/index.html

the rotating star w/ NSE gets an FPE now :/

PR #1747, opened 2021-05-24
Convert gravity tests to inputs-based tagging

maxpkatz:
## PR checklist

### DIFF
--- a/Exec/gravity_tests/uniform_cube_sphere/inputs
+++ b/Exec/gravity_tests/uniform_cube_sphere/inputs
@@ -15,6 +15,11 @@ amr.n_error_buf      = 0 0 0 0 0 0 0 0 0 0 0
 amr.blocking_factor  = 8
 amr.max_grid_size    = 8
 
+amr.refinement_indicators = denerr
+
+amr.refine.denerr.value_greater = 1.0e0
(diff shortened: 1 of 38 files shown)


zingale:
all tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-23-008/index.html

Exec/gravity_tests/uniform_cube_sphere/inputs
```diff
@@ -15,6 +15,11 @@ amr.n_error_buf      = 0 0 0 0 0 0 0 0 0 0 0
 amr.blocking_factor  = 8
 amr.max_grid_size    = 8
 
+amr.refinement_indicators = denerr
+
+amr.refine.denerr.value_greater = 1.0e0
+amr.refine.denerr.field_name = density
+
```
zingale, line 22: don't you need a max level here?

zingale (approved):

Exec/gravity_tests/uniform_cube_sphere/inputs
maxpkatz: No. By default the max level is 1000 so we only need to explicitly set the max level if we want the AMR not to be applied on every level.

PR #1774, opened 2021-05-27
Convert SCF initial model construction to C++

maxpkatz:
## PR summary

The restriction preventing its use with the GPU build is removed.

## PR checklist


maxpkatz:
The SCF iteration scheme is extremely numerically sensitive, and in my experimentation this changes the initial model slightly due to the small numerical difference in enthalpy according to the C++ versus Fortran EOS. However, the converged model is qualitatively nearly identical despite the fact that the fcompare diff is non-zero, so I am confident this change is correct.

zingale:
diffs are roundoff: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-30-005/scf-test.html

zingale (approved):

PR #1793, opened 2021-05-28
Set amr.regrid_on_restart = 1 by default

maxpkatz:
## PR summary

This helps support the SCF workflow we want of manually triggering regrids during the initialization phase, and there are probably not any workflows where regrid on restart is specifically undesired.

## PR checklist


zingale:
lots of failures: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-28-014/index.html

maxpkatz:
Closing, this parameter does something different than I thought it did so this is not desirable.

PR #1798, opened 2021-05-29
Convert wdmerger from probin to inputs

maxpkatz:
## PR checklist


Exec/science/wdmerger/tests/tde/inputs
```diff
@@ -37,3 +37,17 @@ amr.plot_per = 10.0
 amr.plot_int = -1
 amr.small_plot_per = 1.0
 amr.small_plot_int = -1
+
+problem.mass_P = 0.60
```
zingale, line 41: since we include the inputs_3d above, we now get everything that used to be in the probin there as problem.xxx parameters here, so if any are different than the defaults and not reset here, then we will be different.  Did you check that?

zingale: oh... nevermind, we never included a probin here.

zingale: oh wait.. I guess the way it works is that it still uses the probin in this directory previously, but now it is using what was in the ../../probin, so I think there is still an issue.

Exec/science/wdmerger/tests/tde/inputs.test
```diff
@@ -219,3 +213,21 @@ amr.derive_plot_vars = enuc rho_phiGrav angular_momentum_x angular_momentum_y an
...
+problem.problem = 2
+
+problem.tde_separation = 8.0e0
+problem.tde_beta = 6.0e0
+
+problem.nsub = 16
+
+problem.max_stellar_tagging_level = 20
+problem.max_temperature_tagging_level = 0
+problem.stellar_density_threshold = 1.0e0
```
zingale, line 233: you are missing the dxnuc_min tagging -- or did you not include it because the default max level is 0?

zingale:
tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-29-010/index.html

Exec/science/wdmerger/tests/tde/inputs
maxpkatz: I did not test this but I did think about it -- the end result should be that we get all of the problem parameters from the main inputs_3d file, followed by the overrides from this file. That results in something slightly different than what we had before only if the default probin parameters in the old probin in the top-level directory didn't match the _prob_parameters defaults, but I'm willing to deal with that at some later point if it ends up being an issue.

Exec/science/wdmerger/tests/tde/inputs.test
maxpkatz: Right, now that I realized the dxnuc_min tagging wasn't actually being applied in this and a couple other setups, I didn't bother bringing it over.

zingale (approved):

PR #1808, opened 2021-05-29
check the problem.* parameters in the inputs file

zingale:
this will catch misspellings, etc.

## PR checklist

### DIFF
--- a/Source/driver/Castro_setup.cpp
+++ b/Source/driver/Castro_setup.cpp
@@ -223,6 +223,19 @@ Castro::variableSetUp ()
 
   init_prob_parameters();
 
+  // check to make sure that we didn't set any parameters that don't
+  // exist in C++ (like because of misspelling).  All of the problem.*
+  // parameters should have been accessed via parmparse at this point.
+
+  if (ParmParse::hasUnusedInputs("problem")) {
+      amrex::Print() << "Warning: the following problem.* parameters are ignored\n";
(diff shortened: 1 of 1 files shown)


zingale:
I can do this for all of the Castro-controlled runtime paramerters

Also, we can make this an error in debug mode?

Source/driver/Castro_setup.cpp
```diff
@@ -223,6 +223,19 @@ Castro::variableSetUp ()
 
   init_prob_parameters();
 
+  // check to make sure that we didn't set any parameters that don't
+  // exist in C++ (like because of misspelling).  All of the problem.*
+  // parameters should have been accessed via parmparse at this point.
+
+  if (ParmParse::hasUnusedInputs("problem")) {
```
maxpkatz, line 230: I'm fine to do this; but the code already prints out the full list of unused ParmParse variables at the end of the run. Do you think there's value in adding this additional warning on top of that?

zingale: yes, because sometimes we never run things to the "end", aka, for big jobs that we just keep restarting.
I'd actually like to make it abort in debug mode -- then we have an easy test

maxpkatz:
> Also, we can make this an error in debug mode?

I am OK with making it an error in TEST and DEBUG mode.

maxpkatz (approved):

PR #1833, opened 2021-05-30
Remove GPU compatible problem define

maxpkatz:
## PR checklist

### DIFF
--- a/Exec/Make.Castro
+++ b/Exec/Make.Castro
@@ -111,16 +111,6 @@ ifeq ($(USE_GPU),TRUE)
   endif
 endif
 
-# Allow a problem to specify that we want to initialize
-# on the GPU if we're using CUDA. This is a subset of
-# dimension agnostic problems, and it requires a different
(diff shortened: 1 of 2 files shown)


zingale:
why does this get rid of the prefetches?

maxpkatz:
The prefetch to host is no longer meaningful because initialization will always happen on the GPU. The prefetch to the device -- I think that was never actually necessary. Prior to the point where we get to initdata, we will have initialized State_Type to zero (in the class constructor), so it should already be on the GPU. What historically happened is that [I originally added the prefetch to host](https://github.com/AMReX-Astro/Castro/commit/9055c31d60059431015862469a48e22a23c39973) when initdata was CPU-only, and then in #567 when GPU initdata was added, I probably looked at that code and thought, well now we're doing it on the GPU so we should prefetch to the GPU instead. Or maybe there was some other code at the time that caused the statedata to be on the host prior to initdata. In any case I think it's not true today.

zingale (approved):

PR #1835, opened 2021-05-30
Merge compute_avgpres into compute_radial_mass

maxpkatz:
## PR summary

This is a simple way to convert this functionality to C++. This also indirectly fixes errors in the subgrid calculation of the volume fraction per zone that were present in the Fortran code for various geometries and/or for drdxfac > 1.

## PR checklist


maxpkatz:
We still have no GR gravity test in the suite, however I tested this on StarGrav manually by adding USE_GR and the results look reasonable both before and after the change (difference in g is <1% in the interior of the WD).

maxpkatz:
Also worth noting that compute_radial_mass uses the following volume construction in 2D cylindrical coordinates:
```
vol_frac = 2.0_rt * M_PI * dx_frac * dy_frac * octant_factor * xx;
```
whereas the Fortran ca_compute_avgpres does:
```
rlo = abs(lo_i +  dble(ii  ) * dx_frac)
rhi = abs(lo_i +  dble(ii+1) * dx_frac)

vol = (rhi**2 - rlo**2) * dy_frac
```

zingale (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-30/index.html
failures are because I didn't update the probin in the test script

zingale:
those axisymmetric volume factors are algebraically the same, aside from the pi (which I assume cancels) and the octant_factor

PR #1841, opened 2021-05-30
Do not apply CFL constraint in sponge region

maxpkatz:
## PR summary

The meaning of applying a sponge is that we consider the material to be dynamically irrelevant, and want to ensure that it doesn't grow out of control and interfere with the material of interest. So we should be consistent with that and ignore this material in setting the timestep constraint. If the sponge is doing its job, there should not be any problems caused by the ambient material.

## PR checklist

### DIFF
--- a/Source/driver/timestep.cpp
+++ b/Source/driver/timestep.cpp
@@ -13,6 +13,10 @@
 #include <Rotation.H>
 #endif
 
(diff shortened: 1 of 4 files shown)


maxpkatz:
Closing for now, the problem seems to have been solved using a different approach.

PR #1842, opened 2021-05-30
Use S_new instead of S_old in CFL violation check

maxpkatz:
## PR summary

The CFL violation check just prior to the hydro is intended to guess whether the hydro update will violate the CFL stability criterion. For that we can either use the old state or the state updated with the old-time sources. This change switches from the former to the latter. The motivation is the observation that when we fill new fine grids with data from coarse grids, the momentum may be inaccurate at a sharp density gradient like at the edge of a star. This change allows us to apply the sponge to the state (if a sponge is used) before checking whether it violates the CFL criterion, which will avoid spurious CFL violation warnings.

## PR checklist

### DIFF
--- a/Source/driver/Castro_advance_ctu.cpp
+++ b/Source/driver/Castro_advance_ctu.cpp
@@ -153,8 +153,11 @@ Castro::do_advance_ctu(Real time,
     if (do_hydro)
     {
 #ifndef MHD
-      // Check for CFL violations.
-      check_for_cfl_violation(S_old, dt);
+      // Check for CFL violations in S_new. The new-time state has seen
+      // the effect of the old-time source terms, so it is a first-order
+      // accurate estimate of whether the advance will violate the CFL
+      // criterion.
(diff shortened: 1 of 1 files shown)


maxpkatz:
Note that another way to do this which would be even more accurate would be to use S_old + dt/2 * sources_for_hydro. However, this CFL check is just a heuristic which determines whether we need to do a retry before doing the hydro update, it has no dynamical effect other than that. A state which strongly violates the CFL criterion will probably still do so regardless of whether we do a time-centered estimate or not, so it will be caught by the CFL retry check at the time of the timestep. Thus we don't need to be too accurate in this check, the main point is to avoid the kind of updates which are so messed up that they would cause NaNs.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-30-008/index.html
things look good

zingale (approved):

PR #1848, opened 2021-05-31
Only warn on density reset if below retry_small_density_cutoff

maxpkatz:
## PR summary

If the user sets `castro.retry_small_density_cutoff` > 0, they are telling the code not to care about density resets below this threshold. So the user probably does not want to be warned about these resets either; this change turns the warning off for that case. It can still be opted into with `castro.v = 2`.

## PR checklist


zingale:
all tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-31-006/index.html

zingale (approved):

PR #1849, opened 2021-05-31
Clean Sborder after FillPatch

maxpkatz:
## PR summary

The FillPatch that creates Sborder from S_old can cause thermodynamically inconsistent states and/or small densities in the ghost zones, so we do need to clean Sborder (although as a later enhancement, we could change this clean to only operate on the ghost zones).

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-31-003/index.html

maxpkatz:
A problem here is that clean_state requires the magnetic fields for MHD, but Sborder operates on NUM_GROW ghost zones which is than the magnetic fields have. So the call to reset_internal_energy won't make sense, because B doesn't have enough ghost zones.

maxpkatz:
I don't see how to fix the MHD problem so for now I'm just ifdef'ing the call out when MHD is enabled. Please re-test.

zingale:
can we use Bx_old, By_old, and Bz_old?

maxpkatz:
No, they don't have enough ghost zones to do a clean_state over NUM_GROW ghost zones.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-06-02-001/index.html

zingale:
`Detonation-collision-retry-SDC` failed with:


```amrex::Error::3::Invalid mass fraction in Castro::normalize_species() !!!```

zingale:
I think that failure is just because the detonation is tricky and that check is actually meaningless at that point in the SDC algorithm.  I've seen this before with other SDC problems but haven't yet had time to address it.

maxpkatz:
OK, I agree that this is not directly related to this change (it's crashing in a different call to clean_state than the one I added) so we can address it separately.

zingale (approved):

zingale (approved):

PR #1850, opened 2021-05-31
Protect against negative (rho e) in trans

maxpkatz:
## PR summary

When using `transverse_reset_rhoe`, the code detects if the (rho E)-based transverse update makes (rho e) negative, and if so resets to an update that is based on the equation for (rho e). But it's possible for the (rho e) update to also result in a negative (rho e), particularly for ambient material. So we check against this case now and turn off the transverse correction if it would result in a negative (rho e). Also, if there was a density reset then we make (rho e) the unmodified edge state, similar to what we do with pressure.

## PR checklist

### DIFF
--- a/Source/hydro/trans.cpp
+++ b/Source/hydro/trans.cpp
@@ -390,6 +390,14 @@ Castro::actual_trans_single(const Box& bx,
 #endif
             }
 
+            // If (rho e) is negative by this point,
+            // set it back to the original interface state,
+            // which turns off the transverse correction.
+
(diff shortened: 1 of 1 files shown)


zingale:
this looks reasonable.  I'll test shortly

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-05-31-005/index.html

zingale:
this seems to work as intended, based on the tests

zingale (approved):

PR #1851, opened 2021-05-31
Determine time weighting in sponge based on relative velocities

maxpkatz:
## PR summary

Rather than evenly weighting the old-time and new-time sponge sources in the predictor-corrector approach, we determine the weighting based on the relative velocities at the old and new times. The time level which has a greater velocity is sponged to a greater extent. In the asymptotic limit of the old time velocity being zero and the new time velocity being large, then this looks like a time-advanced source; in the limit of the old time velocity being large and the new time velocity being zero, then this looks like a time-lagged source; and if they are equal, then we recover the old time-centered source. This solves problems caused by the velocity changing by multiple orders of magnitude during a timestep, as can happen after a density reset and/or in ambient material near the edge of a star.

## PR checklist


maxpkatz:
Tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-06-04-002/index.html

maxpkatz:
Closing, we are going with #1876 instead.

PR #1852, opened 2021-05-31
Don't limit future timesteps based on previous retries

maxpkatz:
## PR summary

#344, which added dynamic retries, also made a decision to limit the next timestep based on retries in the previous timestep. However with retrospect this appears to have been a bad choice. Because we now always cut by a factor of 2 when we take a retry, this is effectively halving the next timestep. But if the condition that caused the retry (for example, a temporarily large velocity) has gone away by the time we get to the next timestep, there is no need to take a shorter timestep. If the condition is still there, it will appear in the CFL criterion and we will take a shorter timestep anyway. So in the optimistic case, removing this will allow us to take asymptotically larger timesteps.

## PR checklist


zingale:
what about the case where we limited because the burning failed?

maxpkatz:
Today that too would cause us to take a shorter dt in the next coarse timestep. This change would remove that behavior. I am OK with this because burn failures are extremely sensitive, there's no reason to believe that a burn failure in any given step is correlated to a burn failure in a later step.

maxpkatz:
I can imagine a scenario involving generally difficult burns where there are just a bunch of successive burn failures because the hydro timestep was too big relative to the burning timescale, but the user shouldn't be relying on retries to make that case work, they should use something like dtnuc_e.

zingale:
okay, I believe that argument.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-06-01-011/index.html

maxpkatz:
Some of those failures are due to this branch being out of sync with respect to development, which I have now fixed, so we should retest.

zingale:
one problem crashes: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-06-01-013/index.html

zingale (approved):

PR #1873, opened 2021-06-05
Remove CFL violation check for CTU advance

maxpkatz:
## PR summary

The code in question checks the old state (plus, as of #1842, the old-time sources) to see whether it violates the CFL criterion. But this has been superfluous ever since we added the logic at the end of the timestep to see if the timestep constraint is violated at that point. If there is a violation at that point, we will retry the timestep. So this prior additional check is now removed.

## PR checklist

### DIFF
--- a/Source/driver/Castro_advance_ctu.cpp
+++ b/Source/driver/Castro_advance_ctu.cpp
@@ -153,19 +153,6 @@ Castro::do_advance_ctu(Real time,
     if (do_hydro)
     {
 #ifndef MHD
-      // Check for CFL violations in S_new. The new-time state has seen
-      // the effect of the old-time source terms, so it is a first-order
-      // accurate estimate of whether the advance will violate the CFL
-      // criterion.
-      check_for_cfl_violation(S_new, dt);
-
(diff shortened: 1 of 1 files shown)


maxpkatz:
Note: for now, I'm leaving the CFL violation infrastructure in since true SDC doesn't have retry capability yet, so the analogous check in Castro_advance_sdc.cpp is the only way for the user to know if the CFL constraint was violated in true SDC.

maxpkatz:
Tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-06-04-003/index.html

The two failures are wdmerger_retry and wdmerger_retry-SDC. wdmerger_retry fails because it was intentionally relying on this CFL violation check to trigger a retry. This is not a serious problem, by this point we have other tests that invoke a retry unintentionally, but we could change the test again to try and trigger a retry a diffferent way. wdmerger_retry-SDC fails because of an unrelated bug which is addressed by #1874 (I don't know why we only just exposed this but it's definitely a thing that should be fixed).

zingale (approved):

PR #1878, opened 2021-06-05
some fixes for recent versions of yt

zingale:
we need to apply units and work around a bug in the BoxLib frontend detection

## PR checklist


maxpkatz:
Can you elaborate on what the issue is with the yt frontend detection -- is it going to break the ability for all Castro plotfiles to do `yt.load()` or is it isolated? And what version does it start with?

zingale:
it happens with the development version of yt 4.0.  I am hopeful it will be fixed before 4.0 is actually released, but it is affecting usage now, so we might need to start hacking around the yt bug in all of our scripts.

There is an issue here:

https://github.com/yt-project/yt/issues/3005

maxpkatz (approved):

Issue #341, opened 2018-06-03
Apply density flux limiter on a reflux


maxpkatz:
It is possible for a reflux to generate a negative density on the coarse grid due to the fine grid flux correction. We should limit the flux in this case (assuming the density flux limiter is enabled), rather than allowing a reset to occur.

One issue is that it will not be conservative if we do not apply the same correction to the fine grid flux. We could either figure out how to do that in the reflux, or perhaps we could investigate filling in the coarse flux in the ghost zones of the flux array before the hydro advance, so the data is already there to apply to the flux limiter on the fine level.

maxpkatz:
This (well, a closely related variant of this) is a blocking issue for the TDE setup. Refluxes are generating very large (superluminal) velocities in ambient material as the WD moves across the domain. These result either in very small timesteps or in a large number of retries being necessary. So we need to be able to apply the velocity-based flux limiter on the fluxes at the coarse-fine interface.

maxpkatz:
A challenge with this is that in order to avoid the coarse/fine difference causing a problem, we need to add the fine zones together and compare to the coarse flux, but if there is a problem, which fine zones do we modify?

maxpkatz:
Or perhaps we simply limit the flux coming into the coarse zone and let it be non-conservative?

maxpkatz:
#1879 proposes the non-conservative approach since it is straightforward to implement and can be done entirely within the reflux step. The conservative approach is safer but also harder to implement since it involves referring to coarse data during the hydro update. So I think we should go with the simpler approach and assume that the user doesn't have significant amounts of mass involved in the locations where a negative density update would occur. If that later turns out to be untrue for a science problem we care about, we can reinvestigate this.

PR #1879, opened 2021-06-06
Zero out flux corrections that would cause small density

maxpkatz:
## PR summary

As observed in #341, it's possible for the reflux operation to result in a small or negative density. Currently this is dealt with via a density reset. Density resets are fairly ugly and should be avoided since they can cause a large change in density (down to small_dens) that has downstream consequences. So instead what we can do is just zero out the flux corrections if they would cause a small or negative density. This PR applies that change. Another option would have been to use the existing limit_hydro_fluxes_on_small_dens() function, however this function's interface currently relies on having the primitive variables, which are not already available in the reflux operation. A future change could do this if desired.

This PR also modifies the update on the fluxes so that they are done on a direction-by-direction basis. While this is not strictly necessary to support this change, it should save on memory since we don't actually need to have all three directional fluxes constructed simultaneously.

## PR checklist


maxpkatz:
Note for the reviewer: the diff makes this hard to see, but the only new content here is lines 2625 to 2680 (using the line count in the new code). The remainder is existing code that has been rearranged (so that several idir loops are merged into one).

maxpkatz:
Example of a #1650 setup with one level of refinement before and after this change. Some ambient zones (which start with 1e-7 g/cc) at the coarse-fine boundary are going small or negative and being reset to small_dens (1e-8 g/cc). This change makes the boundary much smoother.

Before:

plt00068_density

After:

plt00068_density

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-06-06-002/index.html

zingale (approved):

Issue #680, opened 2019-10-08
setVal on device needs to be launched in a kernel


maxpkatz:
The Fab setVal for the shock variable in Castro_ctu_hydro.cpp needs to be launched in a kernel region, otherwise it will happen on the host and cause unintended race conditions. We should fix this.

zingale:
do we also need to do `hydro_source` in the same file?

maxpkatz:
No, MultiFab setVal is fine.

zingale:
this has been fixed

Issue #1883, opened 2021-06-06
allow problem.center to be set via inputs


zingale:
Some problems already (try to) do this (but use `castro.center`).  But at the moment, our runtime parameter scripts don't generate the parmparse for an array.  So we should extend the scripts to allow this and allow for `problem.center` to be set.

zingale:
this has been done.

PR #1888, opened 2021-06-07
allow for problem.center to be read via inputs

zingale:
## PR summary

This requires https://github.com/starkiller-astro/Microphysics/pull/680

Note: we currently define center as having a size "3" regardless of dimensionality,
so we will always need to supply 3 values if we try to set it via inputs.

## PR checklist


zingale:
addresses #1883

maxpkatz:
What do you think about making a more conscious effort at updating the submodules for mid-release updates like this?

zingale:
Sure, I can do that tomorrow.

I usually imagine that anyone using development is not using submodules, but you're right.

maxpkatz (approved):

Issue #1000, opened 2020-05-26
consider setting small_pres automatically


zingale:
Currently, if small_pres is not set, we default to 1.e-200, which is quite small.  We should try to find a robust way to set it via the EOS.

maxpkatz:
The main challenge here I have faced when I thought about doing this is that small_pres = small_pres(small_dens, small_temp, X); what should we choose for X? Do we need to grid search over all possible combinations of X to find small_pres?

PR #1895, opened 2021-06-09
Make small_pres and small_ener consistent with small_dens and small_temp

maxpkatz:
## PR checklist


zingale:
this addresses #1000

maxpkatz:
Indeed. I decided to punt on the issue of "how to choose X" in this evaluation -- small_pres doesn't need to be a rigorously defined quantity, it just needs to be the right order of magnitude.

Source/driver/Castro_setup.cpp
```diff
@@ -259,6 +259,33 @@ Castro::variableSetUp ()
...
+  eos_state.T = castro::small_temp;
+  for (int n = 0; n < NumSpec; ++n) {
+      eos_state.xn[n] = 1.0_rt / NumSpec;
+  }
+#if NAUX_NET > 0
+  for (int n = 0; n < NumAux; ++n) {
+      // This value is arbitrary, we may want to fix this
+      // later if it causes a problem.
+      eos_state.aux[n] = 1.0_rt;
+  }
```
zingale, line 281: I think if we are doing NSE_THERMO, we should properly call `set_nse_aux_from_X(state_t& state)`

maxpkatz: OK. I did this and removed what was there since it wasn't great, we can find a better solution later if we have non-NSE problems using aux.

zingale:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-06-09-003/index.html

most diffs are small, but there are 2 FPEs right out of the gate.

maxpkatz:
This is happening in gamma law, we are calculating e.g. p = K * rho * T and rho and T are both 1e-200 so this expression is underflowing to zero, resulting in a later divide-by-zero.

maxpkatz:
I've pushed a change that increases the default value of small_temp and small_dens a hundred orders of magnitude (to 1e-100) to workaround this. This may alter more tests though.

zingale:
updated tests look good: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-06-09-005/index.html

zingale (approved):

PR #1901, opened 2021-06-10
Set up reacting_convergence with silicon burning

harpolea:
Sets up initial conditions for the `reacting_convergence` problem to mimic those in the silicon burning layer in the massive star problem.

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-07-06-001/index.html

these are slight bigger than roundoff?

zingale (approved):

zingale:
actually, I think the diffs are just from the renormalization with the new species initialization.

PR #1903, opened 2021-06-10
don't create the Fortran state_indices_nd.F90 unless we really need it

zingale:
that is only in radiation now

## PR checklist


zingale:
quick tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-06-10-005/index.html

maxpkatz (approved):

PR #1915, opened 2021-06-20
Convert ca_integrate_grav to C++

maxpkatz:
## PR checklist


Source/gravity/Gravity.cpp
```diff
@@ -3199,12 +3198,80 @@ Gravity::make_radial_gravity(int level, Real time, RealVector& radial_grav)
...
     // Integrate radially outward to define the gravity
-    ca_integrate_grav(radial_mass_summed.dataPtr(), radial_den_summed.dataPtr(),
-                      radial_grav.dataPtr(),
+
+    Real halfdr = 0.5_rt * dr;
+
+    Real mass_encl = 0.0_rt;
+    Real vol_total_i, vol_outer_shell, vol_upper_shell;
+
+    Real* const den = radial_den_summed.dataPtr();
```
zingale, line 3207: why do we need to use a dataPtr here?

maxpkatz: We don't need to, just keeping the code consistent with what we were doing in Fortran. Plus this keeps the lines shorter :-)

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-06-20-004/index.html

only one failure, which is roundoff

zingale (approved):

PR #1921, opened 2021-06-27
switch the computation of dtnuc_e to use the integral over the timestep

zingale:
for burns trying to reach equilibrium, this is better behaved

## PR checklist


zingale:
2 things:

1. do we want to make this an option?  So we can choose the integral or instantaneous?
2. I think I need to put Reactions_Type back into the checkpoint file for this

zingale:
this seems to help a lot with `subchandra`

zingale:
I no longer think that this is the way to go.

zingale:
this is a plot of |de/dt| over 4 timesteps for the subchandra behind the inward propagating shock.  Given the large swings in the value here, this suggests that using the integral might be the better solution.

dedt

maxpkatz:
If we're going to use information from during the previous burn then we more options than just the average. Particularly, we could record during the burn the maximum instantaneous energy generation rate after each burn step, and then take the max of that after the burn, and use that as the enuc value used in the timestep limiter.

PR #1924, opened 2021-07-01
Initialize riemann_2d

aisclark91:
I have implemented the Liska & Wedroff (2003) Riemman 2d problem. I am satisfied with the result, but I would like to improve the resolution of each case described here.

## PR checklist


Exec/hydro_tests/riemann_2d/problem_initialize_state_data.H
```diff
@@ -0,0 +1,95 @@
...
+	{
+		//State in domain 4
+		state(i, j, k, URHO)  = problem::rho_4;
+
+		state(i, j, k, UMX)   = state(i, j, k, URHO) * problem::ux_4;
+		state(i, j, k, UMY)   = state(i, j, k, URHO) * problem::uy_4;
+		state(i, j, k, UEINT) = problem::p_4 / (eos_gamma - 1.0_rt);
+
+		state(i, j, k, UEDEN) = problem::p_4 / (eos_gamma - 1.0_rt)
+			                     + 0.5 * problem::ux_4 * problem::ux_4;
```
zingale, line 63: this is missing a problem::rho factor

Exec/hydro_tests/riemann_2d/problem_initialize_state_data.H
```diff
@@ -0,0 +1,95 @@
...
+
+
+#if AMREX_SPACEDIM == 2
+	Real xx = problo[0] + dx[0] * (static_cast<Real>(i) + 0.5_rt);
+	Real yy = problo[1] + dx[1] * (static_cast<Real>(j) + 0.5_rt);
+#endif
+
+	//Now we want to fracture the domain in 4 regions as described in
+	//Liska & Wendroff (2003).
+	if (xx < problem::center[0] && yy < problem::center[1])
```
zingale, line 39: can you add a comment just to help make it clear which quadrant is which? like just doing
```
// 2|1
// -+-
// 3|4
```

zingale:
can you do a run with the latest changes and post an image in this discussion so we can see the output?

aisclark91:
> can you do a run with the latest changes and post an image in this discussion so we can see the output?

I build the code with the inputs_c3 case and got the following picture:
riemman_2d_case3

Exec/hydro_tests/riemann_2d/inputs_c12
```diff
@@ -0,0 +1,106 @@
...
+amr.ref_ratio       = 2 2 2 2       # refinement ratio
+amr.regrid_int      = 2             # how often to regrid
+amr.blocking_factor = 8             # block factor in grid generation
+amr.max_grid_size   = 256           # maximum default value
+
+amr.refinement_indicators = presserr pressgrad
+
+amr.refine.presserr.value_greater = 2
+amr.refine.presserr.field_name = pressure
+amr.refine.presserr.max_level = 3 #Since blocking_factor = 2**3, 3 is the max level chosen for refinement.
```
zingale, line 49: blocking_factor means that the grid sizes will be multiples of blocking_factor -- it doesn't affect the maximum level of refinement.  That is controlled by max_level

aisclark91: I see, so the blocking factor essentially constraint the size of the cells that are chosen for the refinement.

zingale (approved):

PR #1931, opened 2021-07-08
enforce X interface states are in [0, 1] with SDC reaction source term

zingale:
## PR checklist


zingale:
This change seems to work well.  In an ideal world, the sources from the SDC reaction integration would be well behaved when we apply them to the interface, but sometimes with the lagging or the tracing of species to the interface, we go outside of [0, 1], so this just clips the states.  This only applies for simplified-SDC.

harpolea (approved):

zingale:
also the massive star diffs: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-07-13-003/massive_star-sdc.html

PR #1949, opened 2021-07-22
make burn_weights storage optional (off by default)

zingale:
## PR summary

this also expands the amount of information: we now separately store
the first and second halfs for Strang, and all iterations separately
for SDC (with 1 extra interation in case of a retry bump).  Also,
the burn_weight info is now at the end of the Reactions_Type array,
since it can be hard to infer how many sdc iterations there are
in the middle of the update.

## PR checklist


maxpkatz:
Note for posterity: this change does not require a checkpoint version number change because Reactions_Type is currently not stored in the checkpoint.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-07-22-001/index.html

yes, they all fail, but that's because the burn_weights field is now renamed

maxpkatz:
The result of this change for Strang will be that the old-time Reactions_Type has the old-time burn weights, and the new-time Reactions_Type will have the new-time burn weights, but both time levels will have stored data for both, with an array of zeros in each case that carries no information. While this isn't actively harmful, it does kind of defeat the purpose of having old-time and new-time data.

zingale:
oh, that's a good point.  I think what I need to do is copy the first half data into R_new at the end

zingale:
it might be simpler to make this not state data.  Then we can just store it in a MF and copy it to the plotfile at output.  The benefit is that we only need a single time-level for this.

zingale:
Max points to this example:

https://github.com/AMReX-Astro/Castro/blob/main/Source/driver/Castro_io.cpp#L1110-L1115

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-07-26-004/index.html

zingale:
note: turn off whitespace when reviewing (the editor got a little indent happy)

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-08-18-002/index.html

zingale:
tests:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-09-07-001/index.html

zingale:
this is ready for review

CHANGES.md
```diff
@@ -1,3 +1,11 @@
+# 21.10
+
+   * The burn_weights are no longer stored by default in the plotfile.
+     Instead, they are not enabled by setting
```
maxpkatz, line 4: typo: not -> now

Source/driver/Castro_io.cpp
```diff
@@ -1114,6 +1132,14 @@ Castro::plotFileOutput(const std::string& dir,
     }
 #endif
 
+#ifdef REACTIONS
+#ifndef TRUE_SDC
+    if (store_burn_weights) {
+        MultiFab::Copy(plotMF, getLevel(level).burn_weights, 0, cnt, Castro::burn_weight_names.size(),0);
```
maxpkatz, line 1138: Can you increment `cnt` here appropriately so that we don't make a mistake if we add something later?

Source/driver/Castro_setup.cpp
```diff
@@ -644,26 +642,40 @@ Castro::variableSetUp ()
...
+  
+  if (store_burn_weights) {
+
+#ifdef STRANG
+      burn_weight_names.push_back("burn_weights_firsthalf");
+      burn_weight_names.push_back("burn_weights_secondhalf");
+#endif
+#ifdef SIMPLIFIED_SDC
+      for (int n = 0; n < sdc_iters+1; n++) {
+          burn_weight_names.push_back("burn_weights_iter_" + std::to_string(n+1));
```
maxpkatz, line 678: Don't tell Alice these names are 1-indexed.

maxpkatz (approved):

CHANGES.md
zingale: fixed

Source/driver/Castro_io.cpp
zingale: done

PR #1953, opened 2021-08-02
for multilevel simplified-SDC, explicitly fill the react source ghosts

zingale:
we do this before hydro because any average down may have made the ghost
cells inconsistent which can be especially problematic at symmetry boundaries

## PR checklist

### DIFF
--- a/Source/driver/Castro_advance_ctu.cpp
+++ b/Source/driver/Castro_advance_ctu.cpp
@@ -148,11 +148,13 @@ Castro::do_advance_ctu(Real time,
 
 
 #ifdef SIMPLIFIED_SDC
+#ifdef REACTIONS
     // the SDC reactive source ghost cells on coarse levels might not
     // be in sync due to any average down done, so fill them here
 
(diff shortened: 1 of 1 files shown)


Source/driver/Castro_advance_ctu.cpp
```diff
@@ -147,6 +147,14 @@ Castro::do_advance_ctu(Real time,
     }
 
 
+#ifdef SIMPLIFIED_SDC
+    // the SDC reactive source ghost cells on coarse levels might not
```
maxpkatz, line 151: Need to guard this also with #ifdef REACTIONS, I believe

zingale: done

zingale:
quck tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-08-02-001/index.html

zingale:
SDC tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-08-02-002/index.html

maxpkatz (approved):

PR #1646, opened 2021-04-25
change how we handle species sources in simplified-SDC

zingale:
instead of adding the Iq source to the normal interface state before the
transverse corrections, we now wait until the end.  This is still 2nd order
accurate, but because we don't trace the state or put it through the
different transverse operations, there is no danger of it violating
sum{Iq} = 0

## PR checklist

- [ ] test suite needs to be run on this PR
- [ ] this PR will change answers in the test suite to more than roundoff level
- [ ] all newly-added functions have docstrings as per the coding conventions
- [ ] the `CHANGES` file has been updated, if appropriate
- [ ] if appropriate, this change is described in the docs

Issue #1964, opened 2021-08-23
remove all SDC sources from characteristic tracing


zingale:
as described in #1962 and #1963, bad things can happen if the SDC source terms to p and (rho e) make those quantities negative.  We should remove those sources from the characteristic tracing and instead apply them only at the very end, after the transverse corrections.  This will allow us to ensure that the final interface states are thermodynamically consistent and will still be overall second order accurate.

PR #1965, opened 2021-08-23
don't trace any SDC sources

zingale:
previously, the pressure and energy source terms were included in the PPM
tracing.  Now we no longer do this and instead explicitly add them at the
end of all of the transverse updates.  This has the advantage of letting
us check whether the interface states are negative and also removes
their influence from the reference states and eigensystem

Note: this also restores the pure hydro (non-SDC) species sources to the tracing
routines (they were taken out in #1646)

## PR checklist


zingale:
this addresses #1964

zingale:
note: for completeness, I should add back in the pure hydro (non-SDC) species sources to trace_ppm (if they exist)

Source/driver/Castro_setup.cpp
```diff
@@ -472,12 +472,13 @@ Castro::variableSetUp ()
...
 #ifdef REACTIONS
   // For simplified SDC, we want to store the reactions source.
+  // these are not traced, so we only need a single ghost cell
 
   if (time_integration_method == SimplifiedSpectralDeferredCorrections) {
 
       store_in_checkpoint = true;
       desc_lst.addDescriptor(Simplified_SDC_React_Type, IndexType::TheCellType(),
-                             StateDescriptor::Point, NUM_GROW_SRC, NQSRC,
+                             StateDescriptor::Point, 1, NQSRC,
```
zingale, line 481: using `1` here causes a Backtrace

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-09-14-001/index.html

these look good

zingale:
I would still like this reviewed and merged.

PR #1966, opened 2021-08-25
add a reacting_bubble inputs that works for simplified-SDC

zingale:
note: I needed to set react_rho_min for this to work

## PR checklist


zingale:
"works" might be a strong word here... it actually crashes.  But we should use this for debugging SDC with gravity.

maxpkatz (approved):

PR #1974, opened 2021-09-09
eliminate PhiRot_Type from StateData

zingale:
## PR checklist

### DIFF
--- a/Source/scf/scf_relax.cpp
+++ b/Source/scf/scf_relax.cpp
@@ -2,6 +2,7 @@
 #include <Castro_F.H>
 #include <fundamental_constants.H>
 #include <Gravity.H>
+#include <Rotation.H>
 
 using namespace amrex;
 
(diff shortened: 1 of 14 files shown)


zingale:
still need to do wdmerger

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-09-09-002/index.html

Source/scf/scf_relax.cpp
```diff
@@ -620,7 +610,8 @@ Castro::do_hscf_solve()
 
         for (int lev = 0; lev <= finest_level; ++lev) {
 
-            const Real* dx = parent->Geom(lev).CellSize();
+            auto geomdata = parent->Geom(lev).data();
+            const Real* dx = geomdata.CellSize();
```
maxpkatz, line 614: This should probably be CellSizeArray for safety.

maxpkatz:
It would be consistent with past behavior if this increased the checkpoint version number, although I am not sure if it's necessary.

zingale:
I'll increment the checkpoint number just to be consistent

Source/scf/scf_relax.cpp
zingale: done

zingale:
updated with checkpoint version increment

maxpkatz (approved):

PR #1977, opened 2021-09-09
Paralell Plane 1/r^2 Gravity Added

aisclark91:
## PR summary

I added the plane parallel gravity 1/r^2 by modifying the Gravity.cpp and _cpp_parameters file.

## PR checklist


zingale:
I don't see where this new function is called.

I think that the original idea was to just change the existing pointmass function, not add a new function.

The pointmass stuff adds to the gravitational acceleration after all of the other gravity options are considered.  So you could just run with ConstantGrav, but set the gravitational acceleration to 0 and then use the pointmass to create the actual gravity you want.

aisclark91:
ohh, I though each option added separately. Ok I will modify it. Maybe I will create a new pull request, without many commits on it.

maxpkatz:
The places where the existing pointmass gravity are added are here:
https://github.com/AMReX-Astro/Castro/blob/21.09/Source/gravity/Gravity.cpp#L902-L906
https://github.com/AMReX-Astro/Castro/blob/21.09/Source/gravity/Gravity.cpp#L975-L980

The call to the new function could be added there. Though I do think it's a little cleaner to just merge this into the existing point mass function.

PR #1981, opened 2021-09-10
Fix _prob_params parsing

### LINKED ISSUE #1980: _prob_params is parsed incorrectly if it doesn't have an ending newline
While investigating #1979, I found a bug in write_probdata.py: if a _prob_params file doesn't end with a newline, then the last character on the last line will be skipped. This can be seen in Exec/reacting_tests/reacting_convergence, where the namelist flag for spec6_frac is missed, and it can't be set by the inputs file:

```
$ ./Castro3d.gnu.ex massive_star_inputs/inputs.64
...
Successfully read inputs file ... 
Warning: the following problem.* parameters are ignored
problem.spec6_frac = 1.376063e-03
```

This has an easy fix, and I'll submit a PR shortly.

yut23:
If a line doesn't have a comment, then `line.find("#")` returns -1.
Normally, `line[:-1]` will just remove a trailing newline, but if the
file doesn't have a newline at the end, this will instead remove the
last character on the line. If this is a "y" in the 4th column, then the
last parameter cannot be set at runtime.
Closes #1980.

## PR checklist

### DIFF
--- a/Util/scripts/write_probdata.py
+++ b/Util/scripts/write_probdata.py
@@ -56,6 +56,8 @@ def get_next_line(fin):
         line = fin.readline()
         pos = line.find("#")
 
+    if pos == -1:
+        return line.strip()
     return line[:pos]
 
 


zingale (approved):

zingale:
nice catch

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-09-10-002/index.html

Issue #1952, opened 2021-07-31
Negative species should trigger a retry


zingale:
Right now we abort if any of the species are really negative in clean_state.  We should instead do a retry

zingale:
this really seems to be an issue with SDC for S_new after we apply the sources

zingale:
I think that the easiest way to do this is to have `clean_state` return a status, which indicates if that species check failed, and if it does, we modify the `advance_status status` variable accordingly in `Castro_advance_ctu.cpp`.

Thoughts?

maxpkatz:
I prefer not to handle it exactly this way -- I would rather have it be that any update which updates the species checks at the time of the update whether it is creating an invalid state, and then reject the step at that point. Everything in clean_state() should be a backstop which resets the fluid state to something reasonable if all else fails, but we should be more rigorous about validating the output of the various updates.

Also, seems like an issue that we are only aborting in the CPU build for this error.

zingale:
yes, we should abort on the GPU as well.  Does abort work there?

maxpkatz:
It's not ideal to handle it that way -- the way to do it would be to reduce over the minimum and maximum X found, and apply the check after the parallel for.

maxpkatz:
(We would still lose the print, although we could probably rewrite it to be a GPU-safe print since either way this kernel will not be that expensive.)

PR #1985, opened 2021-09-11
don't average down Simplified_SDC_React_Type

zingale:
this is not a conserved quantity and it is best to use the value
originally computed on each grid as the predictor for the next
step

## PR checklist


zingale:
I had thought that this would help with the issue of mass fractions being outside of [0, 1] in some cases when using SDC, but it appears that it does not.  Still need more testing...

zingale:
if this does not help with the negative mass fractions, then I think #1952 is the correct fix.

I still want to check to see if I encounter this problem without AMR.

zingale:
I no longer think that this is needed.

PR #2000, opened 2021-09-21
Remove max_radius_all_in_domain

maxpkatz:
## PR summary

This code path appears to be trying to deal with the case where there might be mass outside the domain, but we don't currently have that in the code, so by removing it we can simplify the calculation of the monopole gravity.

## PR checklist


maxpkatz:
I don't really know what it is trying to do, other than that it might have something to do with the corners.

zingale:
errors are roundoff: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-09-21-002/index.html

zingale (approved):

PR #2030, opened 2021-11-01
adjusting nova settings

aisclark91:
## PR checklist


zingale:
this doesn't run for me because `castro.small_temp` is larger than the lowest temperature in the model.

Also, can you remove whichever inputs file is not used here.

Issue #2034, opened 2021-11-09
Trouble building Sedov with MPI (mpich)


joehellmers:
I'm getting the following error
```
Compiling filfc.f90 ...
mpif90  -g -O3 -ffree-line-length-none -fno-range-check -fno-second-underscore -fimplicit-none  -Jtmp_build_dir/o/3d.gnu.MPI.EXE -I tmp_build_dir/o/3d.gnu.MPI.EXE -Itmp_build_dir/s/3d.gnu.MPI.EXE -I. -I../../../external/amrex/Src/Base -I../../../external/amrex/Src/Base/Parser -I../../../external/amrex/Src/AmrCore -I../../../external/amrex/Src/Amr -I../../../external/amrex/Src/Boundary -I../../../external/Microphysics/util -I../../../external/Microphysics/util/gcem/include -I. -I../../../Source/driver -I../../../Source/hydro -I../../../Source/problems -I../../../Source/sources -I../../../Source/mhd -I../../../external/Microphysics/EOS -I../../../external/Microphysics/EOS/gamma_law -I../../../external/Microphysics/networks/general_null -I../../../external/Microphysics/EOS -I../../../external/Microphysics/networks -I../../../external/Microphysics/interfaces -I../../../external/Microphysics/constants -Itmp_build_dir/castro_sources/3d.gnu.MPI.EXE -I../../../external/amrex/Tools/C_scripts  -c ../../../Source/driver/filfc.f90 -o tmp_build_dir/o/3d.gnu.MPI.EXE/filfc.o
../../../Source/driver/filfc.f90:3:6:

    3 |   use amrex_fort_module, only : rt => amrex_real
      |      1
Fatal Error: Cannot open module file ‘amrex_fort_module.mod’ for reading at (1): No such file or directory
compilation terminated.
make: *** [../../../external/amrex/Tools/GNUMake/Make.rules:250: tmp_build_dir/o/3d.gnu.MPI.EXE/filfc.o] Error 1 
```

My GNUMakefile looks like this
```
PRECISION  = DOUBLE
PROFILE    = FALSE

DEBUG      = FALSE

DIM        = 3

COMP       = gnu

USE_MPI    = TRUE
USE_OMP    = FALSE
USE_CUDA   = FALSE
USE_MHD    = TRUE

USE_FORT_MICROPHYSICS := FALSE
BL_NO_FORT := TRUE

# define the location of the CASTRO top directory
CASTRO_HOME  := ../../..

[...]

PR #2037, opened 2021-11-11
enable flattening for MHD Sedov

zingale:
this gets negative densities from the start without flattening

## PR checklist


zingale:
this eliminates the retries reported in #2034.

I want to still look at this some more to see what is happening in the reconstruction.

guadabsb15:
I believe we want to set `castro.ppm_type=1 ` in `inputs.3d.mhd`

zingale:
yes, I will add that as well.

guadabsb15 (approved):

PR #2047, opened 2021-12-01
inelastic scattering is not used since we removed neutrinos

zingale:
## PR checklist


zingale:
tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2021-12-01-005/index.html

maxpkatz:
I don't think this code is related to neutrinos; it appears to implement Compton scattering of photons on matter ([this](https://www.sciencedirect.com/science/article/pii/0021999185900701?via%3Dihub) is the paper it's citing, which I don't have).

maxpkatz (approved):

maxpkatz:
Approving despite the above comment because to our knowledge the code path is unused and it's certainly not tested.

PR #2079, opened 2022-01-28
for the helmholtz inputs for Sedov, fix the tagging

zingale:
we were not creating any levels

## PR checklist

### DIFF
--- a/Exec/hydro_tests/Sedov/inputs.mini-Castro
+++ b/Exec/hydro_tests/Sedov/inputs.mini-Castro
@@ -38,15 +38,15 @@ amr.regrid_int      = 2       # how often to regrid
 amr.blocking_factor = 128     # block factor in grid generation
 amr.max_grid_size   = 128
 
-amr.refinement_indicators = dengrad_rel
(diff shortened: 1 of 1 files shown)


Exec/hydro_tests/Sedov/inputs.mini-Castro
```diff
@@ -44,6 +44,10 @@ amr.refine.dengrad_rel.max_level = 20
 amr.refine.dengrad_rel.relative_gradient = 0.25
 amr.refine.dengrad_rel.field_name = density
 
+amr.refine.dengrad_rel.max_level = 20
+amr.refine.dengrad_rel.relative_gradient = 0.25
+amr.refine.dengrad_rel.field_name = Temp
```
maxpkatz, line 49: I think what you want here is to add a `tempgrad_rel` to `refinement_indicators` and update the name for these lines appropriately (or replace line 45 with `Temp` instead of `density` and rename all of the `dengrad_rel` to `tempgrad_rel`).

zingale: updated

maxpkatz (approved):

Issue #2080, opened 2022-02-05
burning problems on Perlmutter


zingale:
When running a problem with reactions on NERSC Perlmutter, we seem to encounter too many subcycles.  It seems that the burn is not working properly (Castro reports that the burn was unsuccessful).

You can reproduce this as follows:

```
module load PrgEnv-gnu cudatoolkit
```

in `Castro/Exec/science/subchandra`:

```
make NETWORK_DIR=subch USE_CUDA=TRUE COMP=gnu -j 8
```

run with `inputs_2d.N14` using:

```
#!/bin/bash
#SBATCH -A m3018_g
#SBATCH -C gpu
#SBATCH -J Castro
#SBATCH -o sedov_%j.out
#SBATCH -t 1:00
#SBATCH -N 4
#SBATCH --ntasks-per-node=4
#SBATCH --gpus-per-task=1
#SBATCH --gpu-bind=map_gpu:0,1,2,3

srun -n 16 ./Castro2d.gnu.DEBUG.MPI.CUDA.ex inputs_2d.N14 castro.time_integration_method=0 integrator.atol_spec=1.e-5 integrator.jacobian=2
```

zingale:
the same issue happens if we run with SDC

zingale:
I'm able to get through a bit of burning when running with cuda-memcheck (I time out before the crash seen above)

zingale:
The same problem occurs with version 21.12 of the code (that is the first tag where the C++ subch net is available)

zingale:
flame_wave seems to work with inputs_2d.testsuite even if I bump to `amr.max_level=2`
maybe this is a sub-chandra problem?

zingale:
update: this problem works fine on groot (Titan V), so there is something odd with Perlmutter

zingale:
I don't seem to have problems one 1 GPU on Perlmutter (it does fail a burn but recovers without taking too many subcycles).

So maybe there is an MPI issue?

zingale:
running the problem with `compute-sanitizer` didn't turn up any issues

zingale:
Note we still hit the problem if we compile with `DEBUG=TRUE`

zingale:
dropping down to `CUDA_ARCH=70` doesn't fix things

zingale:
running this on groot (on CPU) with valgrind also turns up no issues

maxpkatz:
[...]

PR #2091, opened 2022-03-15
Prefetch radial gravity data before and after MPI reduction

maxpkatz:
## PR summary

When GPU-aware MPI is not being used, it is not generally safe to pass managed memory directly to an MPI call (depending on MPI implementation, this could lead to accessing invalid data). So we need to prefetch the data to the host before calling ParallelDescriptor reduction operations. (This is handled already for multipole gravity, but this was missed when monopole gravity was ported to GPUs.)

This addresses the issue identified in #2080.

## PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-# 22.03
+# 22.04
 
    * Fixed an issue with monopole gravity where running with multiple
      MPI ranks in a GPU build could result in an incorrect gravitational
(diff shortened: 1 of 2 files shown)


maxpkatz:
Depends on AMReX-Codes/amrex#2655.

CHANGES.md
```diff
@@ -1,3 +1,9 @@
+# 22.03
+
```
zingale, line 2: this should be 22.04, since that will be the next tag

PR #2093, opened 2022-03-15
remove PRIM_SPEC_HAVE_SOURCES from simplified-SDC

### LINKED ISSUE #1987: decouple PRIM_SPECIES_HAVE_SOURCES from SDC
Once #1965 is in, we no longer trace the SDC species sources, so they are no longer included in any of the code that requires PRIM_SPECIES_HAVE_SOURCES.  This means that we can remove the requirement that simplified-SDC enables this option and it will only remain for the currently unused case of species have pure-hydrodynamical sources.

zingale:
we no longer trace under the any of the simplified-SDC sources,
including species, so we don't need to explicitly allocate memory
for them anymore with simplified-SDC

## PR summary

This closes #1987 

## PR checklist


zingale:
for some reason I get a segfault on the fillpatch of the SDC react source with this change :(

zingale:
okay, this works now, and gives no diffs, as expected:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-03-16-001/index.html

maxpkatz (approved):

Issue #2096, opened 2022-03-21
invalid mass fraction error when running flame wave model with aprox13 or triple alpha network.


zhichen3:
I'm encountering the error: `amrex::Error::591::Invalid mass fraction in Castro::normalize_species() !!!`, when running the flame_wave model with aprox13 or triple_alpha_plus_cago network. This error occurred when I added `integrator.jacobian = 2` and `castro.retry_small_density_cutoff = 1` to the default input file in git, specifically: `inputs_He/inputs.He.1000Hz`. Whether or not these extra parameters caused the issue remains unknown.

zingale:
update: this was not seen on GPUs / summit, because that check is disabled on GPUs.

Also, it seems that the problem is happening during the reflux step.

zingale:
this is happening at the `expand_state()` call in `initialize_do_advance` midway through the level 1 advance (where level 1 is 4x finer than level 0)

zingale:
even if we run with `castro.state_interp_order = 0`, we still hit this issue:

```
   
... compute diffusive term at level 1
   
... compute diffusive term at level 1
(i, j, k) = 2 24 0 , X[0] = 1.930137441
(i, j, k) = 2 24 0 , X[1] = -0.5402948753
(i, j, k) = 2 24 0 , X[12] = -0.3887283302
Castro::do_new_sources() time = 0.300138036
```

zingale:
here's the backtrace:

```
 0: ./Castro2d.gnu.TEST.MPI.ex() [0x66b6b6]
    amrex::BLBackTrace::print_backtrace_info(_IO_FILE*)
/home/zingale/development/amrex//Src/Base/AMReX_BLBackTrace.cpp:179:39

 1: ./Castro2d.gnu.TEST.MPI.ex() [0x66d5b4]
    amrex::BLBackTrace::handler(int)
/home/zingale/development/amrex//Src/Base/AMReX_BLBackTrace.cpp:85:15

 2: ./Castro2d.gnu.TEST.MPI.ex() [0x43ee68]
    amrex::Error(char const*) inlined at ../../../Source/driver/Castro.cpp:3022:21 in Castro::normalize_species(amrex::MultiFab&, int
)
/home/zingale/development/amrex//Src/Base/AMReX.H:120:5
Castro::normalize_species(amrex::MultiFab&, int)
../../../Source/driver/Castro.cpp:3022:21

[...]

PR #2100, opened 2022-03-27
Normalize species fluxes on reflux

maxpkatz:
## PR summary

This is intended to address #2096, which is observed after a reflux.

## PR checklist


zingale:
this is no longer necessary after #2120, #2122 , #2123 , and further this is not a good idea anyway.

PR #2106, opened 2022-04-15
Enable NaN check in the GPU build

maxpkatz:
## PR summary

The Fab and MultiFab `contains_nan` methods are GPU safe, we just want to avoid the bit that iterates through the memory and prints out the zone value because that's not safe for non-managed-memory (though we could add that back for managed memory builds if needed).

## PR checklist


zingale (approved):

zingale:
tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gpu/2022-04-15-002/index.html

PR #2113, opened 2022-04-18
Radsolve default values should be set before overrides

maxpkatz:
## PR checklist


zingale (approved):

maxpkatz:
As originally written, this change unintentionally overrode the level_solver_flag in 1D, [leading to diffs](http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-04-17/index.html) for inputs files that were not using Hypre SMG (radsolve.level_solver_flag = 0). To simplify the code and avoid the regression test diff, this override has been removed; there is already an abort later in this routine that covers the unsupported case radsolve.level_solver_flag = 1.

zingale (approved):

PR #2114, opened 2022-04-18
Fix box bounds in update_dcf

maxpkatz:
## PR summary

Tiling was originally added to update_dcf in commit c54638feb94ba9172b72b04704d9f5dd2194a9b6. However, I think there was a bug in this commit. The dcf array has a single ghost zone, but not all of the source arrays filling it do, and the original Fortran looped from dcf_lo+1 to dcf_hi-1, which would have been only the valid domain. But this commit used a growntilebox() for the MFIter loop, which is a change from the Fortran and one that will not make sense unless all of the source MultiFabs have a ghost zone too.

## PR checklist


zingale (approved):

zingale:
tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-04-18-002/index.html

PR #2118, opened 2022-04-22
use one-sided PPM stencils at reflecting BCs

zingale:
## PR checklist


zingale:
here's a 1-d flame_wave atmosphere run via `hse_convergence_general` on `development` with reflecting BCs:

hse

and here's the version with these new one-sided PPM stencils:

hse_onesided

Velocity is on the y-axis and height in the atmosphere in on the x-axis.

zingale:
the 2-d axisymmetric Sedov problem NaNs with these new BCs.

zingale:
this seems to work well for the HSE boundary, but it does not see to work with the -x boundary in axisymmetric coords.  This might be because we need to take into account the geometry somehow, like this paper suggested: https://ui.adsabs.harvard.edu/abs/1993ApJS...88..589B/abstract

zingale:
as an example, Sedov 2-d in axisymmetry crashes after a while.

zingale:
this has been updated to work with pslope

zingale:
I think the MHD issues are fixed in #3109

PR #2121, opened 2022-04-23
Issue a retry when X is invalid after a hydro update

### LINKED ISSUE #1952: Negative species should trigger a retry

maxpkatz:
## PR summary

(rho X) can be negative after a hydro update even if rho itself is not, for example due to transverse terms being applied near sharp composition gradients. We should retry with a smaller timestep in that scenario.

## PR motivation

Closes #1952

## PR checklist

### DIFF
--- a/Source/driver/Castro_advance_ctu.cpp
+++ b/Source/driver/Castro_advance_ctu.cpp
@@ -174,57 +174,81 @@ Castro::do_advance_ctu(Real time,
       construct_ctu_mhd_source(time, dt);
 #endif
 
-      // Check for small/negative densities.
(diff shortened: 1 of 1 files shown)


maxpkatz:
This should address one of the two failure modes identified in #2096 (although it could still lead to an abort if too many subcycles are needed to resolve the problem).

zingale (approved):

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-04-23-001/index.html

PR #2123, opened 2022-04-24
Turn off flux correction if it would cause X < 0 or X > 1

maxpkatz:
## PR summary

The reflux currently zeros out any flux corrections that would cause a negative or small density. A check is added to this that also turns off flux corrections if they would cause X > 1 or X < 0 (within some tolerance). Since it is possible that multiple flux corrections at the same zone center could conspire together to cause this, the only general way to do this is to loop over zone centers and consider all adjacent fluxes simultaneously, so this code is switched to a zone center loop instead of a zone face loop (and the same logic is now applied to the density check, which should also make it more accurate).

## PR motivation

This should address #2096.

## PR checklist


zingale:
tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-04-23-002/index.html

zingale (approved):

PR #2141, opened 2022-04-27
Pr nova

aisclark91:
Here I update the nova problem parameters I am using in my runs.


Exec/science/nova/problem_initialize.H
```diff
@@ -27,13 +27,23 @@ void problem_initialize ()
...
         problem::xloc_vortices[i] = (static_cast<Real>(i) + 0.5_rt) * offset + problo[0];
     }
+
+    problem::ih1 = network_spec_index("hydrogen-1");
+    problem::layer_h1_base = 0.0_rt;
+
+    for (int n = 0; n < model::npts; n++) {
+        if (model::profile(0).state(n, model::ispec+problem::ih1) > 0.5_rt) {
+            problem::layer_h1_base = model::profile(0).r(n);
+            break;
```
zingale, line 44: you never declare `problem::layer_h1_base` -- that should be added to the `_prob_params` file

aisclark91: yes, I forgot to add it  _prob_params! In the run is added, otherwise, it will not work.

Exec/science/nova/_prob_params
```diff
@@ -7,11 +7,15 @@ velpert_scale       real            1.e2_rt       y
...
-velpert_height_loc  real            6.5e3_rt      y
-
 num_vortices        integer         1             y
 
 max_num_vortices    integer         10            n
 
 xloc_vortices       real            0.0_rt        n     10
 
+ih1		    integer         0             y
+
```
zingale, line 17: both of these (ih1 and layer_h1_base) should have "n" as the last parameter -- they are not something that is set by the user

zingale (approved):

PR #2143, opened 2022-04-28
Add a per-zone flattening implementation and use it in PPM

maxpkatz:
## PR summary

This PR adds a header implementation of the flattening routine implemented on a per-zone basis (the existing flattening routine is done on a per-box basis). This new version of the function is used in the hydro instead of the flatn array.

## PR motivation

This aligns with a goal of limiting the total number of temporary Fabs created in the hydro. At the moment, we don't actually get rid of the flatn and flatg arrays, because they will still be needed for PLM, so for the moment we will have duplicative work and redundant memory. (This helps keep the PR short and readable.) But if this PR is valid and merged, we can then do the same thing in PLM in a second PR, and then after that remove the temporary array and the call to uflatten.

In general, our strategy should be to only have temporary Fab data for things that are either (a) used more than once and/or (b) involve accesses to multiple zones in the same array in a stencil-like operation, so that the cost of accessing the data is amortized. As currently implemented the flattening satisfies neither of these properties, it's only used once and the PPM update for a zone only needs the flattening coefficient at that zone.

## PR checklist

### DIFF
--- a/Source/hydro/Make.package
+++ b/Source/hydro/Make.package
@@ -9,6 +9,7 @@ CEXE_sources += Castro_ctu.cpp
 CEXE_sources += Castro_mol.cpp
 CEXE_headers += advection_util.H
 CEXE_sources += advection_util.cpp
+CEXE_headers += flatten.H
 CEXE_sources += flatten.cpp
 
 ifeq ($(USE_TRUE_SDC),TRUE)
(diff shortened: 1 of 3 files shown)


zingale (approved):

zingale:
all tests pass:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-04-28-003/index.html

zingale (approved):

Issue #1691, opened 2021-05-08
can we switch the fluxes MF to a fluxregister?


zingale:
Right now, we create a MF `fluxes` where we add the fluxes during a subcycle, eventually copying them into the flux register.  The hydro routines do not operate on the flux register directly, but instead increment the fluxes.  We need this intermediate array to allow for the retry functionality.  But this requires a lot of memory.  Can we get away with replacing the fluxes MF with a second flux register?

maxpkatz:
This can be done with #2147 as a first step, followed by a subsequent change to remove references to the `fluxes` MultiFab in post_timestep (which, I think, ends up being unnecessary since the only flux that is currently indirectly used in post_timestep is the mass fluxes array). Then we can drop the `fluxes` MultiFab entirely and handle this with a flux register. In exchange, we will need to maintain a second copy of the mass_fluxes array (to distinguish the current subcycle's fluxes from the full timestep's fluxes). But this is a good tradeoff overall.

PR #2147, opened 2022-04-28
Update mass fluxes at end of timestep with full timestep's worth

maxpkatz:
## PR summary

This update makes the mass_fluxes array hold the full timestep's worth of mass fluxes at the end of finalize_advance. (This subtlety only matters if there were multiple subcycles in the timestep, in which case mass_fluxes would presently only have the last subcycle's worth of mass fluxes.)

## PR motivation

This is a first step towards implementing the request in #1691.

## PR checklist


zingale:
tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-04-28-005/index.html

Source/driver/Castro_advance.cpp
```diff
@@ -501,6 +501,16 @@ Castro::finalize_advance()
         FluxRegFineAdd();
     }
 
+    // The mass_fluxes array currently holds only the fluxes
+    // from the last subcycle. Override this with the sum of
+    // the fluxes from the full timestep (this will be used
+    // later during the reflux operation).
+
```
zingale, line 508: the PR says this change makes mass_fluxes hold the full timestep's worth of fluxes -- doesn't this comment say the opposite?

maxpkatz: It is assuming that `fluxes`, which is about to be copied to `mass_fluxes`, has the full timestep's worth of flux data at this point.

zingale: ah, I understand.

zingale (approved):

zingale:
tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-10-27-001/index.html

PR #2155, opened 2022-04-29
Fix GPU compilation action

yut23:
## PR checklist


zingale:
maybe we should just switch to this action?
https://github.com/Jimver/cuda-toolkit

or steal whatever amrex does

zingale:
https://github.com/AMReX-Codes/amrex/blob/development/.github/workflows/dependencies/dependencies_nvcc11.sh

^ that's the script that amrex uses to get CUDA installed

zingale (approved):

PR #2174, opened 2022-05-03
Inline get_react_source_prim

### LINKED ISSUE #2156: NQ >= NUM_STATE assumption in SDC is not true when HYBRID_MOMENTUM=TRUE
The copy from S_new into S_noreact for simplified SDC assumes that NQ >= NUM_STATE, as noted in this comment:

https://github.com/AMReX-Astro/Castro/blob/910da6ab7db14af3e940f0d4caa2ad5050e0c5d6/Source/driver/Castro_advance_ctu.cpp#L340-L344

This is not true when USE_HYBRID_MOMENTUM=TRUE, because that adds three new conserved state variables which don't have any primitive counterpart. So this code will result in out-of-bounds accesses.

maxpkatz:
## PR summary

Fixes #2156 

## PR checklist


maxpkatz:
Note: this will not work if USE_SIMPLIFIED_SDC=TRUE and USE_RAD=TRUE. I can try and fix that if we think there's an immediate need for radiation with simplified SDC.

zingale:
Let's not worry about SDC and radiation, just make sure it aborts if we try that.

zingale:
some diffs are big: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-05-03-003/index.html

Source/reactions/Castro_react.cpp
```diff
@@ -461,6 +476,18 @@ Castro::react_state(Real time, Real dt)
             }
 #endif
 
+            // Convert the current state to primitive data.
+            // This state is U* = U_old + dt A where A = -div U + S_hydro.
+
+            Array1D<Real, 0, NQ-1> q_noreact;
+            Array1D<Real, 0, NQAUX-1> qaux_noreact;
+
+            hydro::conservative_to_primitive(i, j, k, U_new,
```
zingale, line 485: should this be SDC_react instead of U_new?

maxpkatz: No -- in this version of the code, we never need to do the copy of the state into SDC_react. At this point in the code flow, U_new still represents U_old + dU_hydro + dU_sources. It's only after calling `burner` that this is modified. So this allows us to do what we were doing before without ever needing to make a temporary MultiFab copy of U_new.

maxpkatz:
> some diffs are big: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-05-03-003/index.html

This appears to be due to some tests changing whether or not they have retries.

maxpkatz:
> Let's not worry about SDC and radiation, just make sure it aborts if we try that.

I looked at this, but concluded that to do it right, we need to finish up the radiation C++/GPU port (specifically of lamborder, to make it thread-safe). At the moment, the code will fail to compile, but we aren't using that combination yet anyway. So I think we should just defer this for now. Though I can insert a hack if needed.

zingale (approved):

zingale:
reacting_convergence convergence is still the same:

```
Level  L1 norm of Error in Each Component
-----------------------------------------------
level: 0
                 density        2.783876e+18       2.047638   6.733635e+17
                    xmom        6.779154e+26       2.447429   1.242871e+26
                    ymom        6.779154e+26       2.447429   1.242871e+26
                    zmom        0.000000e+00   ------------   0.000000e+00
                   rho_E        2.449619e+35       2.352854   4.795335e+34
                   rho_e        2.256441e+35       2.320177   4.518365e+34
                    Temp        2.232080e+21       1.695162   6.893111e+20
                 rho_He4        2.865944e+18       2.023540   7.048900e+17
                 rho_C12        1.694517e+17       1.959464   4.357008e+16
                 rho_O16        1.681287e+14       1.662013   5.312834e+13
                rho_Fe56        2.783868e+08       2.047633   6.733634e+07
```

PR #2176, opened 2022-05-03
Remove castro.show_center_of_mass

maxpkatz:
## PR summary

The center of mass is now calculated and displayed by default if we're already doing sums. The additional marginal cost of this sum is not high, so we should just simplify the code.

## PR checklist


zingale:
quick tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-05-03-002/index.html

zingale (approved):

PR #2193, opened 2022-05-11
Adding High Temperature and Enriched Nova Initial Models

aisclark91:
I have enforced HSE, in the Glasner et.al (2007) enriched, T5 and T7 models.


zingale (approved):

zingale:
can you fix that conflict?

PR #2194, opened 2022-05-12
Store rotation period in the checkpoint file

maxpkatz:
## PR summary

This is upstreamed from wdmerger. For most problems it will have no effect, because most problems don't update the rotation period from the value in the inputs.

## PR checklist


zingale:
quick tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-05-12-001/index.html

PR #2205, opened 2022-05-18
turn off use_pslope by default

zingale:
this only affects ppm_type = 0, but we will soon extend its functionality
to PPM as well.  Disabling it by default means that the behavior of most
problems will not change.  For the inputs files that use gravity and have
ppm_type = 0 currently, I've explicitly re-enabled it.

## PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,7 @@
      (castro.ppm_type = 0) implementation.  Since that method is not the
      default, it is unlikely that this has been used.  This change is being
      done to allow for a PPM implementation to be added without changing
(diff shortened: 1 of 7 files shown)


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-05-18-001/index.html

looks like I need to just fix hse_convergence-SDC2

zingale:
this seems to work now: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-05-18-003/index.html

maxpkatz (approved):

PR #2209, opened 2022-05-20
Add support for early stopping based on exceeding some threshold value

maxpkatz:
## PR summary

This was originally developed for wdmerger and Detonation, but now all applications should be able to use it.

`amr.message_int` is overridden to be 1 by default to support this.

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-05-20-002/index.html

I think the diffs are because it doesn't have all of the changes on development

zingale (approved):

zingale:
can you also update the docs here:
https://amrex-astro.github.io/Castro/docs/timestepping.html

maxpkatz:
> can you also update the docs here: https://amrex-astro.github.io/Castro/docs/timestepping.html

I could, but this isn't really a timestepping constraint, so it seems a little out of place.

zingale:
okay.  I just thought since at the top we list stopping criteria.  But we should find somewhere else to doc this.

PR #2214, opened 2022-05-25
fix PPM use_pslope at the top of an atmosphere

zingale:
## PR summary

This fixes the reconstruction at the top of the atmosphere where p_hse might go negative.  It also adds an inputs file
for flame_wave that runs with this change and reflecting BCs.

## PR checklist


zingale:
here's a 25 cm flame_wave run that uses this change -- it seems to hold together well.  Also takes bigger steps than the non-pslope version (slightly).

flame_wave_1000Hz_25cm_smallplt28420_slice

maxpkatz (approved):

PR #2218, opened 2022-05-31
Fixed wrong variable names

AMReX-Fusion:
Hi,

I noticed that there are actually wrong variable names in the documentation.
i.e. QGDRHO does not exist in the code, but GDRHO does
Similar with others.

Cheers


zingale (approved):

zingale:
thanks!

PR #2219, opened 2022-06-02
Use hydro limiter on negative density in reflux

maxpkatz:
## PR summary

Using a zone-center based limiter in the reflux ends up not working in every case. The flux lives in the cell faces, which means it is dual-valued in MultiFab storage, so both versions of the flux have to be consistent. When looping over cell centers, we might limit the flux in one direction and not another, and if the second one is the one that actually gets applied in the reflux, the limiter will be ignored. So instead we FillPatch the state to add a ghost zone and then apply an edge-based limiter on that state. This PR does it for density and a subsequent PR will apply this to the invalid X check.

## PR checklist

### DIFF
--- a/Source/driver/Castro.cpp
+++ b/Source/driver/Castro.cpp
@@ -2607,6 +2607,12 @@ Castro::reflux(int crse_level, int fine_level)
 
         MultiFab& crse_state = crse_lev.get_new_data(State_Type);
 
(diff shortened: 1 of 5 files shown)


zingale:
tests are good: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-06-02-001/index.html

zingale (approved):

PR #2220, opened 2022-06-02
Move to edge-based strategy for invalid X check in reflux

maxpkatz:
## PR summary

This operates in a similar manner to the edge-based flux limiter. Similar to that, we have to use a safety factor of AMREX_SPACEDIM since up to 3 faces could be updating a given zone in the reflux.

Since the new value of X depends on the new value of rho, which will also be affected by the fluxes, it's not clear whether the factor of AMREX_SPACEDIM should also be applied to the denominator when calculating the new X (i.e. whether it should be included in drho). I've chosen not to do that, but perhaps that could be done in a later change.

## PR checklist

### DIFF
--- a/Source/driver/Castro.cpp
+++ b/Source/driver/Castro.cpp
@@ -2676,86 +2676,47 @@ Castro::reflux(int crse_level, int fine_level)
             auto mask_arr = mask[mfi].array();
 
             // Limit fluxes that would cause a small/negative density.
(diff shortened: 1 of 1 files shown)


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-06-02-002/index.html
this just catches small diffs in flame_wave

zingale (approved):

PR #2225, opened 2022-06-07
Fix test for pre-advance timestep check

maxpkatz:
## PR summary

This should run on everything but the very first fine timestep in a given coarse step.

Also, #2222 missed that estTimeStep always operated on the new-time state, so all of the estdt interfaces are updated to accept an `is_new` parameter and use the old-time state if it is set to zero. The pre-advance check then uses the old-time state as intended.

## PR checklist

### DIFF
--- a/Source/driver/Castro.H
+++ b/Source/driver/Castro.H
@@ -548,39 +548,39 @@ public:
 ///
 /// Estimate time step.
 ///
-    amrex::Real estTimeStep ();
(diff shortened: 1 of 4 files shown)


zingale (approved):

maxpkatz:
Tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-06-07-004/index.html

PR #2230, opened 2022-06-10
t = 0.0 should be std::fixed in diags

maxpkatz:
## PR checklist


maxpkatz:
Note: the Sedov action fails, but this is expected because thisc change is updating the format of the grid_diag.out file that the action uses for comparison.

PR #2250, opened 2022-07-22
don't do source correction with SDC on the first iteration

zingale:
this will now only use the information from the old iteration
of the current timestep, if available.  Also note that we were
accessing the wrong time-level of the Source_Type.

## PR checklist


zingale:
this follows up on #1968

zingale:
this still gives very bad results on `reacting_bubble`

zingale:
closing in favor of #2252

PR #2252, opened 2022-07-24
fix the source term corrector with SDC

zingale:
for iterations after the first, we need to use the "new time" source,
since there is no swap between iterations.

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-07-24/index.html

the wdmerger test fails, as expected, since it is the only one that uses the source term predictor.

maxpkatz (approved):

PR #2270, opened 2022-08-12
move Schlieren plot into main routine

zingale:
also do it as a derived field -- this now works better with jumps in refinement.  Also the old
version was only using a level 0 covering grid.

## PR checklist


zingale:
Here's the old plot:

subch_schlieren_old

and the new version (with this PR):

subch_lap_rho_sequence

maxpkatz (approved):

PR #2272, opened 2022-08-20
add a tool to output state at max enuc

zingale:
this doesn't yet work for parallel (MPI)

## PR checklist


zingale:
this should be moved to the new https://github.com/AMReX-Astro/amrex-astro-diag repo

Issue #328, opened 2018-05-08
output where and why the timestep limit is set


zingale:
We should have an option to print out where in the domain is the most restrictive timestep and what limiter is the cause (hydro, diffusion, reaction, ...)

maxpkatz:
We already print out which limiter is controlling the timestep, so this is half done already :-)

zingale:
it doesn't seem that we separate the diffusion timestep from the hydro timestep.  It would be nice to know which of these is at fault.

maxpkatz:
That is true.

I'll note that since we already perform two AllReduce calls to get both the hydro and reactions limited timesteps, and doing one for diffusion would add a third, we could combine these into a single ReduceRealMin and just do the minimum over all three at once (Weiqun added at some point the ability to hand initializer lists to the ParallelDescriptor routines).

maxpkatz:
Seems like MPI_MINLOC (https://www.mpi-forum.org/docs/mpi-1.1/mpi-11-html/node79.html) could be useful for the location information.

PR #2273, opened 2022-08-24
output the location where the timestep is set

zingale:
## PR summary

addresses #328

## PR checklist


zingale:
CPU tests are good: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-08-24-001/index.html

zingale:
GPU tests: http://groot.astro.sunysb.edu/Castro/test-suite/gpu/2022-08-24-001/index.html

this changes some answers -- not sure why.  But I saw the same phenomena in test_react.

zingale:
running on Summit, I see essentially no difference in the runtime doing a 3-d flame_wave for 100 steps

zingale:
note: all the limiters are now done except for radiation -- that one is a bit more complex, since it has 2 separate boxes.

zingale:
regarding the GPU diffs, I wonder if `TilingIfNotGpu` is the issue?

zingale:
updated tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-08-29-001/index.html

Source/driver/timestep.cpp
```diff
@@ -224,16 +216,15 @@ Castro::estdt_mhd (int is_new)
       dt3 = dt1;
 #endif
 
-      return {amrex::min(dt1, dt2, dt3)};
+      return {ValLocPair<Real, IntVect>{amrex::min(dt1, dt2, dt3), idx}};
 
-    });
+  });
 
+  if (ParallelDescriptor::IOProcessor()) {
```
WeiqunZhang, line 223: This could be misleading, because MPI communication has not been done yet and the final constraint may come from other processes.

zingale: ah indeed... I should move that up to the caller.  But that won't explain the diffs when I run on GPUs.

WeiqunZhang:
Is this test (http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-08-29-001/nova.html) the issue?  There should not be any roundoff errors in min and max.

zingale:
this is one of the failures:
http://groot.astro.sunysb.edu/Castro/test-suite/gpu/2022-10-27-002/flame_wave.html

it only has diffs on GPUs

WeiqunZhang:
Oh.  This, I guess. http://groot.astro.sunysb.edu/Castro/test-suite/gpu/2022-08-24-001/dustcollapse-restart.html

WeiqunZhang:
I think the roundoff errors come from the code that computes the number for reduction.  The results agree with each other if I run in debug mode.

zingale:
thanks @WeiqunZhang !

Source/driver/timestep.cpp
zingale: actually, I don't understand this.

Doesn't the ParReduce already do the reduction over all of the processors?  So isn't this the final value at this point?

zingale:
okay.  Is there a ParallelReduce that can work on `ValLocPair<double, amrex::IntVect>` then?

WeiqunZhang:
We don't.  We need to create custom MPI_Datatype and MPI_Op for that.  For a generic `ValLocPair<T1,T2>`, there is probably not much amrex can do.  However, `ValLocPair<double,amrex::IntVect>` might be the most one that we can add support in amrex.

zingale:
okay, I think I need something like that, because otherwise, I am not sure how to get the location associated with the dt reduced along with dt.

zingale:
actually, I guess I can do it manually, but keeping the reduce on dt and then having each proc output the location if their local reduced dt is the global one.

WeiqunZhang:
https://github.com/AMReX-Codes/amrex/pull/3003

With the PR above, you can do
```
ParallelReduce::Min(r, ParallelDescriptor::IOProcessorNumber(), ParallelDescriptor::Communicator());
ParallelAllReduce::Min(r, ParallelDescriptor::Communicator());
```

zingale:
tests pass:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-10-29-003/index.html

this is ready for review

Source/driver/Castro.cpp
```diff
@@ -1545,28 +1545,34 @@ Castro::estTimeStep (int is_new)
...
 #else
-          estdt_hydro = estdt_cfl(is_new);
+          auto hydro_dt = estdt_cfl(is_new);
 #endif
 
+          amrex::ParallelAllReduce::Min(hydro_dt, MPI_COMM_WORLD);
+          estdt_hydro = amrex::min(estdt_hydro, hydro_dt.value) * cfl;
+          if (verbose) {
+              amrex::Print() << "...estimated hydro-limited timestep at level " << level << ": " << estdt_hydro << std::endl;
+              amrex::Print() << "...hydro CFL timestep constrained at (i,j,k) = " << hydro_dt.index << std::endl;
```
maxpkatz, line 1569: Can you synchronize the text here ("hydro-limited timestep" rather than "hydro CFL timestep") and also reduce the "(i,j,k)" for 1D/2D?

maxpkatz (commented):
The AMReX submodule should also be updated

Source/driver/Castro.cpp
zingale: done

zingale:
AMReX module updated

maxpkatz (approved):

PR #2284, opened 2022-10-28
Remove CFL violation check for MHD CTU advance

maxpkatz:
## PR summary

This follows the same logic as #1873. We are already implicitly checking for CFL violations at the beginning and end of the timestep, so an additional check inside the CTU update is unnecessary.

## PR checklist


zingale:
tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2022-10-27-001/index.html

zingale (approved):

PR #2287, opened 2022-10-28
use a prediction of the reactive term on first simple SDC iter

zingale:
## PR checklist


zingale:
an alternate approach here, for detonations, it to zero out I_q on the first iter and then take an extra SDC iteration.

PR #2301, opened 2022-11-09
Extending the y domain to include the buffer

aisclark91:
I decided to make two important changes in the t9 and t7 setup:

1. The initial models are now expressed in terms of the highest allowed resolution of the model. In this case. the highest resolution in both initial models can be supported is 0.1km.

2. We extended the y domain to incorporate a `rho=1.0e-5` zone, leaving enough space for the buffer.


zingale:
can you remove the old models that are no longer needed?

aisclark91:
yes, I will do it in a separate PR

zingale (approved):

PR #2327, opened 2022-12-12
Convert science/flame to new problem diagnostics scheme

yut23:
## PR summary

The code is the same as toy_flame (see #1458)

## PR checklist


zingale (approved):

zingale:
this is okay, but we should note that the properties are defined for a He flame -- the flame speed is based on the consumption of the primary fuel, and really is only an estimate for stuff that has multiple burning stages.  This is based on Eqs. 5 and 6 here:
https://iopscience.iop.org/article/10.1086/383023/pdf

yut23:
Yeah, I noticed they looked somewhat off for my H/He flames. Anyways, my main goal with this PR is just to get the other diagnostics back.

PR #2337, opened 2023-01-17
[WIP] alternate fix for nse

zingale:
this initialize burn_state.e with UEINT before calling the burner with this change, we don't need to call the EOS in nse_update() But... this gives very different answers for massive_star than if we don't set burn_state.e and call the EOS (and much better).

## PR checklist


zingale:
with this change, we cannot do the EOS call that was added in https://github.com/AMReX-Astro/Microphysics/pull/1098

zingale:
this is not the correct fix.  The issue is with drive_initial_convection.  See #2344

PR #2338, opened 2023-01-18
fix #ifdef REACTION -> #ifdef REACTIONS

zingale:
## PR checklist


zingale:
we should have a script that looks at the `#ifdef`s and checks for misspellings.  I caught this via:

```
git grep "^#if" | cut -d ":" -f 2 | sort | uniq
```

maxpkatz (approved):

PR #2352, opened 2023-01-27
[WIP] Add chemical potentials to StateData

zhichen3:
## PR summary
Add chemical potentials to StateData for self-consistent NSE.

## PR checklist

### DIFF
--- a/Source/driver/Castro.cpp
+++ b/Source/driver/Castro.cpp
@@ -53,6 +53,9 @@
 #ifdef RADIATION
 #include <problem_initialize_rad_data.H>
 #endif
+#ifdef NSE_NET
+#include <problem_initialize_chem_data.H>
+#endif
 #include <problem_tagging.H>
 
 #include <ambient.H>
@@ -1039,7 +1042,7 @@ Castro::initData ()
 
 #ifdef NSE_NET
    MultiFab& ChemPot_new = get_new_data(Chemical_Pot_Type);
-   ChemPot_new.setVal(0.);
+   ChemPot_new.setVal(0.0, ChemPot_new.nGrow());
 #endif
    
 #ifdef MAESTRO_INIT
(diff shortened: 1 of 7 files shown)


Source/driver/Castro.cpp
```diff
@@ -1037,6 +1037,11 @@ Castro::initData ()
 #endif
 #endif
 
+#ifdef NSE_NET
+   MultiFab& ChemPot_new = get_new_data(Chemical_Pot_Type);
+   ChemPot_new.setVal(0.);
+#endif
```
zingale, line 1043: you should add `ChemPot_new.nGrow()` to the `setVal` so the ghost cells are also initialized to 0

Source/driver/Castro_setup.cpp
```diff
@@ -467,6 +467,14 @@ Castro::variableSetUp ()
 #endif
 #endif
 
+#ifdef NSE_NET
+  // Adding descriptor for chemical potentials used for NSE_NET
+  store_in_checkpoint = true;
+  desc_lst.addDescriptor(Chemical_Pot_Type, IndexType::TheCellType(),
+			 StateDescriptor::Point, 1, NCHEM,
+			 interp, state_data_extrap, store_in_checkpoint);
```
zingale, line 475: I am not completely sure we only want `1` ghost cell.  In the Strang burner, we call the burner on `ngrow_state` ghost cells, so I think that we need this to have that same amount.

Source/reactions/Castro_react.cpp
```diff
@@ -111,7 +117,12 @@ Castro::react_state(MultiFab& s, MultiFab& r, Real time, Real dt, const int stra
             burn_state.T = U(i,j,k,UTEMP);
 
             burn_state.T_fixed = -1.e30_rt;
-
+	    
+#ifdef NSE_NET
+	    burn_state.mu_p = ChemPot(i,j,k,CMUP);
+	    burn_state.mu_n = ChemPot(i,j,k,CMUN);
+#endif
+
```
zingale, line 125: I think that after the call to the burner, we need to store the updated chemical potentials back into `ChemPot`

Source/reactions/Castro_react.cpp
```diff
@@ -420,7 +438,11 @@ Castro::react_state(Real time, Real dt)
             burn_state.T = U_old(i,j,k,UTEMP);
 
             burn_state.T_fixed = -1.e30_rt;
-
+#ifdef NSE_NET
+	    burn_state.mu_p = ChemPot(i,j,k,CMUP);
+	    burn_state.mu_n = ChemPot(i,j,k,CMUN);
+#endif
```
zingale, line 444: likewise here, after the burn, we need to store the updated chemical potentials

PR #2356, opened 2023-02-01
add an action to check the ifdefs

zingale:
this makes sure that we don't mistype a define

## PR checklist


.github/workflows/check_ifdefs.py
```diff
@@ -0,0 +1,104 @@
...
+        print(d)
+
+    # now check to make sure that all the defines we found are okay
+
+    invalid = []
+    for d in defines:
+        if d not in good_defines:
+            invalid.append(d)
+
+    if invalid or total_errors > 0:
```
yut23, line 101: This makes it easier to see if there are bad defines:
```suggestion
    if invalid or total_errors > 0:
        if invalid:
            print("\ninvalid defines:")
```

.github/workflows/check_ifdefs.py
```diff
@@ -0,0 +1,104 @@
...
+    # read in the list of good defines
+
+    good_defines = []
+    with open(good_defines_file) as gd:
+        for line in gd:
+            good_defines.append(line.strip())
+
+    all_defines = []
+    total_errors = 0
+    for f in find_source_files():
```
yut23, line 73: I tried running this in my local repo, and got a bunch of errors because it picked up some of the generated files in `tmp_build_dir`.
```suggestion
    for f in find_source_files():
        if "tmp_build_dir" in f.parts:
            # skip generated files
            continue
```

zingale: indeed, that's not an issue in the action though, but I'll make the change

PR #2372, opened 2023-02-21
Adding Difussion

aisclark91:
In this PR, I added the required options and input considerations to include conductive radiation and conduction fluxes, according to Timmes (2000).


Exec/science/nova/inputs_nova_t7
```diff
@@ -102,14 +108,14 @@ integrator.abort_on_failure = 0
 
 amr.refinement_indicators = dens temp edot
 
-amr.refine.dens.max_level = 5
+amr.refine.dens.max_level = 3
 amr.refine.dens.value_greater = 1.0e2
```
zingale, line 112: you don't need to change these -- you can leave them high -- it just means that you always want these indicators to get up to max_level

aisclark91: I restored the values to their original value.

zingale (approved):

PR #2375, opened 2023-02-23
update the clang-tidy options

zingale:
## PR checklist

### DIFF
--- a/.github/workflows/c-linter.yml
+++ b/.github/workflows/c-linter.yml
@@ -36,7 +36,7 @@ jobs:
           if [[ -n "${HEADER_INCLUDES}" ]]; then exit 1; fi
 
       - name: Run cpp linter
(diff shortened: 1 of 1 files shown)


yut23:
It looks like `-header-filter=Castro` only checks header files with "Castro" in the name.

zingale:
oh interesting.  Not sure why that was added.  Maybe we should remove it?

yut23:
I've done some more digging and testing, and `-header-filter` does check part of the path as well. I'm pretty sure it matches against paths like `<include directory>/<file name>`, where `<include directory>` is one of the `-I` flags in compile_commands.json.

Based on that, something like `-header-filter='/Source/|/Util/model_parser_cxx|^\./'` should cover all of the non-generated Castro sources and the problem setups.

zingale:
It doesn't like that change to the header:

```
/bin/sh: 1: ^./: not found
407
/bin/sh: 1: /Util/model_parser_cxx: not found
408
Unable to run clang-tidy.```

yut23:
That's almost certainly a quoting issue. https://github.com/AMReX-Astro/cpp-linter-action/pull/2 should fix it.

zingale:
okay this works now.

zingale:
closing this in favor of #2381

PR #2379, opened 2023-02-25
address some clang-tidy issues

zingale:
mostly narrowing and nullptrs

## PR checklist


zingale:
quick tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-02-25-001/index.html

Source/driver/Castro.cpp
```diff
@@ -1438,7 +1439,7 @@ Castro::init (AmrLevel &old)
 {
     BL_PROFILE("Castro::init(old)");
 
-    Castro* oldlev = (Castro*) &old;
+    Castro* oldlev = dynamic_cast<Castro*>(&old);
```
yut23, line 1442: Do all the compilers we care about support this? cppreference says some implementations use RTTI

zingale: not sure.  I can remove that.

zingale: reverted

Source/driver/Castro.cpp
```diff
@@ -2710,7 +2713,7 @@ Castro::reflux(int crse_level, int fine_level)
 
             // Start with a MultiFab version of the flux register.
 
-            for (OrientationIter fi; fi; ++fi) {
+            for (OrientationIter fi; fi != nullptr; ++fi) {
```
yut23, line 2716: I think `fi.isValid()` is better here

Source/driver/Castro_io.cpp
```diff
@@ -994,10 +994,10 @@ Castro::plotFileOutput(const std::string& dir,
...
         //
-        for (int i =0; i < static_cast<int>(plot_var_map.size()); i++)
+        for (auto& p : plot_var_map)
         {
-            int typ = plot_var_map[i].first;
-            int comp = plot_var_map[i].second;
+            int typ = p.first;
+            int comp = p.second;
             os << desc_lst[typ].name(comp) << '\n';
         }
```
yut23, line 1002: We can use a structured binding declaration here:
```
        for (const auto& [typ, comp] : plot_var_map) {
            os << desc_lst[typ].name(comp) << '\n';
        }
```

yut23 (changes requested):
Looks good aside from my comment on Castro.cpp

yut23 (approved):

PR #2389, opened 2023-03-06
Add nse status of burn_t to statedata

zhichen3:
## PR summary
Add the nse status indication in `burn_t` to StateData for diagnostics.

## PR checklist


zhichen3:
not sure if this is the way to do it since here we're working with a boolean? (does that matter at all?)

zingale:
I think that we should add this to Reactions_Type

PR #2390, opened 2023-03-07
add nse status of burn_t to reactions_type

zhichen3:
## PR summary
add nse status of burn_t to reactions_type

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-03-16-001/index.html

Source/driver/Castro_setup.cpp
```diff
@@ -652,7 +655,16 @@ Castro::variableSetUp ()
       }
 #endif
   }
-
+#ifdef NSE
+  set_scalar_bc(bc,phys_bc);
+  replace_inflow_bc(bc);
+  if (store_omegadot == 1) {
+    desc_lst.setComponent(Reactions_Type, NumSpec+NumAux, "nse", bc, genericBndryFunc);
```
zingale, line 662: shouldn't this be NumSpec+NumAux + 1?

zingale (approved):

zingale:
for the massive_star problem, there are now 2 nse states in the plotfile, `nse` and `in_nse`, and they are different.  `nse` doesn't capture the NSE state correctly.  We should fix this first before the merge.

zhichen3:
Maybe its because you have the `relax_factor`? The condition is different since during NSE bail-out we turn on the relax-factor. So in theory, `nse` should give a larger coverage than `in_nse`.

I'll run some tests to see what's going on.

zhichen3:
> for the massive_star problem, there are now 2 nse states in the plotfile, `nse` and `in_nse`, and they are different. `nse` doesn't capture the NSE state correctly. We should fix this first before the merge.

is there a way for me to reproduce this easily?

zingale:
I'll take a look again

zingale:
it seems to only affect simplified-SDC

in `Castro/Exec/science/massive_star` do:

```
make USE_SIMPLIFIED_SDC=TRUE -j 8
./Castro2d.gnu.MPI.SMPLSDC.ex inputs_2d.nse castro.time_integration_method=3 amr.max_level=0 max_step=1
fextrema.gnu.ex plt00001 | grep nse
```

zingale:
oh, the issue is that we are not storing the NSE state in `Castro_react.cpp` for the SDC version.  There are 2 versions of that code in there, but only one is active due to the `#ifdef`

zingale:
look around line 590.  We are not storing `burn_state.nse` anywhere

zingale:
okay.  This works now

zingale (approved):

PR #2409, opened 2023-03-31
remove derived in_nse plot var and rename new one to in_nse

zingale:
## PR motivation

We switched from deriving whether we are in NSE for the plotfiles and instead
using the NSE state stored in the Reactions_Type StateData.  This change eliminates
the duplicated derived `in_nse` plot variable and renames the new Reactions_Type
plot variable to `in_nse`.  This makes our scripts compatible with the way of computing
it without needing to change them.

## PR checklist


zhichen3:
is this ready to review?

zingale:
yes

zhichen3:
should we also remove `ca_dernse` in `Derive.H` ?

zingale:
yes, good catch.  I'll do that

zingale:
done

Issue #2413, opened 2023-04-10
MGFLD lab frame flux


Saligiare-pwd:
Dear all,
I have been trying to Output lab frame flux when MGFLD is in used, but the results are all zero in Frlab_nx.
I am running modified shock breakout test problem, other test problem with MGFLD also can not output lab frame flux information (comoving flux works well).
Is there anything that I might missed?

Thank you!

Saligiare-pwd:
Thank you!
I also receive same error if I plot both 

radiation.plot_com_flux=1

radiation.plot_lab_flux=1

My initial guess is that in MGFLDRadsolver.cpp:

  if (plot_com_flux) {
      flxcc = plotvar[level].get();
      icomp_flux = icomp_com_Fr;
  } else if (plot_lab_Er || plot_lab_flux) {
      flxsave.reset(new MultiFab(grids, dmap, nGroups*AMREX_SPACEDIM, 0));
      flxcc = flxsave.get();
      icomp_flux = 0;
  } 

Which might skip plotting lab_flux if com_flux is needed?

Saligiare-pwd:
Dear all,
The file corrects the 2D problem!
But for 3D, I get the following error when outputting lab frame flux.

amrex::Abort::7:: (32,32,32,16) is out of bound (32:47,32:47,32:47,0:11) !!! 
SIGABRT
amrex::Abort::0:: (0,0,0,16) is out of bound (0:15,0:15,0:15,0:11) !!! 
SIGABRT
amrex::amrex::Abort::1:: (32,0,0,16) is out of bound (32:47,0:15,0:15,0:11) !!! 
Abort::5:: (32,0,32,16) is out of bound (32:47,0:15,32:47,0:11) !!! 
amrex::SIGABRT
Abort::SIGABRT
6:: (0,32,32,16) is out of bound (0:15,32:47,32:47,0:11) !!! 
SIGABRT
amrex::Abort::2:: (0,32,0,16) is out of bound (0:15,32:47,0:15,0:11) !!! 
amrex::amrex::Abort::Abort::4:: (0,0,32,16) is out of bound (0:15,0:15,32:47,0:11) !!! 
3:: (32,32,0,16) is out of bound (32:47,32:47,0:15,0:11) !!!

zingale:
can you tell me how to reproduce this?

zingale:
[...]

PR #1700, opened 2021-05-11
Convert save_flux_in_plotvar to C++

maxpkatz:
## PR summary

The versions to save the lab or comoving flux were the same up to a minus sign, so I merged them into a single routine.

## PR checklist

- [x] test suite needs to be run on this PR
- [ ] this PR will change answers in the test suite to more than roundoff level
- [ ] all newly-added functions have docstrings as per the coding conventions
- [ ] the `CHANGES` file has been updated, if appropriate
- [ ] if appropriate, this change is described in the docs

PR #2415, opened 2023-04-12
Fix indexing in plotfile radiation lab/COM flux

maxpkatz:
## PR summary

This looks like it was an error in the port from Fortran in #1700. This should address the issue in #2413.

## PR checklist


zingale:
only 1 radiation test fails:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-04-12-002/rad-breakout.html

curiously this stores both the com and lab flux -- it is not clear to me that storing both should be allowed.

Source/radiation/RadPlotvar.cpp
```diff
@@ -193,16 +200,16 @@ void Radiation::save_flux_in_plotvar(int level, const MultiFab& Snew,
...
 #if AMREX_SPACEDIM >= 2
             int ifiy = iflx + NGROUPS;
-            int ifoy = nflx + NGROUPS;
+            int ifoy = icomp_flux + NGROUPS;
 #endif
 
 #if AMREX_SPACEDIM == 3
             int ifiz = iflx + NGROUPS * NGROUPS;
-            int ifoz = nflx + NGROUPS * NGROUPS;
+            int ifoz = icomp_flux + NGROUPS * NGROUPS;
```
zingale, line 212: I think that this code was wrong originally.  We dimension the flux to be of side NGROUPS * AMREX_SPACEDIM, so the offset for 3-d should be 

```
int iflz = iflx + 2 * NGROUPS;
int ifoz = icomp_flux + 2 * NGROUPS;
```

as it stands now we have a bounds error

zingale: the dimensioning is done in `MGFLDRadSolver.cpp` line 186

maxpkatz: ok, fixed

zingale (approved):

PR #2435, opened 2023-05-02
[WIP] loosen nse constraint only after max subcycle is reached

zhichen3:
## PR checklist


zingale:
do we still want to consider this?

zhichen3:
I think its a reasonable thing to do if we were to use these loosen-constraint parameters. But right now I tend to not use them because I think it would make the constraint "too loose".

PR #2466, opened 2023-05-26
[WIP] fix usage of mom_flux_has_p

zingale:
this is always false in 1-d, even for Cartesian, but it does not seem that we were storing the pressure flux in that case

## PR checklist


zingale:
going to close this and take the opposite approach -- for Cartesian, we will always include p in the momentum.  Not sure why we considered 1-d differently.

PR #2473, opened 2023-05-30
add geometric source terms to the PPM tracing

zingale:
## PR summary

Add the geometric source terms to the primitive variable sources so they can be
included in the tracing.  

We can't simply add them in src_to_prim, since the sources depend on the
direction of the reconstruction.  We also can't just add them in the trace_ppm
to srcQ immediately at the start and then do the reconstruction since we need
the geometric source terms in the ghost cells for the reconstruction.  Therefore
we add these geometric sources to the 5-point stencil that will be traced just
before the tracing.  This requires some redundant calculation, but it is not too
bad.

## PR checklist


zingale:
test suite diffs are about as expected:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2024-07-26-001/index.html

zingale:
this seems to cause problems with the massive_star setup after only 24 steps.

zingale:
I have the pressure source wrong

zingale:
updated tests with this PR, after fixing the pressure source:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2024-07-27-003/index.html

zingale:
Here's a comparison of `flame_wave` run for > 90,000 steps with this PR and development:

development:

flame_wave_1000Hz_25cm_smallplt95305_slice_development

this PR:

flame_wave_1000Hz_25cm_smallplt96452_slice_tracing

zingale:
Here's a massive star comparison:

development:

all_massive_star_smallplt12034_w0 10_development

tracing:

all_massive_star_smallplt12034_w0 10_tracing

zhichen3:
I think there will be geometric source terms for velocity equations? Even though those depend on velocity in phi direction.

zingale:
I don't think so -- those terms are handled as source terms outside of hydro.  The terms here are only those that exist for the area factors in the divergence for the normal direction.

zhichen3:
Ah, okay. I think I get it now. We get out normal conserved states from the primitive states and we're simply adding extra terms that came out after expanding the divergence in the normal direction, otherwise those "corrections" go missing after we convert back to the "hat" conserved states. And all the external sources including geometric sources that appear in the conserved variable equations are still added separately after we get the normal conserved states. Is that correct?

zingale:
yes, that's correct

Source/hydro/reconstruction.H
```diff
@@ -78,5 +82,69 @@ load_passive_stencil(Array4<Real const> const& U_arr, Array4<Real const> const&
...
+                         amrex::Real* s) {
+
+    using namespace reconstruction;
+
+    // this takes the form: -alpha rho u / r
+    // where alpha = 1 for cylindrical and 2 for spherical
+
+    // note: this is assumed to be working only in the x-direction
+
+    s[im2] += -dloga(i-2,j,k) * q_arr(i-2,j,k,QRHO) * q_arr(i-2,j,k,QU);
```
zhichen3, line 99: are we not multiplying the source by dt/2 ?

zingale: This is just adding the geometric sources to the other hydrodynamic sources.  Then we reconstruct and integrate under the parabola, and then the dt2 is applied when we add it to the jumps that are accumulated to the interfaces

zhichen3 (approved):

PR #2482, opened 2023-06-04
Remove spherical_star and associated functionality

maxpkatz:
## PR summary

This is not tested anywhere, and I think in its current form in C++ it could not have been useful to anyone, since the radial data that was filled in was local in scope to the `make_radial_data` function.

## PR checklist


zingale:
maybe I want this for massive_star?

zingale:
I'm okay with removing it for now, since it is not tested.  At some point in the future, I might reimplement it, just for filling the hydro ghost cells.  But I would do it by creating a second model file in our model_parser functionality.

maxpkatz:
Yeah, I have no objection to re-implementing this at some point, but right now it's getting in the way of some re-organization I want to do in Castro_advance_ctu(), so I'd rather deal with that later when we really need it.

maxpkatz:
Also, probably for this scenario you would be using MonopoleGrav, and that already has infrastructure for summing up radial density data, so that can probably be re-used too.

zingale:
correct.  I will only want this for monopole

zingale (approved):

zingale:
quick tests pass:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-06-04-001/index.html

PR #2484, opened 2023-06-05
Move CTU burns to new functions

maxpkatz:
## PR checklist

### DIFF
--- a/Source/reactions/Castro_react.cpp
+++ b/Source/reactions/Castro_react.cpp
@@ -11,6 +11,111 @@ using namespace amrex;
 
 #ifndef TRUE_SDC
 
+advance_status
+Castro::do_old_reactions (Real time, Real dt)
+{
+    bool burn_success = true;
+
+    MultiFab& R_old = get_old_data(Reactions_Type);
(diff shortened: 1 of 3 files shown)


zingale:
compilation fails: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-06-05-002/index.html

Source/reactions/Castro_react.cpp
```diff
@@ -11,6 +11,114 @@ using namespace amrex;
...
+advance_status
+Castro::do_old_reactions (Real time, Real dt)
+{
+    amrex::ignore_unused(time);
+    amrex::ignore_unused(dt);
+
+    bool burn_success = true;
+
+#ifndef SIMPLIFIED_SDC
+    MultiFab& R_old = get_old_data(Reactions_Type);
```
zingale, line 23: isn't it the case that `do_old_reactions` should never be called for simplified-SDC?

maxpkatz: The contents of the function are basically ifdef'ed out for that case. I just preferred to have the ifdef inside the function rather than in Castro_advance

zingale (approved):

PR #2492, opened 2023-06-09
Remove an unnecessary clean_state

maxpkatz:
## PR summary

This call should be unnecessary since the hydro and new-time sources both call clean_state after being applied to the state.

## PR checklist


zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-06-15-001/index.html
this only affects radiation, since that has a ghost cell on State_Type StateData

PR #2523, opened 2023-07-09
Only do reactions on leaf zones when not subcycling

maxpkatz:
## PR summary

As a follow-up to #2505, for non-subcycling advances we can skip burning if we're not on a "leaf" zone (i.e. a zone not covered by any fine levels). These zones will be updated by an avgDown() in post_timestep().

## PR checklist


maxpkatz:
Fixed for now by only masking on the valid zones. In a future commit we could possibly expand the functionality of makeFineMask to support ghost zones, or to be clever about it and decide whether to burn in the ghost zone based on whether the nearest valid zone is masked out.

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-07-10-002/index.html

zingale (approved):

PR #2553, opened 2023-09-16
fix the subchandra second pert

zingale:
it was overriding the original pert, not adding to it

## PR checklist


zingale:
okay, this works correctly now

maxpkatz (approved):

PR #2570, opened 2023-09-21
sync up react_state's return type

zingale:
## PR summary

We had the return type as `bool`, but were returning an `int`.  This shouldn't matter,
but this makes it consistent.

## PR checklist


Source/reactions/Castro_react.cpp
```diff
@@ -20,7 +20,7 @@ Castro::do_old_reactions (Real time, Real dt)
     advance_status status {};
 
 #ifndef SIMPLIFIED_SDC
-    bool burn_success = true;
+    int burn_success{1};
```
maxpkatz, line 23: We should also then change the `if (!burn_success)` checks

zingale: fixed

Source/reactions/Castro_react.cpp
```diff
@@ -843,11 +845,7 @@ Castro::react_state(Real time, Real dt)
 
     }
 
-    if (burn_success) {
-        return true;
-    } else {
-        return false;
-    }
+    return burn_success == 1;
```
maxpkatz, line 848: Now that the return type is int, can we just return burn_success?

zingale: yes

zingale: updated

maxpkatz (approved):

PR #2560, opened 2023-09-18
remove density from the true SDC reaction system

zingale:
## PR summary

This makes the linear algebra basically the same as with simplified-SDC.

## PR checklist

- [ ] test suite needs to be run on this PR
- [ ] this PR will change answers in the test suite to more than roundoff level
- [ ] all newly-added functions have docstrings as per the coding conventions
- [ ] the `CHANGES` file has been updated, if appropriate
- [ ] if appropriate, this change is described in the docs

PR #2571, opened 2023-09-22
remove the true-SDC tolerance runtime parameters

zingale:
## PR summary

This gets rid of ``sdc_solver_tol_spec``, ``sdc_solver_tol_ener``, and ``sdc_solver_atol``,
as well as the ability to relax the tolerances.  We instead now rely on the same
tolerances as provided by Microphysics/integration.

## PR checklist


zingale:
depends on #2560

zingale:
This is also ready for review now

maxpkatz (approved):

PR #2579, opened 2023-09-24
Convert sphe to C++

maxpkatz:
## PR checklist

### DIFF
--- a/Source/radiation/RadSolve.cpp
+++ b/Source/radiation/RadSolve.cpp
@@ -888,7 +888,7 @@ void RadSolve::levelDterm(int level, MultiFab& Dterm, MultiFab& Er, int igroup)
                   s[j] = d2 * (std::cos(s[j] - h2) - std::cos(s[j] + h2));
               }
 #endif
-              
+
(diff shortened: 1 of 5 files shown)


zingale:
radiation tests pass:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-09-24/index.html

Source/radiation/RadSolve.cpp
```diff
@@ -1160,8 +1169,17 @@ void RadSolve::getEdgeMetric(int idim, const Geometry& geom, const Box& edgebox,
         geom.GetEdgeLoc(s, reg, I);
       }
       const Real *dx = geom.CellSize();
-      sphe(r.dataPtr(), s.dataPtr(), idim,
-           ARLIM(edgebox.loVect()), ARLIM(edgebox.hiVect()), dx);
+
```
zingale, line 1172: aren't we missing the 1-d case here, r = r**2?

zingale: and the n = 0 case for multid?

maxpkatz: fixed, I think

zingale (approved):

PR #2580, opened 2023-09-24
some fixes to the true-SDC Newton solver

zingale:
missed these in the original density change

## PR checklist


zingale:
this fixes the convergence of the Newton solver, as tested with `reacting_convergence`

maxpkatz (approved):

PR #2581, opened 2023-09-24
Convert compute_lamborder to C++

maxpkatz:
## PR checklist


zingale:
radiation tests pass: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-09-24-002/index.html

Source/radiation/MGFLD.cpp
```diff
@@ -1001,16 +1002,1238 @@ void Radiation::compute_limiter(int level, const BoxArray& grids,
...
+
+                if (Er(i,j,k,g) == -1.e0_rt) {
+                    return;
+                }
+
+                Real r1 = 0.0;
+                Real r2 = 0.0;
+                Real r3 = 0.0;
+
+                if (Er(i-1,j,k,g) == -1.e0_rt) {
```
zingale, line 1038: are we sure this is correct?  The original code is just looking at the boundary of the box, and the expectation is that the boundary is filled with -1.  But now you are looking at the entire box, and checking if things are -1.  I presume that Er should always be >= 0, so this still works?

maybe we should add a comment explaining that -1 marks the boundary?

maxpkatz: The 1D and 2D/3D code approached this differently and I had to decide which pattern to go with. I arbitrarily chose the 3D version just because that's the one I started the port with. In either case you do have to do an operation for every zone, but the 1D code handles the interior in one step and the boundary in a second step, while this code handles both in one loop. If you extended the 1D approach for 3D then you'd have six separate boundary loops which just seems unnecessary, though maybe marginally more computationally efficient.

Source/radiation/MGFLD.cpp
```diff
@@ -1001,16 +1002,1244 @@ void Radiation::compute_limiter(int level, const BoxArray& grids,
...
+#if AMREX_SPACEDIM >= 2
+                for (int k = lam_klo; k <= lam_khi; ++k) {
+                    if (Er(reg_ilo,reg_jlo,k,g) == -1.e0_rt) {
+                        continue;
+                    }
+
+                    for (int j = reg_jlo; j <= reg_jhi; ++j) {
+                        for (int i = reg_ilo; i <= reg_ihi; ++j) {
+                            lam(i,j,k,g) = ff1(0) * lamfil(i,j,k) + ff1(1) * (lamfil(i,j-1,k) + lamfil(i,j+1,k));
+                        }
```
zingale, line 1141: we seem to be missing the lines like:
```
lam(i,j,g) = min(1.e0_rt/3.e0_rt, max(1.e-25_rt, lam(i,j,g))) 
````
that appear in `CastroRad_2d.f90` in this section

zingale: also, shouldn't we be fulling `lamfil` here, since we do a copy down below from lamfil to lam?

maxpkatz: Yes I agree that something was off here but this was another case where there was an inconsistency between 1D and 2D and 3D, so I used the 3D code as the backbone and picked an arbitrary convention where we applied that limiter to lamfil but not to lam (I know, there's no real justification for this). I am game for fixing this but since apparently no one has been using this for 2D or 3D anyway, I'm not sure what actually is. correct.

maxpkatz:
The latest version of this PR fixes the code to make sense in all dimensions. It does this by using temporary filter arrays for the x and y dimensions in 2D and 3D respectively. It also significantly condenses the code by writing a generalized filter function (and accompanying function to load the "stencil" for both lambda and Er). This should reduce the likelihood of mistakes in the code. The boundary step at the end is also simplified into a simple loop over the entire box that copies zones from the edges and corners into the corresponding locations outside the valid region.

zingale:
all tests pass:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-10-07-003/index.html

zingale (approved):

PR #2594, opened 2023-10-02
fix the coefficients for 2nd order Radau SDC

zingale:
## PR checklist


zingale:
here's a comparison of all the integration methods on `bubble_convergence` with the colorscale set to make Max happy.
This uses the Radau 2nd order fix in this PR (without the fix, the simulation crashes)

sdc_compare

maxpkatz (approved):

PR #2597, opened 2023-10-03
address some clang-tidy issues for SDC

zingale:
## PR checklist


Source/driver/Castro_advance_sdc.cpp
```diff
@@ -95,7 +95,6 @@ Castro::do_advance_sdc (Real time,
 #endif
 
       if (apply_sources()) {
-#ifndef AMREX_USE_GPU
```
maxpkatz, line 98: Do you know why this was ifdef'd out?

zingale: well, it doesn't work with GPUs, so we just didn't want to compile it, but this entire block is already wrapped in an `#ifndef AMREX_USE_GPU`, so clang-tidy said this was a dupe.

yut23 (approved):

Issue #669, opened 2019-09-12
make finest level dx() known at probinit time


zingale:
For generating initial models at runtime, it would be useful to know already when we call `probinit` what the finest level `dx(:)` is.  One way we could do this is to just query the parmparse for the domain size, nx, and maximum level, and manually compute this and store it in a `finest_level_module` that can be used in `probinit`.  

This could then be used, for example, in `bubble_convergence`, instead of having to have a separate `probin` for each resolution.

zingale:
I think most of what was needed for this is now done.

PR #2599, opened 2023-10-03
add a helper function that returns the fine grid dx

zingale:
this is for the case when we need dx before the level hierarchy is setup, like in problem_initialize.H

## PR checklist


zingale:
closes #669

maxpkatz:
I think I am confused about whether we still need this. At this point, problem_initialize() is called in Castro::initMFs(), which is done in the Castro constructor after the Amr object has been created, so the geometry object on all levels should already be populated.

zingale:
Oh.  Then how do I get the fine level dx?

WeiqunZhang:
If you make `problem_initialize` a member function of Castro, you can get the Geometry on the max_level with `this->parent->Geom(this->parent->maxLevel())`. You can also change the interface of the function and pass `this` to `problem_initialize(Castro const*)`, if you don't want to make it a member function. It's basically the same thing. A member function has `this` as a hidden argument.

WeiqunZhang:
Another possibility is to store a pointer to Amr after it's created in main.cpp as a global variable. Then you can access it from anywhere.

zingale:
closing in facor of #2608

Issue #2596, opened 2023-10-02
flame_wave dx_model params are out of sync


zingale:
In the various `inputs*` files for `flame_wave`, we use `dx_model` to define the resolution with which to generate the initial model (in `problem_initialize.H`).  Ideally this should be the same as the finest grid resolution.  But it seems like some are out of sync.

We should at a minimum update the different `dx_model` parameters.

But this hack is because of issue #669

zingale:
this can be fixed via #2599 by adding:

```
auto dx_fine = get_fine_grid_dx();
```

to `problem_initialize.H` before we create the model

PR #2610, opened 2023-10-07
update flame_wave and flame_tube to infer the model's dx

zingale:
we no longer need to set problem.dx_model but instead can get it from the_amr_ptr

Note: this will change some of the results, since we were not good at keeping the dx_model in inputs files in sync with the resolution.

## PR checklist


zingale:
closes #2596

yut23 (approved):

PR #2620, opened 2023-10-09
Convert correct_dterm to C++

maxpkatz:
## PR summary

This uses inline metric calculation to support GPU offloading. Note that the vector `s` was unused and was previously only in the code to enable the call to `sphe` to fill in `re` for spherical geometries, which is now handled through edge_center_metric if we're in a spherical geometry. We also do a loop over dimensions to make the code a little more understandable.

## PR checklist


zingale:
all tests pass:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-10-09-004/index.html

zingale (approved):

PR #2628, opened 2023-10-10
update flame profiles.py plot tool

zingale:
now it plots based on dt instead of number of plotfiles

## PR checklist

### DIFF
--- a/Exec/science/flame/analysis/profiles.py
+++ b/Exec/science/flame/analysis/profiles.py
@@ -134,7 +134,7 @@ if __name__ == "__main__":
 
     pfiles = glob.glob(f"{args.prefix}?????")
     pfiles += glob.glob(f"{args.prefix}??????")
-    pfiles += glob.glob(f"{args.prefix}??????")
+    pfiles += glob.glob(f"{args.prefix}???????")
 
     plot_nums = sorted([p.split(args.prefix)[1] for p in pfiles], key=int)
 
--- a/Exec/science/flame/analysis/profiles.py
+++ b/Exec/science/flame/analysis/profiles.py
@@ -18,7 +18,7 @@ yt.set_log_level(50)
 dt = 0.005
 tmax = 0.05
 
-times = np.arange(0.0, tmax, dt)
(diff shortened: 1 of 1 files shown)


Exec/science/flame/analysis/profiles.py
```diff
@@ -100,20 +121,26 @@ def doit(prefix, nums, skip, limitlabels, xmin, xmax):
...
     p.add_argument("--limitlabels", type=float, default=1.,
                    help="Show all labels (default) or reduce to ~ given value")
 
     args = p.parse_args()
 
-    plot_prefix = args.plotfiles[0].split("plt")[0] + "plt"
-    plot_nums = sorted([p.split("plt")[1] for p in args.plotfiles], key=int)
+    pfiles = glob.glob(f"{args.prefix}?????")
+    pfiles += glob.glob(f"{args.prefix}??????")
+    pfiles += glob.glob(f"{args.prefix}??????")
```
yut23, line 137:
```suggestion
    pfiles += glob.glob(f"{args.prefix}???????")
```

zingale: thanks

Exec/science/flame/analysis/profiles.py
```diff
@@ -12,6 +13,28 @@
...
+    files_to_plot = []
+    for pfile in plist:
+        ds = yt.load(pfile, hint="castro")
+        for k, t in enumerate(times):
+            if mask[k]:
+                continue
+            if ds.current_time >= t:
+                files_to_plot.append(pfile)
+                print(f"plotting {pfile} at t = {t}")
+                mask[k] = 1.0
```
yut23, line 35: We should break out of the `times` loop here, so the same file doesn't get plotted multiple times (if the plotfiles are spaced out longer than `dt`). Looks good otherwise.

zingale: yes, I think you are right.  I'll fix that.

zingale: updated

yut23 (approved):

PR #2586, opened 2023-09-25
some changes to the SDC Newton solver to make it more robust

zingale:
we now reject a solve if the mass fractions are too far out of [0, 1] also do the total energy update conservatively

## PR checklist

- [ ] test suite needs to be run on this PR
- [ ] this PR will change answers in the test suite to more than roundoff level
- [ ] all newly-added functions have docstrings as per the coding conventions
- [ ] the `CHANGES` file has been updated, if appropriate
- [ ] if appropriate, this change is described in the docs

PR #2629, opened 2023-10-10
[WIP] add renomalization after the update

zingale:
## PR summary

in some sense, this mirrors what we would do with Strang or
simplified-SDC, since we always renormalize after leaving the
burner.

## PR checklist


zingale:
this requires #2586

PR #2635, opened 2023-10-14
implement inflow BCs for the flame

zingale:
## PR summary

This will allow us to inflow fuel in through the right boundary to hold the flame roughly stationary.  Better
still, it will hopefully remove the influence of an outflow BC on the flame propagation.

## PR checklist

### DIFF
--- a/Exec/science/flame/problem_bc_fill.H
+++ b/Exec/science/flame/problem_bc_fill.H
@@ -1,0 +1,45 @@
+#ifndef problem_bc_fill_H
+#define problem_bc_fill_H
+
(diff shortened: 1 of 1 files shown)


zingale:
I don't think this is correct yet.  We need to fill the ghost cells such that when we interpolate to the boundary we get the inflow velocity on the interface.  So this means I need to use the proper stencil there.  I can either do this in problem_initialize_state_data.H or by modifying the stencils in the interpolation.

zingale:
I am going to do a proper inflow in a separate PR

PR #2641, opened 2023-10-19
Standardize on USE_MPI = TRUE in all makefiles

maxpkatz:
## PR checklist


zingale (approved):

zingale:
we might want to setup an action to enforce some basic standards in the GNUmakefile.  I can do that if we can agree on what we want

like

* COMP = gnu
* USE_CUDA = FALSE
* USE_MPI = TRUE
* DEBUG = FALSE

etc.

PR #2642, opened 2023-10-19
add an action to enforce consistency in GNUmakefiles

zingale:
## PR checklist


zingale:
I think this works now.  Do we also want to enforce a default for `USE_OMP`?  I think we should default it to `FALSE`

maxpkatz (approved):

Issue #2663, opened 2023-11-30
model_parser interpolation is using the wrong index


zingale:
In `model_parser.H`, the `locate()` function returns `loc = ihi`, which would be the index of the zone to the right of the input `r`.  This means that our linear interpolation should use `loc-1` and `loc`, but we are using `loc` and `loc+1`.  Since we are linear and are not clamping we never noticed the issue.

For the cubic interpolation in #2659, this one-off indexing causes an issue.

zingale:
it is actually okay, but just confusing.  This is cleaned up in #2664.

PR #2664, opened 2023-11-30
clean up linear interpolation

zingale:
The locate() function was returning the index of the zone that is higher than the input coordinate r.  The logic in interpolate accounted for this, but things were more complex than they need be.  This simplifies locate and the linear interpolation.

It also adds a unit test to help understand what's happening.

## PR checklist


zingale:
closes #2663

Util/model_parser/model_parser.H
```diff
@@ -94,18 +98,14 @@ locate(const Real r, const int model_index) {
         loc = ihi;
     }
 
-    return loc;
+    return loc-1;
```
maxpkatz, line 101: What would it mean for locate() to return -1?

zingale: should be fixed now.

zingale:
There are now a bunch of unit tests in `Castro/util/model_parser`

These test what happens when you ask for an r outside of the model range and if you get exactly the model data (to roundoff) if you ask for an r which is exactly a model point.

zingale:
okay, looking at this more.  The old interpolation did work fine, in the sense that it had a general enough scheme to check if we need to use i, i-1 or i, i+1 for the bulk of the interpolation.

In fact, the test suite is just roundoff diffs now:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-12-17/index.html

Util/model_parser/model_parser.H
```diff
@@ -102,73 +109,35 @@ AMREX_INLINE AMREX_GPU_HOST_DEVICE
...
-            slope = (model::profile(model_index).state(id, var_index) -
-                     model::profile(model_index).state(id-1, var_index)) /
-                (model::profile(model_index).r(id) - model::profile(model_index).r(id-1));
-            interp = slope * (r - model::profile(model_index).r(id)) + model::profile(model_index).state(id, var_index);
-
-        }
-
-    }
+    // safety check to make sure interp lies within the bounding points
+    Real minvar = std::min(model::profile(model_index).state(id+1, var_index),
```
maxpkatz, line 126: Do we want this for r < profile.r(0)?

zingale: indeed, probably not.  That's a good point.

zingale: updated

maxpkatz (approved):

zingale:
tests are roundoff:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-12-26-001/index.html

PR #2666, opened 2023-12-01
Add a Python parser for diag.out files and fix header formatting

yut23:
## PR summary

This PR adds some Python code that can parse the files and sanitize any duplicate entries (which often occurs when a run is restarted from an earlier checkpoint). It also fixes some mistakes in the gravity_diag.out header formatting, and cleans up some of the other header-writing code.

## Additional details

N.B.: Standard streams in C++ are right-justified by default (compared to Python, where numbers are right-justified and strings are left-justified).

* update the header printing code for gravity_diag.out to match the others (the column numbers were 26 characters long instead of 25, and number 8 was missing)
* simplify species formatting for species_diag.out (the previous code was just adding leading spaces)
* update the header field widths for amr_diag.out to match what's used for the data

## PR checklist

### DIFF
--- a/Source/driver/sum_integrated_quantities.cpp
+++ b/Source/driver/sum_integrated_quantities.cpp
@@ -359,7 +359,7 @@ Castro::sum_integrated_quantities ()
                    header << std::endl;
 
                    data_log1 << std::setw(intwidth) << "#   COLUMN 1";
-                   data_log1 << std::setw(fixwidth) << "                        2";
+                   data_log1 << std::setw(fixwidth) << 2;
 
                    for (int icol = 3; icol <= n; ++icol) {
                        data_log1 << std::setw(datwidth) << icol;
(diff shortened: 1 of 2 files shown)


zingale:
can you add a short description to the docs about this, perhaps in the visualization section, along with an index entry for the diag files.

zingale:
Oh, actually it looks like this is the place in the docs:

https://amrex-astro.github.io/Castro/docs/io.html#index-9

aisclark91:
I just wonder if the source driver modifications alter the output file (adding or removing spaces, etc)?

yut23:
> I just wonder if the source driver modifications alter the output file (adding or removing spaces, etc)?

The gravity_diag.out column numbers are the only things that get changed in the outputs.

yut23:
I made sure it works with both.

PR #2676, opened 2023-12-11
[WIP] Allocate arrays for sources in advected/species/auxilary equations

RemiLehe:
## PR motivation

When trying to add [source terms](https://amrex-astro.github.io/Castro/docs/Hydrodynamics.html#source-terms) in the species equations (e.g. using ` problem_source.H` file that modifies `src(i,j,k,UFS)`), I got an out-of-bound access error. With the changes in this PR, the error disappears.

## PR checklist

- [X] test suite needs to be run on this PR


zingale:
we removed sources to the species to save memory, since we don't need them (for our SDC approach, we manually add the needed bits outside of the source).

If we want to add this back, we should do it as an option, since it can be a lot of memory when we have 20-30 species.

RemiLehe:
Thanks for the feedback. 

How would the user turn the option on? Would that be controlled through the user's `GNUmakefile` for instance? (And how should I modify the PR accordingly.)

zingale:
let me close this and do a new PR that has the option

zingale:
closing in favor of #2678

PR #2702, opened 2024-01-04
[NEVER MERGE THIS!!!] constexpr the runtime parameters

zingale:
## PR summary

This is an experiment to see if keeping the runtime parameters as constexpr instead of managed helps with later
versions of ROCm.  Note: the values are compiled in, so they will need to be hacked to override the defaults for a particular problem.

## PR checklist


zingale:
run this together with this Microphysics PR:
https://github.com/AMReX-Astro/Microphysics/pull/1441

these hardcode the parameters for the subchandra problem, which is one of the problems that fails with later versions of ROCm

zingale:
The way this works is that we pass `--constexpr` to the python scripts that write the parameters headers.  This changes them to declare them as `constexpr` instead of (extern) managed.  The values are no hardcoded in, and there are no parmparse queries for them.

PR #2704, opened 2024-01-05
start of a simpler / better shock detection algorithm

zingale:
## PR summary

The original algorithm was based on one in Flash but it had some normalizations 
in it that did not make sense.  This is much simpler and will also take into account
only the pressure available to make a shock.

## PR checklist


zingale:
Here's the difference in the old (left) vs new (right) shock detection for Sedov

image

zingale:
I still think that the shock region is too thick -- it should only be a single zone

zingale:
Here's an updated comparison, the left is the latest version, which projects the gradient in the direction of motion.

image (2)

maxpkatz (approved):

PR #2705, opened 2024-01-06
remove Elixirs in the MOL/true-SDC code path

zingale:
## PR checklist

### DIFF
--- a/Source/hydro/Castro_mol_hydro.cpp
+++ b/Source/hydro/Castro_mol_hydro.cpp
@@ -57,28 +57,32 @@ Castro::construct_mol_hydro_source(Real time, Real dt, MultiFab& A_update)
 #endif
   {
 
-    // Declare local storage now. This should be done outside the MFIter loop,
-    // and then we will resize the Fabs in each MFIter loop iteration. Then,
-    // we apply an Elixir to ensure that their memory is saved until it is no
-    // longer needed (only relevant for the asynchronous case, usually on GPUs).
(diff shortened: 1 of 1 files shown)


zingale:
test suite: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2024-01-06/index.html

the failures are unrelated

Source/hydro/Castro_mol_hydro.cpp
```diff
@@ -90,17 +94,12 @@ Castro::construct_mol_hydro_source(Real time, Real dt, MultiFab& A_update)
         const Box& obx = amrex::grow(bx, 1);
         const Box& obx2 = amrex::grow(bx, 2);
 
-        FArrayBox &statein  = Sborder[mfi];
-        Array4<Real const> const uin_arr = statein.array();
-
-        FArrayBox &stateout = S_new[mfi];
+        Array4<Real const> const uin_arr = Sborder.array(mfi);
```
maxpkatz, line 97: Stylistically, I strongly prefer Sborder[mfi].array() to Sborder.array(mfi)

maxpkatz (approved):

Source/hydro/Castro_mol_hydro.cpp
zingale: there are > a dozen instances in that file already that do `.array(mfi)` already, so this is consistent with that.
I'd be fine with a follow-on PR that standardizes all of the code to the alternate approach.

Issue #2724, opened 2024-01-22
change how shock flag is computed for reactions


zingale:
Right now, we use the shock flag (stored in the conserved state if `USE_SHOCK_VAR=TRUE`) in two different ways:

* if we are doing the hybrid Riemann solve, it is used to determine if we call the HLL solver
* it is used to check if we are zeroing reactions if we disable burning in shocks.

At the moment, it is computed only once per timestep -- at the start of the advection (in the same place that we compute the flattening coefficients).

But it appears that we are always zeroing it out in `initialize_do_advance()`, so the shock flag will always be zero in the first half of Strang reactions.

We should switch to computing it directly in when we call the burner, so it will be updated twice per step.  For the first half Strang burn, this is easy, since `Sborder` has ghostcells.  But for the second half of the Strang burn, `S_new` does not have ghost cells, so we would need to do a fill into a temporary Fab or MultiFab (we could just overwrite `Sborder` at this point, I think).

zingale:
There is an additional complication though -- we need to remove the source terms, which we don't directly have at the new time (since we time-center the sources).  

So it might just be simpler to compute the shock flag at the start of a timestep and keep it fixed for both halves of the burn -- this is certainly more in line with the way that Strang splitting works.  Also since we are typically using small CFL numbers, it won't matter much.

zingale:
there is an additional complication -- if we are to compute this before the first burn, we need the sources, so that means we need to compute gravity before the burn, which should be fine, since the burn doesn't change Phi

PR #2728, opened 2024-01-23
Compute the shock flag once -- at the start of the timestep

zingale:
## PR summary

This changes the way the shock flag is computed for disabling burning in shocks.  Now we compute it
at the top of the timestep, before the first Strang burn, using the time-level n state and sources. 
There is some redundancy here (we need to compute the old sources before the burn now),
and there is a FillPatch needed.  But this should be minimal and only in effect if we compile with SHOCK_VAR.

Note: this does not change how we compute the shock flag for hydro at the moment
when doing the hybrid Riemann solve (although we can easily change that in a future PR).

## PR checklist


zingale:
depends on #2726 
fixes #2724

zingale:
this is ready for review now

yut23 (approved):

PR #2734, opened 2024-01-26
use std:: for some math functions

zingale:
## PR checklist

### DIFF
--- a/Diagnostics/DustCollapse/main.cpp
+++ b/Diagnostics/DustCollapse/main.cpp
@@ -105,15 +105,15 @@ int main(int argc, char* argv[])
         auto x_maxdist = fmax(fabs(problo[0]), fabs(probhi[0]));
         auto y_maxdist = fmax(fabs(problo[1] - yctr), fabs(probhi[1] - yctr));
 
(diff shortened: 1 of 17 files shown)


maxpkatz (approved):

Diagnostics/DustCollapse/main.cpp
```diff
@@ -105,15 +105,15 @@ int main(int argc, char* argv[])
         auto x_maxdist = fmax(fabs(problo[0]), fabs(probhi[0]));
         auto y_maxdist = fmax(fabs(problo[1] - yctr), fabs(probhi[1] - yctr));
```
yut23, line 106: Should these be updated too?

zingale: yeah, probably.  I'll do a pass on those and add it to this PR

yut23: I also see a bunch of places we can use `powi`, but I figure we should do that in another PR

zingale: yes, that should be separate

PR #2737, opened 2024-01-28
add some checks on dynamic_casts in gravity

zingale:
## PR summary

this addresses some coverity errors

## PR checklist

### DIFF
--- a/Source/gravity/Gravity.cpp
+++ b/Source/gravity/Gravity.cpp
@@ -1843,8 +1843,11 @@ Gravity::fill_multipole_BCs(int crse_level, int fine_level, const Vector<MultiFa
         MultiFab::Copy(source, *Rhs[lev - crse_level], 0, 0, 1, 0);
 
         if (lev < fine_level) {
-            const MultiFab& mask = dynamic_cast<Castro*>(&(parent->getLevel(lev+1)))->build_fine_mask();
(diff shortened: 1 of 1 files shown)


Source/gravity/Gravity.cpp
```diff
@@ -1843,8 +1843,11 @@ Gravity::fill_multipole_BCs(int crse_level, int fine_level, const Vector<MultiFa
         MultiFab::Copy(source, *Rhs[lev - crse_level], 0, 0, 1, 0);
 
         if (lev < fine_level) {
-            const MultiFab& mask = dynamic_cast<Castro*>(&(parent->getLevel(lev+1)))->build_fine_mask();
-            MultiFab::Multiply(source, mask, 0, 0, 1, 0);
+	    auto *castro_level = dynamic_cast<Castro*>(&(parent->getLevel(lev+1)));
+	    if (castro_level != nullptr) {
```
maxpkatz, line 1847: I feel like there should be an `else` here which aborts. We shouldn't be getting to the codepath where the pointer is not null.

zingale: yeah, I'll add that

zingale: updated

maxpkatz (approved):

PR #2742, opened 2024-02-05
Update documentation regarding species sources and normalization

KiranEiden:
## PR summary

This PR updates the Hydrodynamics section documentation to highlight that sources for passive variables in the conserved state are off by default. It also removes the reference to `castro.normalize_species`, which does not appear to be a valid runtime parameter any longer. I am happy to introduce additional documentation regarding the species normalization if that would be helpful -- my understanding is that it is on by default now.

## PR checklist


Docs/source/Hydrodynamics.rst
```diff
@@ -297,7 +297,11 @@ Source Terms
 We now compute explicit source terms for each variable in :math:`\Qb` and
 :math:`\Ub`. The primitive variable source terms will be used to construct
 time-centered fluxes. The conserved variable source will be used to
-advance the solution. We neglect reaction source terms since they are
+advance the solution. To reduce memory usage, we do not include source terms
```
zingale, line 300: I would suggest putting this new text in a `.. note::` so it stands out

also add an index entry, 

`.. index:: USE_SPECIES_SOURCES`

zingale (approved):

zingale:
Oh, I missed that there was only a single `

image

can you fix the typesetting of `USE_SPECIES_SOURCES` in a new PR?

KiranEiden:
Sure, sorry about that. I’ll do that later tonight.

Issue #2569, opened 2023-09-21
Castro reactions fail with ROCm > 5.3.0


zingale:
Using this issue to track problems with more recent versions of ROCm

zingale:
To build with ROCm 5.7.0 on Frontier:

```
module load cpe/23.09
module load rocm/5.7.0
module load PrgEnv-gnu craype-accel-amd-gfx90a cray-mpich
```

zingale:
Running the 2D subchandra problem on 4 nodes / 32 GPUs:

* We run fine if compiled with `DEBUG=TRUE` (although the compilng / linking takes a Hubble time)
* We crash with problems in the reaction integration right away (step 1) if compiled normally.

zingale:
More precisely, the burning fails and the step is thrown out and we need to subcycle.

zingale:
The ``reacting_convergence`` problem works fine, it seems.  This could be because of Strang vs. SDC, or the type of network.

zingale:
okay, with subchandra and Strang, and ROCm 5.7.0 I now get:

```
... Entering burner on level 0 and doing half-timestep of burning.

:0:rocdevice.cpp            :2692: 142346779052 us: [pid:90391 tid:0x7fffd6530700] Callback: Queue 0x7ffe91600000 aborting with 
error : HSA_STATUS_ERROR_MEMORY_APERTURE_VIOLATION: The agent attempted to access memory beyond the largest legal address. code:
 0x29
SIGABRT
Memory access fault by GPU node-8 (Agent handle: 0x2975b60) on address 0x800033773000. Reason: Unknown.
SIGABRT
See Backtrace.24 file for details
See Backtrace.7 file for details
```

BenWibking:
Is this issue we're seeing when running problems with Microphysics in Quokka possibly related: https://github.com/quokka-astro/quokka/issues/394?

It only happens on AMDGPU, and only when we use reaction networks.

zingale:
We run without problem with ROCm 5.3.0 and reactions.  So I am not sure if it is the same.

BenWibking:
> We run without problem with ROCm 5.3.0 and reactions. So I am not sure if it is the same.

Hmm, that makes it probably unrelated. Have you happened to run with ROCm 5.2.3? We only have access to this version on our machine.

zingale:
possibly related:
https://github.com/AMReX-Codes/amrex/issues/3623

zingale:
[...]

PR #2749, opened 2024-02-09
ROCm workaround: Use ParallelFor instead of Reduce

WeiqunZhang:
Assuming the failure is not often, we can use ParallelFor with atomicAdd to
obtain the number of failures. With this change, the ROCm memory issue seems
to be gone.

### DIFF
--- a/Source/reactions/Castro_react.cpp
+++ b/Source/reactions/Castro_react.cpp
@@ -193,7 +193,7 @@ Castro::react_state(MultiFab& s, MultiFab& r, Real time, Real dt, const int stra
     int num_failed = 0;
 
 #ifdef _OPENMP
-#pragma omp parallel reduce(+:num_failed);
+#pragma omp parallel reduction(+:num_failed);
 #endif
     for (MFIter mfi(s, TilingIfNotGPU()); mfi.isValid(); ++mfi)
     {
--- a/Source/reactions/Castro_react.cpp
+++ b/Source/reactions/Castro_react.cpp
@@ -193,7 +193,7 @@ Castro::react_state(MultiFab& s, MultiFab& r, Real time, Real dt, const int stra
     int num_failed = 0;
 
 #ifdef _OPENMP
-#pragma omp parallel reduction(+:num_failed);
+#pragma omp parallel reduction(+:num_failed)
 #endif
     for (MFIter mfi(s, TilingIfNotGPU()); mfi.isValid(); ++mfi)
     {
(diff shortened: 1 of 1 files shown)


zingale:
this fixes #2569

BenWibking:
@WeiqunZhang would you recommend that we use atomics instead of `amrex::ParReduce` in Quokka as well:
https://github.com/quokka-astro/quokka/blob/2d863d655ec1e1f684f55ea745021b9c97a03c60/src/hydro_system.hpp#L204

WeiqunZhang:
No, not for the functions in your links. Those kernels are small. Only if the kernel is very big and you see an issue, you might try to use atomics. Note that using atomics is very slow if it needs to be done a lot of times.

BenWibking:
> No, not for the functions in your links. Those kernels are small. Only if the kernel is very big and you see an issue, you might try to use atomics. Note that using atomics is very slow if it needs to be done a lot of times.

Is there a way to identify kernels that might be affected by the same issue?

The kernel that calls our reaction network is here: https://github.com/quokka-astro/quokka/blob/2d863d655ec1e1f684f55ea745021b9c97a03c60/src/Chemistry.hpp#L43

The reaction network itself (6k LOC) is here:
https://github.com/AMReX-Astro/Microphysics/blob/main/networks/primordial_chem/actual_rhs.H#L20

WeiqunZhang:
You can try to insert `Gpu::streamSynchronize` like in this PR. Without it, I saw
```
:0:rocdevice.cpp            :2724: 441955224434 us: [pid:105317 tid:0x7fffdd5ff700] Callback: Queue 0x7ffec2f00000 Aborting with error : HSA_STATUS_ERROR_OUT_OF_RESOURCES: T
he runtime failed to allocate the necessary resources. This error may also occur when the core runtime library needs to spawn threads or create internal OS-specific events. 
Code: 0x1008 Available Free mem : 13506 MB
```
It might help in your case.

BenWibking:
> You can try to insert `Gpu::streamSynchronize` like in this PR. Without it, I saw
> 
> ```
> :0:rocdevice.cpp            :2724: 441955224434 us: [pid:105317 tid:0x7fffdd5ff700] Callback: Queue 0x7ffec2f00000 Aborting with error : HSA_STATUS_ERROR_OUT_OF_RESOURCES: T
> he runtime failed to allocate the necessary resources. This error may also occur when the core runtime library needs to spawn threads or create internal OS-specific events. 
> Code: 0x1008 Available Free mem : 13506 MB
> ```
> 
> It might help in your case.

We can try that. You added it immediately after the burn kernel, right?

WeiqunZhang:
Yes.

WeiqunZhang:
It might also help to turn on tiling `MFIter mfi(mf, IntVect(32,32,2048))` for the failed MFIter loop. If you do that, you need to use `tilebox()`, not `validbox()`.

BenWibking:
> It might also help to turn on tiling `MFIter mfi(mf, IntVect(32,32,2048))` for the failed MFIter loop. If you do that, you need to use `tilebox()`, not `validbox()`.

We could try that. Does this change the kernel launch parameters (and/or, do you know why it helps)?

WeiqunZhang:
It changes the number of blocks in the kernel launch.

zingale (approved):

PR #2754, opened 2024-02-12
Turn Sphinx warnings into errors in CI

yut23:
## PR checklist


yut23:
For reference: Sphinx counts warnings through the logging framework, and doesn't distinguish between log levels, so errors and warnings are treated the same. The "Malformed table" and "line-length-limit" errors come from docutils.

zingale (approved):

PR #2793, opened 2024-03-19
get rid of asctime -- it is unsafe

zingale:
flagged by clang-tidy bugprone-unsafe-functions

## PR checklist


yut23:
It looks like we can get rid of the buffer if we use `std::put_time()`: https://en.cppreference.com/w/cpp/io/manip/put_time#Example

zingale:
updated

yut23 (approved):

PR #2797, opened 2024-03-24
define runtime params in .cpp instead of *_declares.H

zingale:
This fixes a potential ODR problem with *_declares.H, which is where we did the actual
definition of the runtime parameters.  Now we output a single runtime_params.cpp
that has all of the runtime parameters defined in source instead of headers.

This also removes padding from the struct that we can alternately use for runtime
params

## PR checklist


Source/driver/parse_castro_params.py
```diff
@@ -255,20 +266,25 @@ def write_headers(params, out_directory, struct_name):
         params_nm = [q for q in params if q.namespace == nm]
         # sort by repr since None may be present
         ifdefs = sorted({q.ifdef for q in params_nm}, key=repr)
-
         sf.write(f"struct {nm}_t {{\n")
         print("namespace = ", nm)
         for ifdef in ifdefs:
+            params_if = [q for q in params_nm if q.ifdef == ifdef]
+            types = set(q.dtype for q in params_if)
```
yut23, line 273: `set` ordering is non-deterministic for strings:
```suggestion
            types = sorted(set(q.dtype for q in params_if))
```

yut23:
We should sort `namespaces` as well, on line 139.

Source/driver/parse_castro_params.py
```diff
@@ -41,16 +41,16 @@
...
-
   -- name_queries.H  (for castro, included in Castro.cpp):
      does the parmparse query to override the default in C++
 
   -- name_job_info_tests.H
      this tests the current value against the default and outputs
      into a file
 
+  -- name_params.cpp
+     has the actual definition of the variables (without extern)
```
yut23, line 52: This is now `runtime_params.cpp`

zingale:
added the sort on namespaces

Source/driver/parse_castro_params.py
zingale: fixed

Source/driver/parse_castro_params.py
```diff
@@ -129,45 +129,19 @@ def read_param_file(infile):
 
     return params
 
-def write_headers(params, out_directory, struct_name):
+def write_headers_and_source(params, out_directory, struct_name):
 
     # output
 
     # find all the namespaces
     namespaces = {q.namespace for q in params}
```
yut23, line 137: Sorting once here is cleaner (you missed the loops on lines +226 and +293)
```suggestion
    namespaces = sorted({q.namespace for q in params})
```

zingale:
okay, I removed the redundant sorts

yut23 (approved):

PR #2805, opened 2024-03-26
use std::filesystem::current_path in job_info

zingale:
this eliminates a C routine with a buffer

## PR checklist


zingale:
alternately, we could use `amrex::CurrentPath()` ?

yut23 (approved):

PR #2812, opened 2024-04-04
make clean should remove temporary .cpp files too

zingale:
## PR checklist


zingale:
I guess this is kinda redundant though, since it appears that amrex removes all of tmp_build_dir

maybe this is a vestige from when `clean` and `realclean` did different things?

yut23:
Also, `CASTRO_AUTO_SOURCE_DIR` will only be correct if you specify all the right variables (`USE_MPI`, `DEBUG`, etc.) when you call `make clean`.

zingale:
right.  I think that this goes back to the old difference between `make clean` and `make realclean`, which no longer exists anymore

maxpkatz (approved):

PR #2824, opened 2024-04-12
more fixes for amr.subcycling_mode = None + retry

zingale:
## PR checklist


Source/driver/Castro_advance_ctu.cpp
```diff
@@ -157,11 +157,18 @@ Castro::retry_advance_ctu(Real dt, const advance_status& status)
...
+            max_level_to_advance = parent->finestLevel();
         }
-        else {
-            dt_subcycle = std::min(dt, dt_subcycle) * retry_subcycle_factor;
+
+        for (int lev = level; lev <= max_level_to_advance; ++lev) {
+            if (status.suggested_dt > 0.0_rt && status.suggested_dt < dt) {
+                getLevel(lev).dt_subcycle = status.suggested_dt;
+            } else {
+                getLevel(lev).dt_subcycle = std::min(dt, dt_subcycle) * retry_subcycle_factor;
```
yut23, line 170: This was setting `dt_subcycle` for the finer levels to be different from the value for level 0. It doesn't appear to fix the negative densities, though.
```suggestion
                getLevel(lev).dt_subcycle = std::min(dt, getLevel(lev).dt_subcycle) * retry_subcycle_factor;
```

yut23:
Is this bite taken out of the secondary an issue? inputs_2d_collision.test says it sets `problem.collision_separation = 1.25` to start with the stars already merging, but these don't look like very well-behaved initial conditions.
Screenshot from 2024-04-18 19-11-08

zingale:
if you build Sedov in 2D and run as:

```
./Castro2d.gnu.MPI.ex inputs.2d.sph_in_cylcoords castro.change_max=1.01 amr.subcycling_mode=None castro.fixed_dt=2.e-5
```
then it retries after step 2 and dies in the Riemann solver

zingale:
actually, it runs if we add `castro.riemann_solver=0`

zingale:
okay... with this PR, the Sedov retries a bit at the start and then runs fine.  Looking at the output, it looks reasonable (by eye).  I will compare to the subcycling case in a bit.

On development, it constantly retries and then dies with too many subcycles.  

So I think that these fixes are needed.

yut23:
This number (which I'm assuming is some sort of placeholder) seems to be leaking into the state variables. Changing it to 1.e40 changes all the invalid density values, for example from -5e+65 to -5e+45. https://github.com/AMReX-Astro/Castro/blob/5f44c133d85369cba7d2a12edc5f40ffaa271b3e/Source/gravity/Gravity.cpp#L396

Source/driver/Castro_advance_ctu.cpp
```diff
@@ -331,19 +346,18 @@ Castro::subcycle_advance_ctu(const Real time, const Real dt, int amr_iteration,
...
-            swap_state_time_levels(0.0);
+            for (int lev = level; lev <= max_level_to_advance; ++lev) {
+                getLevel(lev).swap_state_time_levels(0.0);
 
 #ifdef GRAVITY
-            if (do_grav) {
-                gravity->swapTimeLevels(level);
-            }
+                if (do_grav) {
+                    getLevel(lev).gravity->swapTimeLevels(level);
```
yut23, line 354: Found it.
```suggestion
                    getLevel(lev).gravity->swapTimeLevels(lev);
```

yut23 (approved):

PR #2829, opened 2024-04-20
fix the wdmerger case when the 2 stars are very close

zingale:
the initialization was overlapping and doing bad things now we simply look at which model provides the largest density at a location and use that.

## PR checklist


zingale:
Here's the wdmerger 2D collision test case now:

image

compare to the image in #2824

zingale:
tests: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2024-04-20/index.html

yut23 (commented):
The velocities need to be corrected too. I think the same condition should work.

zingale:
velocities updated, but I want to rerun the suite on this now.

yut23 (approved):

PR #2866, opened 2024-06-05
initialize aux variables for NSE_TABLE for subchandra

zhichen3:
This problem doesn't have any zones in NSE initially, so this currently doesn't check that.

## PR checklist

### DIFF
--- a/Exec/science/subchandra/problem_initialize_state_data.H
+++ b/Exec/science/subchandra/problem_initialize_state_data.H
@@ -45,6 +45,7 @@ void problem_initialize_state_data (int i, int j, int k,
     }
 
     burn_t burn_state;
+
     burn_state.rho = state(i,j,k,URHO);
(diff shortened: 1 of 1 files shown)


Exec/science/subchandra/problem_initialize_state_data.H
```diff
@@ -52,6 +52,15 @@ void problem_initialize_state_data (int i, int j, int k,
         burn_state.xn[n] = state(i,j,k,UFS+n);
     }
 
+#ifdef NSE_TABLE
```
zingale, line 55: I think it's better if we use the `AUX_THERMO` define here, which is set when `NSE_TABLE` is set.  That makes it clear that this is just about setting the aux quantities.

zhichen3: done

zingale (approved):

PR #2869, opened 2024-06-06
use adaptive stepping for exact Riemann integration

zingale:
## PR summary

There are 2 main changes here:

1. the riemann_support.H header is now split into 2, one for shocks and one for rarefactions
2. the rarefaction integration no longer uses a fixed stepsize, but instead now does an adaptive RK4 integration

This changes results to roundoff / tol error, as expected.

## PR checklist


zingale:
Here are comparisons of density for the 4 test problems from our 2015 paper:

riemann_compare_test1
riemann_compare_test2
riemann_compare_test3
riemann_compare_test4

Util/exact_riemann/riemann_rarefaction.H
```diff
@@ -0,0 +1,432 @@
...
+#define RIEMANN_RAREFACTION_H
+
+#include <utility>
+
+#include <AMReX_REAL.H>
+
+#include <eos_type.H>
+#include <eos.H>
+
+
```
aisclark91, line 11: Because we assume only rarefractions in this header file, we plan to integrate the Riemann invariants $\mathbf{L}^{(+,-)} \cdot \mathbf{dq} = 0$, and  $\mathbf{L}^{(0)} \cdot \mathbf{dq} = 0$. The variable `iwave` represents the characteristic wave, 1 for the left wave, 2 for the contact wave and 3 for the right one.

zingale: are you suggesting to add this as a comment?

Util/exact_riemann/riemann_rarefaction.H
```diff
@@ -0,0 +1,432 @@
...
+    amrex::ignore_unused(u);
+
+    // here, p is out independent variable, and tau, u are the
+    // dependent variables.  We return the derivatives of these
+    // wrt p for integration.
+
+    // T should be an initial guess coming in
+
+    // get the thermodynamics
+
```
aisclark91, line 28: From the equation of state, we hold the density $\rho = 1/\tau$, the pressure $p$, the composition $X_i$, and the temperature $T$  in `eos_rep_t`. Then we perform all the eos calculations based on $\rho(\tau)$ and $p$, override the input temperature, and compute the sound speed $C$.

Util/exact_riemann/riemann_rarefaction.H
```diff
@@ -0,0 +1,432 @@
...
+#include <eos_type.H>
+#include <eos.H>
+
+
+AMREX_INLINE
+void
+riemann_invariant_rhs(const amrex::Real p, const amrex::Real tau, const amrex::Real u,
+                      const amrex::Real* xn, const int iwave,
+                      amrex::Real& T,
+                      amrex::Real& dtaudp, amrex::Real& dudp) {
```
aisclark91, line 17: Probably calling `dtaudp` is not necessary here.

Util/exact_riemann/riemann_rarefaction.H
```diff
@@ -0,0 +1,432 @@
...
+        dudp = -1.0_rt / C;
+    } else if (iwave == 3) {
+        dudp = 1.0_rt / C;
+    }
+
+}
+
+
+AMREX_INLINE
+void
```
aisclark91, line 55: (O) Now we proceed with the $\mathbf{L}^{(0)} = 0$.

zingale:
Here are some of my notes for this:
[exact_riemann.pdf](https://github.com/user-attachments/files/15874855/exact_riemann.pdf)

aisclark91 (approved):
Everything looks great, there is only one issue:

In `riemman_star_state.H`, after `ustar_l` and `ustar_r` are computed of each iteration, in line 141 we obtain :
`ustar = 0.5_rt * (ustar_l + ustar_r);`
This expression is not right, the actual expression is:
`ustar = (Z_l * ustar_l + Z_r * ustar_r)/(Z_l + Z_r)`
In accordance with Van Leer [22] eq A3. I double checked it to be sure.

zingale:
I don't think that the expression I am using is wrong.  If the iteration were successful, then u_star,l = u_star,r, so this average is just dealing with the tolerance of the solver.

zingale:
it is actually just a different average, the Z's there are the slope at the intersection, which should be the same if we've converged the root finding.

PR #2882, opened 2024-06-24
Current inputs for sdc vs strang runs

khanakbhargava:
## PR checklist


zingale:
can we move these all into their own subdirectory under `wdmerger/tests`, like `double_det` or something  -- that will be cleaner and consistent with the other subdirectories already in there.

khanakbhargava:
Done.

zingale (approved):

Issue #2442, opened 2023-05-12
skip computing the hybrid sources if castro.hybrid_hydro=0


zingale:
We still compute the hybrid sources even if we disable hybrid hydro at runtime.

zingale:
I actually don't think this is a problem -- we still need to compute the sources for the hybrid momenta
even if we don't couple them into the solution, since we still want to use them as a diagnostic.

PR #2929, opened 2024-07-26
don't compute hybrid momentum sources unless needed

zingale:
if we compile with USE_HYBRID_MOMENTUM=TRUE but run with castro.hybrid_hydro=0, we were still computing the hybrid_sources. These are not needed.

This logic now mirrors the rotation logic.

## PR checklist

### DIFF
--- a/Source/hydro/Castro_hybrid.cpp
+++ b/Source/hydro/Castro_hybrid.cpp
@@ -14,6 +14,10 @@ Castro::construct_old_hybrid_source(MultiFab& source, MultiFab& state_old, Real 
 
     BL_PROFILE("Castro::construct_old_hybrid_source()");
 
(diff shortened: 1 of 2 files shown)


zingale:
closes #2442

zingale:
it occurs to me now that we might still want the sources though, even if we are not using the hybrid momentum to replace the linear momentum, since we still have the hybrid momentum components in the StateData / plotfiles.

So maybe the alternative is to just disable `USE_HYBRID_MOMENTUM` in `wdmerger` by default, since we are not using it.

zingale:
closing this -- I think the current behavior is correct.  Because otherwise, the state variables for the hybrid momentum are not correct, even though they are never coupled into the system.

PR #2941, opened 2024-08-01
[WIP] create a problem to study impact of species composition gradient on shock burning in 1D

brady-ryan:
Our initial results from `subch_planar` suggest that the behavior of the initial detonation in a ddet SN Ia is dependent on the composition gradient between WD and accretion layer. Here, I am setting up a 1D problem based on `Detonation`. The goal is to apply a sinusoidal function to the composition makeup to represent the 2D gradient.


brady-ryan:
det_0 3
det_0 5
det_0 8

This is starting to look better

brady-ryan:
This now has the transition region about the center of the domain, with a transition width given by `problem.delta`
del_20
del_100

brady-ryan:
This now creates a new parameter, `T_ign`, that represents the initial hotspot temperature. I have separated this from the accretion envelope temperature to better represent the initial He detonation. Additionally, `T_l` now falls to `T_r` in the same fashion as the species composition.

det_Te-2

brady-ryan:
Accomplished what we wanted by adding 1D support to `subch_planar`

PR #2945, opened 2024-08-20
implement full well-balancing in PPM

zingale:
## PR checklist


zingale:
This seems to work.  Here's a version of `bubble_convergence` using the new `ppm_well_balanced`:

bubble_ppm_full_well_balancing

and here it is with normal PPM and reflecting BCs:

bubble_ppm_reflect

Source/hydro/trace_ppm.cpp
```diff
@@ -459,6 +477,10 @@ Castro::trace_ppm(const Box& bx,
     }
 
 
+    // for well-balanced, the velocity sources should not be added
+
+    amrex::Real fac = (castro::ppm_well_balanced && in_hse) ? 0.0_rt : 1.0_rt;
```
zhichen3, line 482: is it better to just skip the source trace tracing for `un` above if we use `ppm_well_balaneced` and `in_hse`?

zingale: I guess so, but somehow I thought this was more clear when reading the code.

Source/hydro/ppm.H
```diff
@@ -244,14 +254,15 @@ ppm_reconstruct_pslope(const Real* rho, const Real* p, const Real* src, const Re
...
+                amrex::Real* Ip, amrex::Real* Im) {
 
-  // Integrate the parabolic profile to the edge of the cell.
+    // Integrate the parabolic profile to the edge of the cell.
 
-  // compute x-component of Ip and Im
-  Real s6 = 6.0_rt * sc - 3.0_rt * (sm + sp);
+    // compute x-component of Ip and Im
+
+    Real s6 = 6.0_rt * sc - 3.0_rt * (sm + sp);
```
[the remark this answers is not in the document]
zingale: fixed

Source/hydro/trace_ppm.cpp
```diff
@@ -229,19 +229,37 @@ Castro::trace_ppm(const Box& bx,
...
+
+    if (castro::use_pslope || castro::ppm_well_balanced) {
         Real trho[nslp];
         Real src[nslp];
 
         load_stencil(q_arr, idir, i, j, k, QRHO, trho);
         load_stencil(srcQ, idir, i, j, k, QUN, src);
 
-        ppm_reconstruct_pslope(trho, s, src, flat, dL, sm, sp);
+        in_hse = ppm_reconstruct_pslope(trho, s, src, flat, dx[idir], sm, sp);
```
zhichen3, line 247: probably better to pass in `dL` rather than `dx[idir]` although it only makes a difference in spherical-theta direction but gravity is in r so there will be no difference, but I still feel its more consistent this way. Same goes with using `dL` to get `p_m_hse` and `p_p_hse` later.

zingale: done

zhichen3 (approved):

Issue #2927, opened 2024-07-24
2D spherical geometry


zingale:
We want to add 2D spherical geometry $(r, \theta)$ on the grid with axisymmetry in $\phi$.

The following tasks need to be done:

- [x] update the geometry source terms in `Castro/Source/sources/Castro_geom.cpp` to have the geometric terms (done in https://github.com/AMReX-Astro/Castro/pull/2955)
- [x] add the primitive variable source terms  (perhaps via #2473) and ensure that d/dtheta has 1/r in interface prediction (done in #2964)
- [x] make sure that the geometry terms are included in the transverse flux correction functions (done in https://github.com/AMReX-Astro/Castro/pull/2961)
- [x] add the geometry factors to diffusion (added in https://github.com/AMReX-Astro/Castro/pull/3048 and further verified in https://github.com/AMReX-Astro/Castro/pull/3054)
- [x] update artifical viscosity (done in https://github.com/AMReX-Astro/Castro/pull/2954)
- [x] update the shock detection algorithm (done in https://github.com/AMReX-Astro/Castro/pull/2959)
- [x] add the correct cross product for the Coriolis force in the rotational source term (done in #2967)
- [x] update AMReX to know about the area and volume factors (done in https://github.com/AMReX-Codes/amrex/pull/4141)
- [x] make constant gravity work in the radial direction (done in https://github.com/AMReX-Astro/Castro/pull/2957)
- [x] update our local area and volume functions (in `Castro_util.H`) to have the same factors (done in https://github.com/AMReX-Astro/Castro/pull/2953)
- [x] address `mom_flux_has_p` (done in https://github.com/AMReX-Astro/Castro/pull/2958)
- [ ] add a `ptheta` flux register analogous to the `pradial` one (done in https://github.com/AMReX-Astro/Castro/pull/2960)
- [x] make sure that the interpolation in AMReX uses the edge volume weights (these were inlined--see discussion in https://github.com/AMReX-Codes/amrex/pull/4141)(Done in https://github.com/AMReX-Codes/amrex/pull/4441)
- [x] update the timestep computation (to use r dtheta) (done in https://github.com/AMReX-Astro/Castro/pull/2962)

zhichen3:
I think also update mom_flux_has_p and related stuff

zhichen3:
also need to modify cfl_violation/timestep constraint since in spherical2d dy is just dtheta but we want physical length , so rdtheta, when checking.

PR #2954, opened 2024-09-08
update divu calculation for spherical 2D geometry

zhichen3:
This updates divu calculation for spherical 2D. After this artificial viscosity should just work I think.
This is to deal with #2927 .

## PR checklist

### DIFF
--- a/Source/hydro/advection_util.cpp
+++ b/Source/hydro/advection_util.cpp
@@ -205,30 +205,30 @@ Castro::divu(const Box& bx,
 
 #if AMREX_SPACEDIM == 1
     if (coord_type == 0) {
-      div(i,j,k) = (q_arr(i,j,k,QU) - q_arr(i-1,j,k,QU)) * dxinv;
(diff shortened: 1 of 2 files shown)


Source/hydro/advection_util.cpp
```diff
@@ -237,31 +237,64 @@ Castro::divu(const Box& bx,
...
 
     } else {
-      if (i == 0) {
-        ux = 0.0_rt;
-        vy = 0.0_rt;  // is this part correct?
-      } else {
+
+        // Spherical R-Theta
+
         Real rl = (i - 0.5_rt) * dx[0] + problo[0];
```
zingale, line 276: I don't think that these are correct

I think they should be

```
rl = i * dx[0] + problo[0]
rr = (i  + 1.0) * dx[0] + problo[0]
rc = (i + 0.5) * dx[0] + problo[0]
```

that makes rc "cell-centered"

zingale: oh... wait... this is supposed to be node centered though

zingale (approved):

PR #2959, opened 2024-09-16
update the shock detection algorithm for r-theta

zingale:
## PR checklist

### DIFF
--- a/Source/hydro/advection_util.cpp
+++ b/Source/hydro/advection_util.cpp
@@ -113,7 +113,6 @@ Castro::shock(const Box& bx,
 #endif
 #endif
 
-#if AMREX_SPACEDIM == 1
     } else if (coord_type == 2) {
 
       // 1-d spherical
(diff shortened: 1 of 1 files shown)


Source/hydro/advection_util.cpp
```diff
@@ -134,7 +142,10 @@ Castro::shock(const Box& bx,
 
     // now compute (grad P - rho g) . dx
     // We subtract off the hydrostatic force, since the pressure that
-    // balances that is not available to make a shock.
+    // balances that is not available to make a shock.  We compute this
+    // as:
+    // P'_{i+1} = P_{i+1} - [ P_i - \int_{x_i}^{x_{i+1}} rho g dx
```
zhichen3, line 147: extra `[` in comment here?

zingale: I was missing the close `]`.  Fixed.

Source/hydro/advection_util.cpp
```diff
@@ -134,7 +142,10 @@ Castro::shock(const Box& bx,
 
     // now compute (grad P - rho g) . dx
     // We subtract off the hydrostatic force, since the pressure that
-    // balances that is not available to make a shock.
+    // balances that is not available to make a shock.  We compute this
+    // as:
+    // P'_{i+1} = P_{i+1} - [ P_i - \int_{x_i}^{x_{i+1}} rho g dx ]
```
zhichen3, line 147: wouldn't that make it `+rho g dx` since there are two minus?

Source/hydro/advection_util.cpp
zingale: I've made the comment clearer

Source/hydro/advection_util.cpp
zingale: I've made the comment clearer and fixed the sign

zhichen3 (approved):

PR #2962, opened 2024-09-19
address timestep and cfl violation for Spherical2D

zhichen3:
## PR summary
address #2927 .  Change timestep and cfl_violation so we use rdtheta, rdx[1], which is the physical cell length instead of just dx[1] for cell length in theta direction for spherical 2d.

## PR checklist


zhichen3:
we might need to change things in `radiation`, especially since dx.array is passed on to `estimate_gamrPr`, but I don't really know how it works for now.

zingale (approved):

PR #2967, opened 2024-09-25
address rotation source in spherical 2d coordinate

zhichen3:
## PR summary
This addresses #2927 regarding the rotation bit.
I mainly separated `get_omega()` to return magnitude of the angular frequency,
then made `get_omega_vec()` to depend on j, since $\Omega \hat{z} = \Omega[\cos(\theta) \hat{r} - \sin(\theta) \hat{\theta})]$.
To deal with the problem of position vector, I just made sure `r` is (r,0,0) before the cross_product call.
I also made computing dt_omega_matrix into a separate function since omega vector now varies.

## PR checklist


Source/rotation/rotation_sources.cpp
```diff
@@ -178,53 +182,6 @@ Castro::corrrsrc(const Box& bx,
       dx[i] = 0.0_rt;
   }
 
-  auto omega = get_omega();
-
-  Real dt_omega[3];
-
-  Array2D<Real, 0, 2, 0, 2> dt_omega_matrix = {};
-
-  if (implicit_rotation_update == 1) {
```
zingale, line 187: what happened to this check on `implicit_rotation_update` ?

zingale: I think it is missing in fill_dt_omega_matrix

zhichen3: `fill_dt_omega_matrix` is called within `if (implicit_rotation_update == 1)` I think.

zingale:
all tests pass:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2024-12-18-001/index.html

Source/rotation/Rotation.H
```diff
@@ -63,12 +93,21 @@ rotational_acceleration(GpuArray<Real, 3>& r, GpuArray<Real, 3>& v,
...
   bool c2 = (castro::rotation_include_coriolis == 1 && coriolis) ? true : false;
 
+  auto r_vec = r;
+
   GpuArray<Real, 3> omega_cross_v;
   cross_product(omega, v, omega_cross_v);
 
   if (c1) {
+      // For Spherical coordinate, the physical position vector is just r rhat,
+      // but the input r will be in the format of (r,theta,0). So change it to (r,0,0)
```
zingale, line 103: why can we do this?

zhichen3: in spherical coord, the position vector is just $r \hat{r}$, so theta should be zeroed out when we do the cross products I think.

zingale: Oh, this is because it is the position vector from the origin?

Source/rotation/rotation_sources.cpp
zingale: yes,  you are right.

zingale (approved):

zingale:
okay, everything looks good.

can you also update the docs?
https://amrex-astro.github.io/Castro/docs/rotation.html

zhichen3:
> okay, everything looks good.
> 
> can you also update the docs? https://amrex-astro.github.io/Castro/docs/rotation.html

done

PR #2988, opened 2024-11-07
add a distance() function and subtract off problem::center in position()

zhichen3:
## PR summary

1) Adds a separate function to calculate distance, mainly because the old way gets the 2d spherical case wrong.
2) Also subtracts off problem::center in position(). I think that's done in all cases where we call position except in rad_util.H, but that happens only in nonCartesian coordinate, and problem::center should be 0 at those cases. So I think it doesn't affect things.
3) Most places calculate the position separately, I changed it to using position()
3) Also make sure problem::center is 0 for spherical2d

## PR checklist


zingale:
a lot of failures: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2024-11-07-002/index.html

zhichen3:
will separate this out into separate prs

PR #2993, opened 2024-11-13
update source_flag to work with spherical and allow geometric source terms for 1d spherical

zhichen3:
## PR summary
update source_flag to work with spherical and allow geometric source terms for 1d spherical

## PR checklist

### DIFF
--- a/Source/sources/Castro_sources.cpp
+++ b/Source/sources/Castro_sources.cpp
@@ -61,10 +61,10 @@ Castro::source_flag(int src)
 #endif
 
     case geom_src:
-        if (geom.Coord() == 1) {
-          return true;
-        } else {
+        if (geom.Coord() == 0) {
           return false;
+        } else {
+          return true;
         }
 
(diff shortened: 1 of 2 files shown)


zingale:
good catch!

Source/sources/Castro_sources.cpp
```diff
@@ -61,10 +61,10 @@ Castro::source_flag(int src)
 #endif
 
     case geom_src:
-        if (geom.Coord() == 1) {
-          return true;
-        } else {
+        if (geom.Coord() == 0) {
```
zingale, line 64: there aren't any geometry sources for 1-d spherical or cylindrical either, right?

zhichen3: I think it depends on what does 1d spherical mean exactly. Are we assuming purely 1d with velocities in theta and phi direction being 0 or are we assuming 3d but with theta and phi symmetry?

zingale (approved):

PR #3019, opened 2025-01-16
fix a tiling issue with the shock detection

zingale:
we need to grow the q FAB, not use a growntilebox() this only affects CPUs, and only seemed to be an issue in 3D

## PR checklist


zingale:
this fixes `Array4` bounds errors seen with `subchandra` in 3D on CPU.

maxpkatz (approved):

PR #3020, opened 2025-01-16
pin some CI at 22.04

zingale:
CUDA needs the older compilers for now
exact Riemann seems to have some roundoff issues

## PR checklist


zingale:
actually, the issue is Microphysics changes for most of these

PR #3030, opened 2025-02-07
Improvements to the `flame_wave` flame front analysis scripts

yut23:
### front_tracker.py

* dataset processing can run in parallel
* added code to make interactive plots of each dataset that show the raw slice data, the averaged profile, and the positions for each metric
* metric thresholds are calculated using both the local maximum and global maximum at the same time (under `<field>[<fac>%]` and `<field>[<fac>%]_gmax`, respectively)
* the local maximum for each field is stored in the output file, in the `max_<field>` column
* if a field value never goes below the threshold for a metric, its position will be stored as NaN
* --[xyz]lim arguments now work, despite some yt bugs
* allow ignoring plotfiles after a certain time with --tmax
* default FRB resolution is set to match the finest level, and is automatically adjusted if custom limits are used

### flame_speed.py

Now takes min and max times for the linear regression as optional arguments, and can plot multiple metrics at the same time.

## PR checklist


zingale (approved):

yut23:
Screenshot of the interactive plots:
Screenshot from 2025-02-07 15-27-08

yut23:
The missing checks might be because I opened this PR against main initially. I just ran codespell locally, and it didn't find anything.

zingale:
I'm fine with it.  You can merge if you want.

PR #3041, opened 2025-03-04
plot analytic solution in the plotting script for thermal diffusion unit test

zhichen3:
## PR summary
plot analytic solution in profile analysis.

## PR checklist


Exec/unit_tests/diffusion_test/analysis/profiles.py
```diff
@@ -22,6 +22,30 @@ def rgba_to_hex(rgba):
     b = int(rgba[2]*255.0)
     return '#{:02X}{:02X}{:02X}'.format(r, g, b)
 
+def get_analytic_profile(x_coord, time, problo, probhi,
```
zingale, line 25: the analytic solution is a derived variable for this problem, so can't we just read it from the plotfile?

zhichen3: yah, you're right. Didn't realize about that.

zingale (approved):

Issue #3042, opened 2025-03-04
FPE when running in 3D with castro.transverse_use_eos=1


zingale:
To demonstrate, build `wdmerger` as:

```
make TEST=TRUE DIM=3 -j 4
```

then run as:

```
./Castro3d.gnu.TEST.MPI.ex inputs amr.n_cell=64 64 64 castro.transverse_use_eos=1 amrex.fpe_trap_invalid=1
```

and you'll see an FPE right away.

PR #3043, opened 2025-03-04
fix NaN issue when using transverse_use_eos

zhichen3:
## PR summary
This fixes issue #3042 . Temperature was 0, causing tempi to be infinity during the eos call.
I'm able to run wdmerger to the step # 3.

## PR checklist

### DIFF
--- a/Source/hydro/trans.cpp
+++ b/Source/hydro/trans.cpp
@@ -446,6 +446,8 @@ Castro::actual_trans_single(const Box& bx,  // NOLINT(readability-convert-member
                 eos_rep_t eos_state;
                 eos_state.rho = rrnewn;
                 eos_state.e = qo_arr(i,j,k,QREINT) / rrnewn;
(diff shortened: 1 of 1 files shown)


Source/hydro/trans.cpp
```diff
@@ -446,6 +446,8 @@ Castro::actual_trans_single(const Box& bx,  // NOLINT(readability-convert-member
                 eos_rep_t eos_state;
                 eos_state.rho = rrnewn;
                 eos_state.e = qo_arr(i,j,k,QREINT) / rrnewn;
+                eos_state.T = small_temp;
+
```
zingale, line 450: I think we should use `T_guess` instead of `small_temp` -- this is what the `T_guess` runtime parameter is intended for.

zhichen3: fixed

zingale (approved):

PR #3046, opened 2025-03-11
fix spherical 2d boundary check

zhichen3:
## PR summary
cellsize is not set at this point. So manually set it here.

## PR checklist


Source/driver/Castro.cpp
```diff
@@ -322,8 +322,16 @@ Castro::read_params ()
         {
           amrex::Abort("ERROR:Castro::read_params: Theta must be within [0, Pi] for spherical coordinate system in 2D");
         }
+#ifdef MHD
+        int ngrow = 6;
+#else
+        int ngrow = 4;
+#endif
```
zingale, line 329: I think it will be safer if we only set these values once.  So what about moving the logic

```
 #ifdef MHD                                                                                                         
  NUM_GROW = 6;                                                                                                    
 #else                                                                                                              
  NUM_GROW = 4;                                                                                                    
 #endif   
```

out of `Castro_setup.cpp` and putting it into `castro_limits.H`.  We might want to make them `constexpr`.

same with the `NUM_GROW_SRC` -- we can use preprocessor to handle the `TRUE_SDC` check.

zingale: we probably want to do that first in a separate PR

zingale (approved):

PR #3055, opened 2025-04-08
update ca_derdivu, ca_dermagvort, and ca_derdiffterm derived variables for 2d spherical geometry

zhichen3:
## PR summary
updated `ca_derdivu()`, `ca_dermagvort()`, and `ca_derdiffterm()` for 2d spherical geometry.

Also I think diffterm is calculated incorrectly for cylindrical geometry? I think we should multiply by face-centered r
instead of cell centered, since `kgradT` is on the face.

## PR checklist

### DIFF
--- a/Source/driver/Derive.cpp
+++ b/Source/driver/Derive.cpp
@@ -405,23 +405,29 @@ extern "C"
 
         } else if (coord_type == 1) {
           // axisymmetric coords (2-d)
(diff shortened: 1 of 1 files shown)


zingale:
I think you are right about the diff term in cylindrical.  Why did we do it this way?

zingale (approved):

Issue #2652, opened 2023-11-17
should executable include network name?


zingale:
A common issue with users is that when we want to run the same problem with different networks, we sometimes forget to do a `make clean` between builds.  For some combination of networks, the build still works, but it is not clear what the executable will be doing.

One way to prevent this is to add the network name to the executable.  Then the build directories in `tmp_build_dir` will also be separate and no `make clean` would be needed to switch networks.

I think that networks are unique among the microphysics in something that we change frequently within a problem setup.

Comments welcomed.

PR #3073, opened 2025-05-09
when we "make" without a "clean", ensure that the same net is used

zingale:
a common error is to build the executable, do a run, and then change the network by building with NETWORK_DIR set, without a "make clean". This leads to an undefined behavior.  This adds a check to ensure that when we make we are using the same network as done originally, and if not, aborts with a message saying to do "make clean"

## PR checklist


zingale:
this is a different fix to the same problem identified in #2652

yut23 (approved):

PR #3102, opened 2025-07-15
update the H/He flame

zingale:
this is a 1-d setup that works
note that the flame is very slow, so this is a very coarse resolution

## PR checklist


zingale:
Here's the flame that results:

image

PR #3109, opened 2025-07-28
use load_stencil to PPM / MHD

zingale:
## PR checklist

### DIFF
--- a/Source/mhd/mhd_plm.cpp
+++ b/Source/mhd/mhd_plm.cpp
@@ -280,28 +280,16 @@ Castro::plm(const Box& bx,
 
     // species
     for (int n = 0; n < NumSpec; n++) {
-      Real un;
-      Real X[5];
+      Real un{};
+      Real X[nslp];
+
+      load_stencil(s, idir, i, j, k, QFS+n, X);
 
       if (idir == 0) {
-        for (int m = 0; m < 5; m++) {
-          int offset = m - 2;
-          X[m] = s(i+offset,j,k,QFS+n);
-        }
         un = s(i,j,k,QU);
-
       } else if (idir == 1) {
-        for (int m = 0; m < 5; m++) {
-          int offset = m - 2;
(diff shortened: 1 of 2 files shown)


zingale:
I think that this is ready for review.  I believe I caught all the places where we hardcoded the number of points in the stencil (5).

Source/mhd/mhd_plm.cpp
```diff
@@ -186,13 +162,13 @@ Castro::plm(const Box& bx,
...
         // construct the ii-th primitive variables
-        Real W[5] = {};
+        Real W[nslp] = {};
         for (int n = 0; n < NEIGN; n++) {
-          W[0] += leig(ii,n) * Q[n][0];
-          W[1] += leig(ii,n) * Q[n][1];
-          W[2] += leig(ii,n) * Q[n][2];
-          W[3] += leig(ii,n) * Q[n][3];
-          W[4] += leig(ii,n) * Q[n][4];
+          W[im2] += leig(ii,n) * Q[n][im2];
```
maximumcats, line 167: Should this construction be generalized too?

zingale: I think that this is alreadyy as general as it can be, since this is just reconstructing the primitive variables from the characteristic variables.  There is not a load_stencil operation for that, since it is l . dq, where l is the left eigenvector.

maximumcats: What I mean is that there could be an inner loop over the number of points in the stencil, so that if it were a seven point stencil you wouldn't need to add W[im3] and W[ip3] lines. Unless I am also misunderstanding what is happening inside this loop.

zingale: I see what you are saying.  Yes, you are right.  I've updated it.  Let's see if the MHD CI test passes.

maximumcats (approved):

PR #3119, opened 2025-08-05
Only include leaf zones in estdt_burning when not subcycling

maximumcats:
## PR summary

When we're not using AMR subcycling, we don't burn in zones that aren't leaf zones. So we don't need to consider them in the estDt calculation either. (Ultimately there could be a broader refactor to skip the entire advance in zones that aren't leaf zones.)

## PR checklist


maximumcats:
Note: this PR will change simulation results in some cases. If the coarse zones that are covered by fine zones have a more restrictive timestep than the fine zones, they may have controlled the overall advance timestep, and that would no longer be true.

zingale (approved):

PR #3129, opened 2025-08-21
Documentation: Clarify that `CUDA_ARCH` can only be overridden from command line

hafeoz:
## PR summary

Clarify that `CUDA_ARCH` can only be overridden from the command line; setting `CUDA_ARCH` in `GNUmakefile` will be overridden and has no effect.

## PR motivation

Overriding this parameter from the command line will [prevent AMReX from changing it](https://www.gnu.org/software/make/manual/html_node/Overriding.html), while specifying it in `GNUmakefile` will be overridden.

I've learned it the hard way in AMReX-Codes/amrex#4621.

## PR checklist


WeiqunZhang:
One could still specify it in GNUmakefile with `override CUDA_ARCH = 80`.

maximumcats (approved):

PR #3142, opened 2025-09-19
add check to make sure hydrostatic pressure at interface is also positive for HSE methods

zhichen3:
## PR summary
This fixes the issue where hydrostatic pressure at the interface could be negative due to external source terms.

## PR checklist


Source/hydro/ppm.H
```diff
@@ -186,12 +186,19 @@ ppm_reconstruct_pslope(const amrex::Real* rho, const amrex::Real* p, const amrex
     amrex::Real pm1_hse = p0_hse - 0.25_rt*dx * (rho[i0] + rho[im1]) * (src[i0] + src[im1]);
     amrex::Real pm2_hse = pm1_hse - 0.25_rt*dx * (rho[im1] + rho[im2]) * (src[im1] + src[im2]);
 
+    // compute hydrostatic pressure at the interface
+
+    amrex::Real p_p_hse = p0_hse + 0.5_rt*dx * rho[i0] * src[i0];
```
zingale, line 191: since you compute these here, you can just reuse them below (lines 238 and 239)

zhichen3: done

PR #3160, opened 2025-10-24
fix UEINT -> UEDEN in initData()

zhichen3:
## PR checklist


zingale:
This was only an issue if we were doing 4th order and we were initializing the data as cell-averages directly.

PR #3167, opened 2025-11-04
remove unused declare in radiation

zingale:
## PR checklist


Source/radiation/HypreMultiABec.cpp
```diff
@@ -3908,8 +3908,6 @@ void HypreMultiABec::boundaryFlux(int level,
 #pragma omp parallel
 #endif
     {
-        Real foo=1.e200;
```
zhichen3: looks like there is a same thing in `HypreABec.cpp`

PR #3169, opened 2025-11-04
 Updating wdmerger set up to disable burning in shocks

khanakbhargava:
## PR checklist


khanakbhargava:
Benchmark from frontier did not work. I got it from groot then. That codespell error needs to be checked

PR #3181, opened 2025-12-12
in the transverse update we recomputed density every passive iteration

zingale:
this pulls that out.
A later optimization is to precompute the inverse, but that introduces roundoff diffs, so I'll save that for a follow-on PR

## PR checklist


zingale:
this seems to perhaps slow things down?

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2025-12-17-001/index.html

PR #3205, opened 2026-02-05
address docs accessibility issues

zingale:
fix color contrast
add back link underlining

## PR checklist


Myndex:
Underlining by default can be anti-accessible in many cases, as it can significantly interfere with readability. Underlines in blocks of body text in particular.

Underlines are far from "best practice". Like most things related to accessibility, user preference is king.

Here is a link to a draft of a white paper and related discussion we are working on:

https://github.com/Myndex/SAPC-APCA/discussions/65

In short, the context of where the underline is, and the white space and density of the text, all play a role here. 



Cheers,


Andrew Somers
Director of Research
Inclusive Reading Technologies
https://readtech.org/ARC/

Issue #3209, opened 2026-02-24
`cot()` in `Castro_math.H` has always-true assert and incorrect angle reflection for `x > pi`


WeiqunZhang:
### Location
`Source/driver/Castro_math.H:28`, `Source/driver/Castro_math.H:30`

### Problem
- `AMREX_ASSERT(x != 0.0_rt || x != M_PI);` is logically always true (should not use `||`).
- For `x > pi`, `x = pi - x` does not preserve `cot(x)`; sign/period behavior is wrong for many values.

### Impact
- Missed protection against singular inputs (`0`, `pi`).
- Potentially incorrect geometric source terms where `cot(theta)` is used.

### Steps to Reproduce
1. Evaluate `cot(5*pi/4)` in current implementation.
2. Compare against `std::cos(x)/std::sin(x)` at same `x`.
3. Observe sign mismatch.

### Expected Behavior
`cot(x)` should respect periodicity and avoid singular points robustly.

### Actual Behavior
Assertion does not guard singularities; out-of-range normalization is incorrect.

### Proposed Fix
```cpp
AMREX_GPU_HOST_DEVICE AMREX_INLINE
amrex::Real cot(amrex::Real x) {
    x = std::fmod(x, M_PI);
    if (x < 0.0_rt) x += M_PI;

    constexpr amrex::Real eps = 1.e-14_rt;
    AMREX_ASSERT(x > eps && x < M_PI - eps);

    return std::cos(x) / std::sin(x);
}
```

Prepared by Codex

PR #3261, opened 2026-02-24
fix amrex_assert on cot

zhichen3:
## PR summary
Fix always true assert pointed out in #3209. 
The sign issue is done this way on purpose to achieve reflection boundary condition on 0 and pi.
I added more comment to explain.

## PR checklist

### DIFF
--- a/Source/driver/Castro_math.H
+++ b/Source/driver/Castro_math.H
@@ -25,8 +25,12 @@ cross_product(amrex::GpuArray<amrex::Real, 3> const& a,
 AMREX_GPU_HOST_DEVICE AMREX_INLINE
 amrex::Real cot(amrex::Real x) {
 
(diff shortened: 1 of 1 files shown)


zhichen3:
I'm not sure if I should check via `eps` as suggested in #3209  or not.

maximumcats:
> I'm not sure if I should check via `eps` as suggested in #3209 or not.

I think it would be good to do that, yes, to avoid potentially spurious behavior in cos and sin.

zhichen3:
> > I'm not sure if I should check via `eps` as suggested in #3209 or not.
> 
> I think it would be good to do that, yes, to avoid potentially spurious behavior in cos and sin.

Ok. I changed to check `sin(x) > eps` instead.

zingale (approved):

Issue #3235, opened 2026-02-24
Second-order reactive SDC update passes `A_new[m_end]` where `R_old[m_start]` is required


WeiqunZhang:
## Summary
In `do_sdc_update()`, the call to `sdc_update_o2()` passes `A_n = A_new[m_end]` into the argument slot named `R_m_old`. Inside `sdc_update_o2()`, that argument is used in t\
he initial predictor as a reaction source term. This mixes advection and reaction sources and corrupts the first-iteration nonlinear guess.

## Location
- `Source/sdc/Castro_sdc.cpp:179`
- `Source/sdc/Castro_sdc.cpp:185`
- `Source/sdc/Castro_sdc_util.H:120`
- `Source/sdc/Castro_sdc_util.H:154`

## Problem Details
Current call site:
```cpp
auto A_m = (*A_new[m_start]).array(mfi);
auto A_n = (*A_new[m_end]).array(mfi);
...
sdc_update_o2(i, j, k, k_m, k_n, A_m, A_n, C_arr, dt_m, sdc_iteration, m_start);
```

But `sdc_update_o2` interprets the second source array as old reaction source:
```cpp
U_new[n] = U_old[n] + dt_m * A_m(i,j,k,n) + dt_m * R_m_old(i,j,k,n);
```

## Impact
- Incorrect initial guess for reactive implicit solve in O2 SDC (`sdc_iteration == 0`).
- Can degrade convergence/accuracy and produce biased updates by injecting advection source where reaction source is expected.

## Suggested Patch
```diff
diff --git a/Source/sdc/Castro_sdc.cpp b/Source/sdc/Castro_sdc.cpp
--- a/Source/sdc/Castro_sdc.cpp
+++ b/Source/sdc/Castro_sdc.cpp
@@
             auto k_m = (*k_new[m_start]).array(mfi);
             auto k_n = (*k_new[m_end]).array(mfi);
             auto A_m = (*A_new[m_start]).array(mfi);
-            auto A_n = (*A_new[m_end]).array(mfi);
+            auto R_m_old = (*R_old[m_start]).array(mfi);
             auto C_arr = C2.array();
@@
-                sdc_update_o2(i, j, k, k_m, k_n, A_m, A_n, C_arr, dt_m, sdc_iteration, m_start);
+                sdc_update_o2(i, j, k, k_m, k_n, A_m, R_m_old, C_arr, dt_m, sdc_iteration, m_start);
             });
```

Prepared by Codex

PR #3275, opened 2026-03-01
fix initial guess for the 2nd order true SDC nonlinear solve

zingale:
we were passing in an advective term instead of the reactive source.
This only affects the convergence rate -- the solve itself was okay.
Also clean up some formatting in the C++

## PR checklist


zingale:
closes #3235

zingale:
Note: review this with whitespace ignored.

zhichen3 (approved):

Issue #3257, opened 2026-02-24
`sdc_newton_subdivide()` retry path does not reset `U_begin` to `U_old`


WeiqunZhang:
## Summary
When a subdivided Newton attempt fails after one or more successful substeps, `sdc_newton_subdivide()` doubles `nsub` and retries, but does not restore `U_begin` to the original `U_old` before the new attempt.

## Location
- `Source/sdc/sdc_newton_solve.H:266`
- `Source/sdc/sdc_newton_solve.H:270`
- `Source/sdc/sdc_newton_solve.H:305`

## Problem Details
`U_begin` is initialized from `U_old` once before the outer retry loop, then advanced in-place after each successful substep:
```cpp
for (int n = 0; n < NUM_STATE; ++n) {
    U_begin[n] = U_new[n];
}
```

If a later substep fails, the loop breaks and retries with larger `nsub`, but `U_begin` is left at the partially advanced state from the failed attempt.

## Impact
- Retry integrations can start from the wrong state.
- Can distort the intended adaptive-subdivision behavior and produce biased updates/failures.

## Suggested Patch
Reset `U_begin` at the start of each outer retry attempt:

## Impact
- Retry integrations can start from the wrong state.
- Can distort the intended adaptive-subdivision behavior and produce biased updates/failures.

## Suggested Patch
Reset `U_begin` at the start of each outer retry attempt:

```diff
diff --git a/Source/sdc/sdc_newton_solve.H b/Source/sdc/sdc_newton_solve.H
--- a/Source/sdc/sdc_newton_solve.H
+++ b/Source/sdc/sdc_newton_solve.H
@@
     while (nsub < MAX_NSUB && ierr != newton::NEWTON_SUCCESS) {
+        for (int n = 0; n < NUM_STATE; ++n) {
+            U_begin[n] = U_old[n];
+        }
         if (nsub > 1) {
             for (int n = 0; n < NUM_STATE; ++n) {
                 U_new[n] = U_old[n];
             }
```

Prepared by Codex

PR #3276, opened 2026-03-02
fix the initial state in the SDC Newton sub-divide algorithm

zingale:
if we bailed on an attempt with nsub sub-intervals, then we never reset the starting state to U^n for the attempt with 2*nsub intervals.

## PR checklist


zingale:
closes #3257

maximumcats (approved):

Issue #3233, opened 2026-02-24
`do_hscf_solve()` uses level-0 cell size for all AMR levels in SCF diagnostics


WeiqunZhang:
## Summary
Inside the per-level diagnostics loop, `dx` is fetched with `parent->Geom(level).CellSizeArray()` instead of `parent->Geom(lev).CellSizeArray()`. Since `scf_relaxation()` runs on `level == 0`, this uses level-0 spacing for every level.

## Location
- `Source/scf/scf_relax.cpp:617`

## Problem Details
Current code:
```cpp
for (int lev = 0; lev <= finest_level; ++lev) {
    auto geomdata = parent->Geom(lev).data();
    const auto dx = parent->Geom(level).CellSizeArray();
    ...
}
```

`dV` and coordinate-based diagnostics in this loop then use inconsistent geometry on refined levels.

## Impact
- Incorrect mass/energy/virial diagnostics during SCF iterations when AMR has `lev > 0`.
- Convergence reporting can be misleading because diagnostic integrals are mis-scaled.

## Suggested Patch
```diff
diff --git a/Source/scf/scf_relax.cpp b/Source/scf/scf_relax.cpp
--- a/Source/scf/scf_relax.cpp
+++ b/Source/scf/scf_relax.cpp
@@
-            const auto dx = parent->Geom(level).CellSizeArray();
+            const auto dx = parent->Geom(lev).CellSizeArray();
```

Prepared by Codex

PR #3278, opened 2026-03-02
fix dx in the SCF diagnostic code

zingale:
we were not using the dx on the current level

## PR checklist


zingale:
closes #3233

I don't understand why the diagnostics don't use a mask though -- it seems we are counting the mass on all levels.

maximumcats (approved):

Issue #3243, opened 2026-02-24
`errorEst` computes `ltime` for post-step regrid but still evaluates built-in tags at `time`


WeiqunZhang:
## Summary
When `post_step_regrid` is active, `Castro::errorEst` correctly computes `ltime` as the new state time, but built-in AMR error tags still call `derive(...)` and `etag(...)` with the original `time` argument. This defeats the intended post-step regrid time correction for built-in tagging criteria.

## Location
- `Source/driver/Castro.cpp:3501`
- `Source/driver/Castro.cpp:3516`
- `Source/driver/Castro.cpp:3518`

## Problem Details
Current logic:
```cpp
Real ltime = time;
if (post_step_regrid) {
    ltime = get_state_data(State_Type).curTime();
}
...
mf = derive(etag.Field(), time, etag.NGrow());
etag(tags, mf.get(), TagBox::CLEAR, TagBox::SET, time, level, geom);
```

So user/problem tagging uses `ltime`, but built-in tags continue using `time`.

## Impact
- Inconsistent regrid tagging times during post-step regrid.
- Built-in indicators can tag based on stale (old-time) data while user/problem restrictions use new-time intent.

## Suggested Patch
Use `ltime` consistently in built-in tagging:
```diff
diff --git a/Source/driver/Castro.cpp b/Source/driver/Castro.cpp
--- a/Source/driver/Castro.cpp
+++ b/Source/driver/Castro.cpp
@@
     for (const auto & etag : error_tags) {
         std::unique_ptr<MultiFab> mf;
         if (! etag.Field().empty()) {
-            mf = derive(etag.Field(), time, etag.NGrow());
+            mf = derive(etag.Field(), ltime, etag.NGrow());
         }
-        etag(tags, mf.get(), TagBox::CLEAR, TagBox::SET, time, level, geom);
+        etag(tags, mf.get(), TagBox::CLEAR, TagBox::SET, ltime, level, geom);
     }
```

Prepared by Codex

PR #3284, opened 2026-03-02
for post-step regrid, using the correct time in the error estimators

zingale:
we should use ltime, which is correctly set depending on the value of `post_step_regrid`

## PR checklist


zingale:
closes #3243

maximumcats (approved):

Issue #3218, opened 2026-02-24
`apply_metric` in RZ assumes radial origin at zero and ignores `ProbLo(0)`


WeiqunZhang:
## Summary
`apply_metric` multiplies by radius `r` using index-based formulas (`i*dx[0]`, `(i+0.5)*dx[0]`) without adding the domain radial offset. For domains with `geometry.prob_lo[\
0] != 0`, metric factors are wrong.

## Location
- `Source/driver/MGutils.cpp:27`
- `Source/driver/MGutils.cpp:33`
- `Source/driver/MGutils.cpp:40`

## Problem Details
Current code computes:
- cell center `r = (i+0.5)*dx[0]`
- x-edge `r = i*dx[0]`
- y-edge `r = (i+0.5)*dx[0]`

This is only valid when radial `ProbLo(0) == 0`.

## Impact
- Incorrect elliptic operator metric weighting for shifted radial domains.
- Can bias solves in setups that do not start radial coordinate at zero.

## Suggested Patch
```diff
diff --git a/Source/driver/MGutils.H b/Source/driver/MGutils.H
--- a/Source/driver/MGutils.H
+++ b/Source/driver/MGutils.H
@@
 void apply_metric(const amrex::Box& bx,
                   amrex::Array4<amrex::Real> const& rhs, const amrex::Box& rbx,
                   amrex::Array4<amrex::Real> const& ecx, const amrex::Box& xbx,
 #if AMREX_SPACEDIM >= 2
                   amrex::Array4<amrex::Real> const& ecy, const amrex::Box& ybx,
 #endif
                   amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> dx,
+                  amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> problo,
                   const int coord_type);
```

```diff
diff --git a/Source/driver/MGutils.cpp b/Source/driver/MGutils.cpp
--- a/Source/driver/MGutils.cpp
+++ b/Source/driver/MGutils.cpp
@@
 void
 apply_metric(const Box& bx,
              Array4<Real> const& rhs, const Box& rbx,
              Array4<Real> const& ecx, const Box& xbx,
 #if AMREX_SPACEDIM >= 2
              Array4<Real> const& ecy, const Box& ybx,
 #endif
              GpuArray<Real, AMREX_SPACEDIM> dx,
+             GpuArray<Real, AMREX_SPACEDIM> problo,
              const int coord_type)
@@
[...]

PR #3288, opened 2026-03-03
Add problo in applymetric

zhichen3:
## PR summary
addresses #3218

## PR checklist

### DIFF
--- a/Source/driver/MGutils.H
+++ b/Source/driver/MGutils.H
@@ -14,6 +14,7 @@ apply_metric(const Box& bx,
 #if AMREX_SPACEDIM >= 2
              Array4<Real> const& ecy, const Box& ybx,
 #endif
+             GpuArray<Real, AMREX_SPACEDIM> problo,
              GpuArray<Real, AMREX_SPACEDIM> dx,
              const int coord_type);
 
--- a/Source/driver/MGutils.H
+++ b/Source/driver/MGutils.H
@@ -14,8 +14,8 @@ apply_metric(const Box& bx,
 #if AMREX_SPACEDIM >= 2
              Array4<Real> const& ecy, const Box& ybx,
 #endif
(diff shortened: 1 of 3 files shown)


Source/driver/MGutils.cpp
```diff
@@ -10,6 +10,7 @@ apply_metric(const Box& bx,
 #if AMREX_SPACEDIM >= 2
              Array4<Real> const& ecy, const Box& ybx,
 #endif
+             GpuArray<Real, AMREX_SPACEDIM> problo,
```
zingale: should these be `const` references?

Source/driver/MGutils.H
```diff
@@ -14,6 +14,7 @@ apply_metric(const Box& bx,
 #if AMREX_SPACEDIM >= 2
              Array4<Real> const& ecy, const Box& ybx,
 #endif
+             GpuArray<Real, AMREX_SPACEDIM> problo,
```
zingale: same here , `const` reference?

Source/driver/MGutils.cpp
zhichen3: fixed

Source/driver/MGutils.H
zhichen3: fixed

Source/driver/MGutils.cpp
```diff
@@ -10,7 +10,8 @@ apply_metric(const Box& bx,
 #if AMREX_SPACEDIM >= 2
              Array4<Real> const& ecy, const Box& ybx,
 #endif
-             GpuArray<Real, AMREX_SPACEDIM> dx,
+             const GpuArray<Real, AMREX_SPACEDIM> problo,
```
zingale: I still think that this should be passed as a reference to avoid a copy.  Same with dx.

zingale (approved):

Issue #3211, opened 2026-02-24
`Castro_mol_hydro.cpp` HYBRID_MOMENTUM path uses undefined `qx_arr`/`qy_arr` (and lacks 1D guard)


WeiqunZhang:
### Location
`Source/hydro/Castro_mol_hydro.cpp:647`, `Source/hydro/Castro_mol_hydro.cpp:648`

### Problem
Inside the `#ifdef HYBRID_MOMENTUM` block, the code computes a pressure-gradient correction using `qx_arr` and `qy_arr`, but neither is declared in this function scope.

This is inconsistent with the analogous CTU implementation, which explicitly defines:
- `Array4<Real> const qx_arr = (qe[0]).array();`
- `Array4<Real> const qy_arr = (qe[1]).array();` (guarded for `AMREX_SPACEDIM >= 2`)

Additionally, the MOL block unconditionally references the y-term (`qy_arr`, `dx_arr[1]`) without an `AMREX_SPACEDIM >= 2` guard.

### Impact
- Builds with `HYBRID_MOMENTUM` enabled can fail due to undefined identifiers.
- In 1D builds, the unguarded y-term is structurally invalid.

### Steps to Reproduce
1. Enable `HYBRID_MOMENTUM` and compile hydro MOL code.
2. Compile `Source/hydro/Castro_mol_hydro.cpp`.
3. Observe compile errors for `qx_arr` / `qy_arr` (and/or dimensional invalidity in 1D).

### Expected Behavior
HYBRID_MOMENTUM correction should compile and use properly defined Godunov pressure arrays with dimensional guards.

### Actual Behavior
The code references undeclared arrays and unguarded y-direction terms.

### Proposed Fix
Define `qx_arr`/`qy_arr` similarly to CTU and guard the y-term by dimension.

```cpp
#ifdef HYBRID_MOMENTUM
auto dx_arr = geom.CellSizeArray();
Array4<Real const> const qx_arr = qe[0].array();
#if AMREX_SPACEDIM >= 2
Array4<Real const> const qy_arr = qe[1].array();
#endif

amrex::ParallelFor(bx,
[=] AMREX_GPU_DEVICE (int i, int j, int k) noexcept
{   
    GpuArray<Real, 3> loc;
    position(i, j, k, geomdata, loc);

    for (int dir = 0; dir < AMREX_SPACEDIM; ++dir)
        loc[dir] -= problem::center[dir];

[...]

PR #3292, opened 2026-03-03
fix hybrid momentum for MOL in 1D

zingale:
## PR checklist


zingale:
closes #3211

zhichen3:
I think hybrid momentum can only be used in 3d cartesian -- otherwise we would raise a compilation error.

zingale:
ah indeed -- we don't check this in Castro.cpp, but rather in Make.Castro.  

I guess we're fine here then.

Issue #3245, opened 2026-02-24
`test_composite_phi` indexes residual vectors with AMR level IDs instead of local offsets


WeiqunZhang:
## Summary
`Gravity::test_composite_phi` allocates `phi`, `g_rhs`, and `res` with `nlevels = finest_level_local - crse_level + 1`, but later indexes `res` using absolute AMR level indices. When `crse_level > 0`, this causes out-of-bounds indexing in both the residual averaging loop and residual norm reporting loop.

## Location
- `Source/gravity/Gravity.cpp:1309`
- `Source/gravity/Gravity.cpp:1312`
- `Source/gravity/Gravity.cpp:1318`

## Problem Details
Current code:
```cpp
for (int amr_lev = finest_level_local-1; amr_lev >= 0; --amr_lev)
{
    const IntVect& ratio = parent->refRatio(amr_lev);
    int ilev = amr_lev - crse_level;
    amrex::average_down(*res[ilev+1], *res[ilev], 0, 1, ratio);
}

for (int amr_lev = crse_level; amr_lev <= finest_level_local; ++amr_lev) {
    Real resnorm = res[amr_lev]->norm0();
    ...
}
```

`res` is indexed from `0..nlevels-1`, so indexing by `amr_lev` is only valid when `crse_level == 0`. The first loop also iterates below `crse_level`, making
`ilev` negative.

## Impact
- Potential out-of-bounds access and undefined behavior when `test_composite_phi` is called with `crse_level > 0`.
- Incorrect or unstable debug residual checks during nonzero-base-level solve diagnostics.

## Suggested Patch
```diff
diff --git a/Source/gravity/Gravity.cpp b/Source/gravity/Gravity.cpp
--- a/Source/gravity/Gravity.cpp
+++ b/Source/gravity/Gravity.cpp
@@
-    for (int amr_lev = finest_level_local-1; amr_lev >= 0; --amr_lev)
+    for (int amr_lev = finest_level_local-1; amr_lev >= crse_level; --amr_lev)
     {   
         const IntVect& ratio = parent->refRatio(amr_lev);
         int ilev = amr_lev - crse_level;
         amrex::average_down(*res[ilev+1], *res[ilev],
                              0, 1, ratio);
     }
@@  
     for (int amr_lev = crse_level; amr_lev <= finest_level_local; ++amr_lev) {
-        Real resnorm = res[amr_lev]->norm0();
+        int ilev = amr_lev - crse_level;
[...]

PR #3295, opened 2026-03-03
clarify test_composite_phi coarse level

zingale:
this adds an assert that we only call this with crse_level = 0 but also fixes some loop bounds to use crse_level in case we change things in the future (we will need further testing then)

## PR checklist


zingale:
closes #3245

maximumcats (approved):

Issue #3305, opened 2026-03-07
add ability to stop Detonation once we reach a certain fraction of the domain


zingale:
For most detonation runs, we propagate from left to right.  Once the detonation leaves the domain, we should stop.  We can use:

the AMReX `maxIndex()` function on energy generation to see where the detonation is.  We can do this on level 0, perhaps in the `problem_post_timestep()`

zingale:
this has been completed

PR #3318, opened 2026-03-19
add a method to abort the Detonation once it reaches the edge of the domain

zingale:
## PR checklist

### DIFF
--- a/Exec/science/Detonation/Prob.cpp
+++ b/Exec/science/Detonation/Prob.cpp
@@ -1,0 +1,40 @@
+// Implementations of functions in Problem.H go here
+
+#include <AMReX_REAL.H>
+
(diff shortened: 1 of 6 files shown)


zingale:
this seems to work.  I've tested this on single proc and 3 procs.

closes #3305

zingale:
detonation regression tests pass.

I need to understand the collision problem to see if the default of 0.95 of the domain = abort is okay.

zingale:
okay, for the collision setup, the converging flow will setup the detonation in the center, which will then propagate outward.  So this default value for the abort parameter (0.95) is still good.

zingale:
This is ready for review.

khanakbhargava (approved):

PR #3320, opened 2026-03-19
abort the flame simulation when it reaches the edge of the domain

zingale:
This is an extension of the method of #3318 to the flame problem

## PR checklist

### DIFF
--- a/Exec/science/Detonation/Prob.cpp
+++ b/Exec/science/Detonation/Prob.cpp
@@ -1,0 +1,40 @@
+// Implementations of functions in Problem.H go here
+
+#include <AMReX_REAL.H>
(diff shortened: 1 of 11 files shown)


zingale:
I just tested a He flame with this parameter set to 0.51 and it stopped successfully when the flame reached that position.

zingale:
this is ready for review.

khanakbhargava:
docs build is failing here

zingale:
The doc failure is because of the linkcheck -- it seems that ADS is complaining that we are doing too many requests now.  We might need to just disable the linkcheck.

maximumcats (approved):

PR #3341, opened 2026-05-02
add an analysis script for the MHD LoopAdvection problem

zingale:
## PR checklist


guadabsb15 (approved):
Looks very similar to what I had. The only extra thing I had was something like `sp.annotate_timestamp()` before saving. But it is totally optional, It was just easier for me to show then the plots with the time.

PR #3363, opened 2026-07-24
Printing StarLib deviates in job_info for Detonation

RethLopes:
## PR summary

This PR prints out the deviates assigned to StarLib rates in job_info within plotfiles. The preprocessor directive that flags the presence of StarLib rates is defined in a corresponding PR to pynucastro.

## PR motivation

The goal of printing out deviates for StarLib rates is to explore the covariance of these deviates with the mean shock speed for detonation simulations.


Source/driver/Castro_io.cpp
```diff
@@ -820,6 +820,25 @@ Castro::writeJobInfo (const std::string& dir, const Real io_time)
     }
   jobInfoFile << "\n\n";
 
+  #ifdef STARLIB
+    jobInfoFile << PrettyLine;
+    jobInfoFile << " Deviates for StarLib Rates \n";
+    jobInfoFile << PrettyLine;
+    jobInfoFile << OtherLine;
```
zingale: this should come after the heading

zingale:
Right now the table prints as:

```
  ==============================================================================                                                                                                  
   Deviates for StarLib Rates                                                                                                                                                     
  ==============================================================================                                                                                                  
  ------------------------------------------------------------------------------                                                                                                  
   index                                   name        deviate                                                                                                                    
       1                   Co56_to_Fe56_weaktab        -0.0512656                                                                                                                 
       2                   Co56_to_Ni56_weaktab        -1.90693                                                                                                                   
       3                   Fe56_to_Co56_weaktab        1.23919                                                                                                                    
       4                   Ni56_to_Co56_weaktab        -0.212754                                                                                                                  
       5                         n_to_p_weaktab        -0.376303                                                                                                                  
       6                         p_to_n_weaktab        0.0660327                                                                                                                  
       7              C12_C12_to_p_Na23_starlib        -1.01706                                                                                                                   
       8            C12_C12_to_He4_Ne20_starlib        0.907624                                                                                                                   
       9              C12_O16_to_p_Al27_starlib        -0.682124                                                                                                                  
      10            C12_O16_to_He4_Mg24_starlib        -0.461218                
 ```

The separator between the heading and data is in the wrong place

**Comment by zingale** [comment]
also, why do the weaktab rates have deviates?  I think we messed that up

**Comment by zingale** [comment]
oh... no the issue is that you are using `rate_names[i]`, but that includes all rates, not just starlib ones.  So you need to index only the starlib rates when printing the names, so your names are not in the correct line in the table.

**Comment by zingale** [comment]
look at how the `get_p_random()` function works.  That gives the association between the rate and the deviate.

**Comment by zingale** [comment]
oh... I see you are not outputting if the deviate is 0.  But what happens if the random number generator outputs 0?

I think this might be simpler:

```
  #ifdef STARLIB
    jobInfoFile << PrettyLine;
    jobInfoFile << " Deviates for StarLib Rates \n";
    jobInfoFile << PrettyLine;

    jobInfoFile <<
    std::setw(6) << "index" << SkipSpace <<
    std::setw(mlen+11) << "name" << SkipSpace <<
    std::setw(7) << "deviate" <<  "\n";
    jobInfoFile << OtherLine;

    int idx = 1;
    for (int i = 1; i <= Rates::NumRates ; i++) {
        if (rate_names[i].ends_with("_starlib")) {
            jobInfoFile <<
                std::setw(6) << i << SkipSpace <<
                std::setw(mlen+11) << rate_names[i] << SkipSpace <<
                std::setw(7) << starlib::prand(idx) << "\n";
            idx++;
        }
    }
    jobInfoFile << "\n\n";
  #endif

```

File: Source/driver/Castro_io.cpp [hunk 2]
```diff
@@ -820,6 +820,31 @@ Castro::writeJobInfo (const std::string& dir, const Real io_time)
     }
   jobInfoFile << "\n\n";
 
+    #ifdef STARLIB
```
zingale: can you remove the spaces before `#ifdef` and `#endif`

RethLopes: done

zingale:
okay, now you need to make `STARLIB` a recognized ifdef so the `check_ifdefs` CI passes.  You do this by editing `.github/workflows/good_defines.txt`

Source/driver/Castro_io.cpp
```diff
@@ -820,6 +820,31 @@ Castro::writeJobInfo (const std::string& dir, const Real io_time)
...
+      jobInfoFile <<
+      std::setw(6) << idx << SkipSpace <<
+      std::setw(mlen+11) << rate_names[i] << SkipSpace <<
+      std::setw(7) << starlib::prand(idx) << "\n";
+      idx++;
+    }
+  }
+  jobInfoFile << "\n\n";
+
+  AMREX_ASSERT(starlib::NumStarLibRates == idx);
```
zingale: actually, this should compare to `idx-1`, since we increment `idx` at the end of the loop.

If you run as-is, with `DEBUG=TRUE`, then you get an assertion error.

RethLopes: ah, I didn't realize amrex assertions are only checked when compiling with `DEBUG=TRUE`. Fixed it now.

zingale: if you want, you can do `AMREX_ALWAYS_ASSERT` to have it always take effect.

zingale (approved):

PR #3353, opened 2026-06-20
subchandra: switch over to Rosenbrock

zingale:
Rosenbrock runs this problem about 2x faster, and seems to have less issues with hard burns.
also sync up the N14 and NSE 2D inputs files

## PR checklist

- [ ] test suite needs to be run on this PR
- [ ] this PR will change answers in the test suite to more than roundoff level
- [ ] all newly-added functions have docstrings as per the coding conventions
- [ ] the `CHANGES` file has been updated, if appropriate
- [ ] if appropriate, this change is described in the docs

PR #3365, opened 2026-07-24
sync the subchandra NSE inputs with the N14 one

zingale:
this uses a coarser level 0 grid + updates refinement now the two inputs largely agree

## PR checklist


zingale:
oh... looks like I did this already in #3353

PR #3369, opened 2026-07-26
after build, add a note about doing "make clean" if anything changes

zingale:
this also adds a make function to color the text green so it stands out

## PR checklist


zingale:
This is what the build output looks like with this change:

image

khanakbhargava (approved):
I like this

PR #3370, opened 2026-07-27
Make transition from CO to He in wdmerger set up smooth 

khanakbhargava:
Currently, we use a sharp transition when building the 1D profile for the wdmerger. It is based on `M_enclosed = mass_want - envelope_mass`, where `mass_want` is the total mass of the WD and `envelope_mass` is the He envelope mass. 

We want to make this transition smooth. I used the same tanh profile we use in `subchandra` 1D profile. The temperature profile is still isothermal as used in the default set up. However, since we are changing the composition as we move outward through the transition, the he mass needs to be integrated to converge to the requested envelope mass.


zingale:
can you post a plot showing a 1-d slice through the old and new model with this change?

khanakbhargava:
image (3)

khanakbhargava:
image (2)

zingale:
I'm confused how the original code was working -- where was it checking for convergence for the desired envelope mass?  Are we sure that the envelope mass was correct originally?  We don't seem to print it out.

khanakbhargava:
So the original code does not check for convergence separately. It starts assigning He comp once `M_enclosed = mass_want - envelope_mass` is reached. It then outputs the requested He shell mass in the output file when the model is generated. For example - 

> Creating primary with CO core with mass fractions C = 0.5and O = 0.5 and a He shell of solar mass =5.9652e+31.
> Creating secondary with CO core with mass fractions C = 0.5and O = 0.5 and a He shell of solar mass =5.9652e+31.
> 
> Generated initial model for primary WD of mass 1.05 solar masses, central density 4.767e+07 g cm**-3, and radius 5.047e+08 cm.
> 
> Generated initial model for secondary WD of mass 0.7 solar masses, central density 6.274e+06 g cm**-3, and radius 7.641e+08 cm.

Issue #3017, opened 2025-01-08
sponge is not implemented for true SDC


zingale:
For true-SDC, the sponge does not work since `construct_old_sponge_source` is a no-op.   This is because we treat the sponge implicitly with the CTU evolution, but SDC always needs the source at the time of the quadrature node, and therefore only ever calls the old time sources.

zingale:
the way we implement the Sponge in Castro is essentially that of Almgren et al. 2008 (Maestro paper III), which is an implicit update to the velocity:

U^new = U^old - dt kappa f U^new

this is equivalent to doing:

U^new = U^old / (1 + dt kappa f)

this is also referenced in the original Castro paper (Almgren et al. 2010) and the wdmerger paper (Katz et al. 2016).

For a MOL integration, we can't really do the implicit update, since we are just computing the advective terms at a particular instance in time.  Instead, we should just implement a sponge source in these cases that has the form:

S_sponge = - kappa f U

PR #3379, opened 2026-09-04
add the sponge for true-SDC

zingale:
## PR summary

this only should work explicitly and is a direct damping force.
This formulation matches how it is done in pyro for the FV4 solver.

## PR checklist


zingale:
closes #3017

zingale:
with an explicit sponge, there is an additional timestep limiter for stability, basically dt < tau_sponge.

Issue #3399, opened 2026-09-16
Saved SDC fluxes omit the endpoint contribution


zingale:
Locations: `Source/driver/Castro_advance_sdc.cpp:86-87` and
`Source/hydro/Castro_mol_hydro.cpp:681-711`.

## Finding

The driver skips the final time node on the final SDC sweep. The MOL routine
saves nonzero-node fluxes only on that final sweep. Thus the last node's flux is
never saved, despite its nonzero quadrature weight for every supported node set.

For SDC2 Lobatto, a constant flux F is stored as `dt*F/2`, since only node zero
contributes. SDC4 Lobatto stores `5*dt*F/6`. The state update still uses the
old-sweep endpoint RHS, so its flux divergence is not represented by the saved
flux data. P_radial uses the same flawed condition.

## Scope and evidence

This is a confirmed bookkeeping defect from the stage conditions and weights in
`Source/driver/Castro_setup.cpp:1085-1132`. True SDC currently rejects AMR, so
it should not be described as a demonstrated AMR conservation failure. It would
corrupt refluxing if those buffers were used for AMR without fixing this
inconsistency. The local state update does not use the incomplete saved flux
sum.

The numerical check script verifies the missing constant-flux weights.

## Proposed fix

Retain node fluxes alongside the old and new advective RHS and form the final
integrated flux using the same correction formula as the state update. For a
final sweep, sum the old-sweep quadrature flux and the explicit subinterval
corrections `h_m*(F_new[m]-F_old[m])`. Apply the identical logic to P_radial.
Merely evaluating the final node once more fills the obvious hole but does not
make arbitrary unconverged sweeps' saved fluxes match their actual updates.

## Validation after fixing

First verify that constant flux integrates to dt*F for all node sets. Then
compare the divergence of the saved integrated flux against the source-free
state change, including cases with a small number of sweeps. Validate the
pressure contribution separately in non-Cartesian geometry.

zingale:
Note: this is only an issue with AMR, which is not currently supported with true-SDC.

PR #3405, opened 2026-09-17
fix how the true-SDC fluxes are accumulated

zingale:
they should match how the fluxes are used in the final corrector for the SDC update
note: this is only needed for AMR, which is not supported currently for true-SDC, so this is just fixing the storage in case we do eventually support AMR

## PR checklist


zingale:
closes #3399

